PR #1, opened 2016-11-08
Provide a way to mark android devices optional

xpconanfan:
Sometimes users want to continue the test even when some Android devices in a test bed are not operable.
So here we provide an option to allow the test to execute even if some Android controller objects failed to instantiate.

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -207,8 +207,15 @@ def get_instances_with_configs(configs):
             raise Error(
                 "Required value 'serial' is missing in AndroidDevice config %s."
                 % c)
-        ad = AndroidDevice(serial)
-        ad.load_config(c)
+        is_required = c.pop("required", True)
+        try:
+            ad = AndroidDevice(serial)
+            ad.load_config(c)
+        except:
+            if not is_required:
+                logging.warning("Failed to initialize Android device %s", serial)
+                continue
+            raise
         results.append(ad)
     return results
 
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -33,10 +33,15 @@ from mobly.controllers.android_device_lib import sl4a_client
 MOBLY_CONTROLLER_CONFIG_NAME = "AndroidDevice"
 
 ANDROID_DEVICE_PICK_ALL_TOKEN = "*"
+
 # Key name for adb logcat extra params in config file.
 ANDROID_DEVICE_ADB_LOGCAT_PARAM_KEY = "adb_logcat_param"
 ANDROID_DEVICE_EMPTY_CONFIG_MSG = "Configuration is empty, abort!"
 ANDROID_DEVICE_NOT_LIST_CONFIG_MSG = "Configuration should be a list, abort!"
+
+# Keys for attributes in configs that alternate device behavior
+KEY_SKIP_SL4A = "skip_sl4a"
+KEY_DEVICE_REQUIRED = "required"
 
 
 class Error(signals.ControllerError):
@@ -123,11 +128,16 @@ def _start_services_on_ads(ads):
     for ad in ads:
         running_ads.append(ad)
         try:
-            ad.start_services(skip_sl4a=getattr(ad, "skip_sl4a", False))
-        except:
-            ad.log.exception("Failed to start some services, abort!")
-            destroy(running_ads)
-            raise
+            ad.start_services(skip_sl4a=getattr(ad, KEY_SKIP_SL4A, False))
+        except Exception as e:
+            is_required = getattr(ad, KEY_DEVICE_REQUIRED, True)
+            if is_required:
+                ad.log.exception("Failed to start some services, abort!")
+                destroy(running_ads)
+                raise
+            else:
+                logging.warning("Failed to start some service on %s: %s",
+                                ad.serial, e)
 
 
 def _parse_device_list(device_list_str, key):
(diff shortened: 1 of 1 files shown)


mobly/controllers/android_device.py
```diff
@@ -207,8 +207,15 @@ def get_instances_with_configs(configs):
             raise Error(
                 "Required value 'serial' is missing in AndroidDevice config %s."
                 % c)
-        ad = AndroidDevice(serial)
-        ad.load_config(c)
+        is_required = c.pop("required", True)
```
adorokhine: Why 'pop' and not 'get'? This would delete the attribute right?

mobly/controllers/android_device.py
```diff
@@ -207,8 +207,15 @@ def get_instances_with_configs(configs):
...
+        try:
+            ad = AndroidDevice(serial)
+            ad.load_config(c)
+        except:
+            if not is_required:
```
adorokhine: Does this take into account other issues with the device, e.g. not responsive to adb/cannot connect to sl4a?

mobly/controllers/android_device.py
adorokhine: It seems weird to me to mutate the config if we don't have to.
In the README we are teaching the user that attributes they add to ad config show up in the ad object. Except for this particular one? Why the inconsistency?

mobly/controllers/android_device.py
xpconanfan: This part only covers adb responsiveness.
As for the services, I guess we should also skip errors for optional devices?
But then it bares the question of which services are essential and which are not.

If the user sets sl4a required to True and device's required to False, what should we do when the device is discoverable in adb yet sl4a fails to start?

mobly/controllers/android_device.py
adorokhine: But 'serial' is a reserved attribute, but we don't delete that one, do we?

mobly/controllers/android_device.py
adorokhine: You could confirm with the users. I'm fine with it either way; we can launch and iterate and see if causes problems.

mobly/controllers/android_device.py
xpconanfan: Let's leave everything in with 'get', so that we could check 'required' during start services call as well.

mobly/controllers/android_device.py
```diff
@@ -123,11 +128,16 @@ def _start_services_on_ads(ads):
...
+            ad.start_services(skip_sl4a=getattr(ad, KEY_SKIP_SL4A, False))
+        except Exception as e:
+            is_required = getattr(ad, KEY_DEVICE_REQUIRED, True)
+            if is_required:
+                ad.log.exception("Failed to start some services, abort!")
+                destroy(running_ads)
+                raise
+            else:
+                logging.warning("Failed to start some service on %s: %s",
+                                ad.serial, e)
```
adorokhine: Will this print the whole stacktrace? If not, do we want it to?
(would it make sense to use logging.exception here?)
Also should we mention "continuing anyway" in the message?

mobly/controllers/android_device.py
```diff
@@ -207,15 +217,16 @@ def get_instances_with_configs(configs):
...
-            if not is_required:
-                logging.warning("Failed to initialize Android device %s", serial)
-                continue
-            raise
+        except Exception as e:
+            if is_required:
+                raise
+            logging.warning("Failed to initialize optional device %s: %s",
+                            serial, e)
+            continue
```
adorokhine: I think we wanted to append None instead of continuing, right? Otherwise you can't find out how many devices you're supposed to have and which ones failed to come up. It also makes the indexes wrong:
"AndroidDevice": ["A", "B", "C"]

so you'd expect ad[2] to be C, but if B failed to come up it's actually [1].

mobly/controllers/android_device.py
xpconanfan: Seems odd to print stacktrace if users don't care what happened to this device.
We'll end up with a bunch of stack trace causing false alarms?
I thought about using exception but decided not to.
But I could do either way.

mobly/controllers/android_device.py
xpconanfan: For the current use case requested by users, they just want to skip these devices and don't care about the indexing. Because they are simply applying the same logic to a list of devices with no assumption on order etc.

I don't think appending None is a good idea because that says nothing about the identity of the missing object, and adds complications for logics that iterate through the device list.

Since one has to use the json format, the user should not rely on index order to obtain device object should it be necessary to id certain devices in the test bed. They should figure out the identity via android_device.get_device and filter by 'label' or other field.

So if a logic does need to know the object's role in the test bed, it can be properly managed. Appending None doesn't really benefit anything.

mobly/controllers/android_device.py
adorokhine: If you've thought about it and intentionally chose to do it this way, sgtm :)
How about saying "continuing anyway because device is not required" in the message?

xpconanfan:
**What's the disadvantage of using the index to identify which phone is which?**
Making the assumption by order is less manageable, since the assumption is satisfied in the config file and is completely obscure to a new code reader. It'll be very difficult for the reader to figure out what the test expects a device to be capable of purely by index.
Forcing tests that care about device identity to explicitly acquire device objects by 'label' or some kind of id clarifies the assumption in code, which makes the test code more maintainable.
In the meantime, we are not sacrificing any convenience for tests that do not care about the device identity. Since such tests only care about they got the right number of objects.
We do need to add an option to specify the expected number of objects in register_controller though.

PR #4, opened 2016-12-07
Cleanup

xpconanfan:


mobly/controllers/android_device_lib/sl4a_client.py
```diff
@@ -128,10 +128,11 @@ def is_sl4a_running(adb_proxy):
         True if the sl4a app is running, False otherwise.
     """
     # Grep for process with a preceding S which means it is truly started.
-    out = adb_proxy.shell('ps | grep "S com.googlecode.android_scripting"')
-    if len(out) == 0:
+    try:
+        out = adb_proxy.shell('ps | grep "S com.googlecode.android_scripting"')
```
adorokhine: Android <=M doesn't have the new adb, so it will never propagate ret codes regardless of how new the host binary is.
So I think this new code will think Android <= M always has sl4a.

I think you're going to have to do something like this:

        # Grep for process with a preceding S which means it is truly started
        try:
          out = adb_proxy.shell('ps | grep "S com.googlecode.android_scripting"')
          return bool(out)
        except adb.AdbError as e:
          if e.ret_code == 1 and not e.stdout and not e.stderr:
            return False
          raise e

mobly/test_runner.py
```diff
@@ -247,7 +247,7 @@ def is_testfile_name(name, ext):
                         # under py2, because "raise X from Y" syntax is only
                         # supported under py3.
                         self.log.exception(msg)
-                        raise USERError(msg)
+                        raise ValueError(msg)
```
adorokhine: Is this codepath still accessible?
I thought this was dead code now that the test is the main.

mobly/controllers/android_device_lib/sl4a_client.py
adorokhine: The logic in the except block necessary when >M.

mobly/test_runner.py
adorokhine: In general YAGNI unless you think it's an important use case, but I'm ok with leaving proper removal for a later PR :)

xpconanfan: Yeah, let's try the issue board and track the removal there ;)

PR #8, opened 2016-12-07
Rename and refactor the tests tree.

adorokhine:
* Make the tests tree match the src tree instead of lumping everything into a flat dir.
* Remove unnecessary mobly_* prefix on all tests.
* Remove relative imports from test tree.
* Reorder tests alphabetically within the unit test suite.


xpconanfan:
Do we need the 'mobly' subdir at all? Since 'tests' is under 'mobly' anyway?
Or do you foresee more subdir under 'tests'?

adorokhine:
I added the 'mobly' dir to mirror the fact that all the source is under a 'mobly' dir.
Do you expect there to ever be another dir at the top level?
If yes we need to keep it because the tests tree will contain tests for both mobly and the new dir.
If no we could drop it.

xpconanfan:
I see, I don't see another top level dir for now. But who knows...
As long as it's easy to run, it's ok.
Does this mean we want to organize the test files in the same way as the actual files?
Say, android_device test should live under 'tests/mobly/controllers'?

tests/mobly_unittest_suite.py
```diff
@@ -17,26 +17,25 @@
...
-import mobly_android_device_test
-import mobly_asserts_test
-import mobly_base_class_test
-import mobly_logger_test
-import mobly_records_test
-import mobly_sl4a_client_test
-import mobly_test_runner_test
-import mobly_utils_test
-
+from tests.mobly import asserts_test
```
xpconanfan: i think this file is no longer needed since the tests are run by setup.py's 'test' action now.

adorokhine:
> Say, android_device test should live under 'tests/mobly/controllers'?
That's what I did. If you look at the diffs in the PR view you can see a little arrow showing new and old names for the renames.

tests/mobly_unittest_suite.py
adorokhine: Deleted the file.

PR #10, opened 2016-12-07
Add an arg to check the number of controller objects

xpconanfan:
Add an arg in register_controller so users can declare the requirement for the min number of controller objects.

Also clean up unit tests for register_controller:
* Add a test for the new arg
* Remove an old test for the now-removed builtin controller code path
* Rename some tests to match removing the builtin controller code path

### DIFF
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -383,13 +383,13 @@ class TestRunner(object):
             raise signals.ControllerError(
                 "Controller module %s did not return a list of objects, abort."
                 % module_ref_name)
-        if min_number is not None:
-            actual_number = len(objects)
-            if actual_number < min_number:
-                module.destroy(objects)
-                raise signals.ControllerError(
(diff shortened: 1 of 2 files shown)


mobly/test_runner.py
```diff
@@ -377,6 +383,13 @@ def get_info(objects):
             raise signals.ControllerError(
                 "Controller module %s did not return a list of objects, abort."
                 % module_ref_name)
+        if min_number is not None:
```
adorokhine: Why support None here? I'd prefer to just simplify the code since the default is 1 anyway.

xpconanfan: Oops, my mistake.
Originally I used None, then forgot to remove this.

xpconanfan: Also, any preference on the arg name? 'min_num'? 'min_num_of_obj'?

PR #11, opened 2016-12-07
Skip adb root if a device is on a build that's not rootable.

xpconanfan:


mobly/controllers/android_device.py
```diff
@@ -366,7 +366,7 @@ def __init__(self, serial="", host_port=None, device_port=8080):
         self.adb_logcat_file_path = None
         self.adb = adb.AdbProxy(serial)
         self.fastboot = fastboot.FastbootProxy(serial)
-        if not self.is_bootloader:
+        if not self.is_bootloader and self.is_rootable:
```
xpconanfan: That's intentionally left there.
Since you brought up the point (during internal commit) that sometimes users do want to force adb root on a device regardless of the build, and we decided not to block it if user calls ad.root_adb intentionally.

adorokhine: Sorry, wrong line... 724

PR #12, opened 2016-12-07
Remove support for signaling test result by return values

xpconanfan:
Also add an entry in .gitignore.

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -69,7 +69,7 @@ class BaseTestClass(object):
         return self
 
     def __exit__(self, *args):
-        self._exec_func(self.clean_up)
+        self._safe_exec_func(self.clean_up)
 
     def unpack_userparams(self,
                           req_param_names=[],
@@ -131,9 +131,7 @@ class BaseTestClass(object):
         """Setup function that will be called before executing any test case in
         the test class.
 
-        To signal setup failure, user asserts or raise your own exception. The
-        stack trace would appear in log, but the exceptions would not propagate
-        to upper levels.
+        To signal setup failure, use asserts or raise your own exception.
 
         Implementation is optional.
         """
@@ -162,9 +160,7 @@ class BaseTestClass(object):
         """Setup function that will be called every time before executing each
         test case in the test class.
 
-        To signal setup failure, user asserts or raise your own exception. The
-        stack trace would appear in log, but the exceptions would not propagate
-        to upper levels.
+        To signal setup failure, use asserts or raise your own exception.
 
         Implementation is optional.
         """
@@ -408,7 +404,7 @@ class BaseTestClass(object):
                 failed_settings.append(s)
         return failed_settings
 
-    def _exec_func(self, func, *args):
+    def _safe_exec_func(self, func, *args):
         """Executes a function with exception safeguard.
 
         This will let signals.TestAbortAll through so abort_all works in all
@@ -419,8 +415,7 @@ class BaseTestClass(object):
             args: Arguments to be passed to the function.
 
         Returns:
-            Whatever the function returns, or False if unhandled exception
-            occured.
+            Whatever the function returns.
         """
         try:
             return func(*args)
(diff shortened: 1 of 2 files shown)


xpconanfan:
It does beg the question: what should happen if a test returns something?

mobly/controllers/android_device.py
```diff
@@ -366,7 +366,7 @@ def __init__(self, serial="", host_port=None, device_port=8080):
         self.adb_logcat_file_path = None
         self.adb = adb.AdbProxy(serial)
         self.fastboot = fastboot.FastbootProxy(serial)
-        if not self.is_bootloader:
+        if not self.is_bootloader and self.is_rootable:
```
adorokhine: Seems like this PR also mixes in the adb root change?
I'm guessing you didn't start it on a clean branch?

xpconanfan: ohhh, each branch needs to be started on master, is that so?.... ughhh

adorokhine:
Still vestiges of this left in other methods, like setup_class:
"To signal setup failure, return False or raise an exception."

And the return path...

mobly/base_test.py
```diff
@@ -126,15 +126,15 @@ def _setup_class(self):
...
+        self.setup_class()
 
     def setup_class(self):
         """Setup function that will be called before executing any test case in
         the test class.
 
-        To signal setup failure, return False or raise an exception. If
-        exceptions were raised, the stack trace would appear in log, but the
-        exceptions would not propagate to upper levels.
+        To signal setup failure, user asserts or raise your own exception. The
```
adorokhine: use asserts

mobly/base_test.py
```diff
@@ -126,15 +126,15 @@ def _setup_class(self):
...
+        stack trace would appear in log, but the exceptions would not propagate
```
adorokhine: not sure what 'exceptions would not propagate' means... that makes it sound like setup_class can't fail the test.

mobly/base_test.py
```diff
@@ -157,15 +157,15 @@ def _setup_test(self, test_name):
...
+        self.setup_test()
 
     def setup_test(self):
         """Setup function that will be called every time before executing each
         test case in the test class.
 
-        To signal setup failure, return False or raise an exception. If
-        exceptions were raised, the stack trace would appear in log, but the
-        exceptions would not propagate to upper levels.
+        To signal setup failure, user asserts or raise your own exception. The
```
adorokhine: use asserts

mobly/base_test.py
```diff
@@ -157,15 +157,15 @@ def _setup_test(self, test_name):
...
+        stack trace would appear in log, but the exceptions would not propagate
```
adorokhine: same as above

mobly/base_test.py
```diff
@@ -545,8 +533,7 @@ def run(self, test_names=None):
         class_record.test_begin()
         # Setup for the class.
         try:
-            if self._setup_class() is False:
-                asserts.fail("Failed to setup %s." % self.TAG)
+            self._setup_class()
         except Exception as e:
             self.log.exception("Failed to setup %s.", self.TAG)
```
adorokhine: _exec_func swallows the exception and returns False, but that return value is never used anywhere.

- Is that return still needed?
- Can we give _exec_func a better name to more clearly show that it swallows exceptions?

tests/mobly_base_class_test.py
```diff
@@ -199,24 +199,6 @@ def test_something(self):
                             "Skipped 0, Unknown 0")
```
adorokhine: We don't have any tests that tested "return False" from tests themselves? Only setup?

mobly/base_test.py
xpconanfan: Which part do you want to replace?...

xpconanfan: oh, got it

tests/mobly_base_class_test.py
xpconanfan: I think I hated that behavior so much that I didn't write a test for it...

mobly/base_test.py
xpconanfan: Ah that's a relic.
I think back in the days tests didn't fail when setup failed and that bug was there for a long time............. Removing it from doc strings

PR #13, opened 2016-12-08
Update README to utilize min_number check in examples.

xpconanfan:


README.md
```diff
@@ -253,7 +253,7 @@ from mobly.controllerse import android_device
 class HelloWorldTest(base_test.BaseTestClass):
 
   def setup_class(self):
-    self.ads = self.register_controller(android_device)
+    self.ads = self.register_controller(android_device, min_number=2)
```
adorokhine: Do you think this needs a description? Seems weird to just throw it in there without any context

xpconanfan: I'd think it's pretty self explanatory?
People can always read the code or docs (when that exists)...

adorokhine: I don't think it's self explanatory, especially for people who don't really know what register_controller does. It seems obvious to you because you implemented :)

xpconanfan: If they don't know what register_controller does at this point in README, we have bigger problems?...
Looks like we need more inline comments in general.

PR #23, opened 2016-12-12
Dropping code paths for importing test class files in test_runner.

### LINKED ISSUE #5: Remove code paths related to mobly.test_runner.TestRunner.import_test_modules
In the new invocation model, the import_test_modules function and its associated code paths are not needed any more and should be removed.

### LINKED ISSUE #21: testpaths required to be in config even though it is unused

xpconanfan:
This includes:
* Remove mobly.test_runner.TestRunner.import_test_modules
* Remove the need for key 'testpaths'
* Refactor code to accommodate the removals. (cleaner code)

fixes #21 
fixes #5

### DIFF
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -146,7 +146,7 @@ def execute_one_test_class(test_class, test_config, test_identifier):
     """
     tr = TestRunner(test_config, test_identifier)
     try:
-        tr.run(test_class)
+        tr.run([test_class])
         return tr.results.is_all_pass
     except signals.TestAbortAll:
         raise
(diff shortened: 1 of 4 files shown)


tests/mobly/test_runner_test.py
```diff
@@ -179,7 +180,7 @@ def test_run_two_test_classes(self, mock_get_all, mock_list_adb,
         tr = test_runner.TestRunner(mock_test_config,
                                     [('IntegrationTest', None),
                                      ('IntegrationTest', None)])
-        tr.run()
+        tr.run([IntegrationTest.IntegrationTest])
```
adorokhine: Seems like all the unit tests specify a single class?
Should we add a test for when multiple classes are specified?

mobly/test_runner.py
```diff
@@ -472,7 +421,7 @@ def run_test_class(self, test_cls_name, test_cases=None):
                 self.results += e.results
                 raise e
 
-    def run(self, test_class=None):
+    def run(self, test_classes):
```
adorokhine: How come we were ok with a single test_class before, but now we need a list? Is there a usecase?

xpconanfan: Because for the plural case, which happens when test_class is None, 'run' used to import test files.
Now that the import test files functions are removed, 'run' should take a list instead.

xpconanfan: It was used extensively in Android. It's the basis for executing multiple test classes in one command (-tc XxxTest YyyTest)
The intention is that users can easily construct their test suite in code, similar to in unittest. More work is needed to document this, but leaving the hook here is useful

PR #26, opened 2016-12-13
Add controller code for attenuator.

### LINKED ISSUE #9: Add attenuator controller
Refactor and open source the attenuator controller code.

xpconanfan:
* Add an attenuator device lib implemented for minicircuit device.
* The minicircuit implementation showcases the interface expected
  for an attenuator controller.
* Move keys only used by the attenuator module out of mobly.keys.

fixes #9


mobly/controllers/attenuator.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    def __init__(self, attn_device, idx=0, name=None):
+        self.model = attn_device.model
+        self.attn_device = attn_device
+        self.idx = idx
+        if (self.idx >= attn_device.path_count):
+            raise IndexError(
+                "Attenuator index out of range for attenuator attn_device")
+
+    def set_attn(self, value):
```
dthkao: Same for these. Why shorten the method names?

mobly/controllers/attenuator.py
```diff
@@ -0,0 +1,116 @@
...
+        """This function sets the attenuation of Attenuator.
+
+        Args:
+            value: This is a floating point value for nominal attenuation to be
```
dthkao: units? db?

mobly/controllers/attenuator.py
```diff
@@ -0,0 +1,116 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import importlib
+import logging
+
+MOBLY_CONTROLLER_CONFIG_NAME = "Attenuator"
+KEY_ADDRESS = "Address"
```
dthkao: This assumes any attenuator device will be telnet. Is there any chance one might use a different connection method?

xpconanfan: No, this is just a macro.
When a new way of connection is needed, can just add more macros

mobly/controllers/attenuator.py
xpconanfan: 'attn' and 'atten' have been mixed among doc, code, and reference.
I'm making the abbreviated reference consistent everywhere by using 'attn'

mobly/controllers/attenuator_lib/minicircuits.py
```diff
@@ -0,0 +1,134 @@
...
+from mobly.controllers.attenuator_lib import telnet_client
+
+
+class AttenuatorDevice(object):
+    """This provides a specific telnet-controlled implementation of
+    AttenuatorDevice for Mini-Circuits RC-DAT attenuators.
+
+    Attributes:
+        path_count: The number of signal attenuation path this device has.
+        _telnet_client: The telnet client used to communicate with the
```
adorokhine: I don't think private attributes are typically documented in the user-facing module documentation. If this comment is valuable, it should probably be a comment at the variable declaration site.

mobly/controllers/attenuator_lib/telnet_client.py
```diff
@@ -0,0 +1,85 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Helper module for common telnet capability to communicate with
+AttenuatorDevice(s).
+
+User code shouldn't need to directly access this class.
+"""
+
+import telnetlib
```
adorokhine: Does this require setup.py changes?

mobly/controllers/attenuator_lib/telnet_client.py
```diff
@@ -0,0 +1,85 @@
...
+from mobly.controllers import attenuator
+
+
+def _ascii_string(uc_string):
+    return str(uc_string).encode('ASCII')
+
+
+class TelnetClient(object):
+    """This is an internal helper class for Telnet+SCPI command-based
```
adorokhine: Is there anything specific to SCPI in this class? If so should it be telnet_scpi_client?

mobly/controllers/attenuator.py
```diff
@@ -0,0 +1,116 @@
...
+KEY_PORT = "Port"
+
+
+def create(configs):
+    objs = []
+    for c in configs:
```
adorokhine: Could you avoid abbreviating "c" here? Makes the long loop a little hard to follow.

mobly/controllers/attenuator.py
```diff
@@ -0,0 +1,116 @@
...
+        attn_model = c["Model"]
```
adorokhine: android_device has lowercase keys right? Could we be consistent?

mobly/controllers/attenuator.py
```diff
@@ -0,0 +1,116 @@
...
+        # Default to telnet.
+        protocol = c.get("Protocol", "telnet")
+        # Import the correct driver module for the attenuator device
```
adorokhine: Not sure why we need to do this import; won't register_controller(minicircuits) take care of this?

mobly/controllers/attenuator.py
```diff
@@ -0,0 +1,116 @@
...
+        module_name = "mobly.controllers.attenuator_lib.%s" % attn_model
+        module = importlib.import_module(module_name)
+        # Create each
+        path_cnt = c["PathCount"]
+        attn_device = module.AttenuatorDevice(path_cnt)
+        attn_device.model = attn_model
+        insts = attn_device.open(c[KEY_ADDRESS], c[KEY_PORT])
```
adorokhine: instances

mobly/controllers/attenuator.py
```diff
@@ -0,0 +1,116 @@
...
+    objs = []
+    for c in configs:
+        attn_model = c["Model"]
+        # Default to telnet.
+        protocol = c.get("Protocol", "telnet")
+        # Import the correct driver module for the attenuator device
+        module_name = "mobly.controllers.attenuator_lib.%s" % attn_model
+        module = importlib.import_module(module_name)
+        # Create each
+        path_cnt = c["PathCount"]
```
adorokhine: What is a PathCount? A single attenuator can have more than one path?

mobly/controllers/attenuator.py
adorokhine: You have a lot of inline keys in the module. I'm ok with that (no need to make them all constants) but could you add an example config fragment showing what it should look like?

xpconanfan: Done

mobly/controllers/attenuator.py
xpconanfan: Done

mobly/controllers/attenuator.py
xpconanfan: which case do we want to choose?

mobly/controllers/attenuator.py
xpconanfan: No, you would not register minicircuits directly, you would register this module, which in turn loads the appropriate 'driver' based on the config

mobly/controllers/attenuator.py
xpconanfan: Done

mobly/controllers/attenuator_lib/minicircuits.py
xpconanfan: Done

mobly/controllers/attenuator.py
xpconanfan: Done

mobly/controllers/attenuator.py
adorokhine: What are the advantages of implementing another system of indirection versus just using register_controller?

mobly/controllers/attenuator.py
```diff
@@ -0,0 +1,116 @@
...
+        self.attenuators[0].set_attn(0, 50)
+        self.attenuators[0].set_attn(1, 40)
+
+    The benefit the former is that the physical test bed can use either four
+    single-channel attenuators, or one four-channel attenuators. Whereas the
+    latter forces the test bed to use a four-channel attenuator.
+    """
+    def __init__(self, attn_device, idx=0, name=None):
+        self.model = attn_device.model
+        self.attn_device = attn_device
```
xpconanfan: Spelling out every 'attenuation' is gonna make things unnecessarily long?
should be ok as long as it's consistent?
and set_attn and get_attn seem pretty straightforward?

dthkao: The varnames you chose in the example code, the classnames, and package names spell it out, and I think it makes it clearer. Are we really worried about verbosity?

mobly/controllers/attenuator.py
```diff
@@ -0,0 +1,134 @@
...
+    for config in configs:
+        attn_model = config["Model"]
+        # Import the correct driver module for the attenuator device
+        module_name = "mobly.controllers.attenuator_lib.%s" % attn_model
+        module = importlib.import_module(module_name)
+        # Create each
+        path_cnt = config["PathCount"]
+        attn_device = module.AttenuatorDevice(path_cnt)
+        attn_device.model = attn_model
+        instances = attn_device.open(config[KEY_ADDRESS], config[KEY_PORT])
```
dthkao: This is what I mean as assuming the attenuator device will be telnet.

xpconanfan: Yes it is.
We can easily change this later if we need to support a different driver. And no test code needs to change if someone chooses to try running with a different hardware.
The point is to avoid test scripts having to change because hardware changed.

xpconanfan: But if you want to abstract it out, we can do that later.

mobly/controllers/attenuator.py
xpconanfan: Discussed offline, use 'atten' instead of 'attn' because 'attn' can also mean other things like 'attention'

PR #32, opened 2016-12-15
Allow passing argv to main directly.

xpconanfan:


mobly/test_runner.py
```diff
@@ -35,7 +35,7 @@
 from mobly import utils
 
 
-def main():
+def main(argv=sys.argv[1:]):
```
adorokhine: You should never provide mutable default arguments in python:
http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

xpconanfan: But that takes away the fun of uncertainty!

mobly/test_runner.py
```diff
@@ -35,7 +35,7 @@
 from mobly import utils
 
 
-def main():
+def main(argv=None):
     """Execute the test class in a test module.
 
     This is the default entry point for running a test script file directly.
```
adorokhine: Needs an Args: block?

PR #34, opened 2016-12-15
Refactor the shell code into a reusable library.

adorokhine:
This makes it easier to create a shell for the snippet lib.

For #27.


adorokhine:
Rebased

PR #35, opened 2016-12-15
Implement a client for snippet support and integrate into android_device.

adorokhine:
For #27.

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -486,7 +486,7 @@ class AndroidDevice(object):
         self.adb.root()
         self.adb.wait_for_device()
 
-    def load_snippets(self, name, package):
+    def load_snippet(self, name, package):
         """Starts the snippet apk with the given package name and connects.
 
         Args:
@@ -497,7 +497,7 @@ class AndroidDevice(object):
 
         Examples:
             >>> ad = AndroidDevice()
-            >>> ad.load_snippets(
+            >>> ad.load_snippet(
                     name='maps', package='com.google.maps.snippets')
             >>> ad.maps.activateZoom('3')
         """
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -29,6 +29,7 @@ from mobly.controllers.android_device_lib import event_dispatcher
 from mobly.controllers.android_device_lib import fastboot
 from mobly.controllers.android_device_lib import jsonrpc_client_base
 from mobly.controllers.android_device_lib import sl4a_client
+from mobly.controllers.android_device_lib import snippet_client
 
 MOBLY_CONTROLLER_CONFIG_NAME = "AndroidDevice"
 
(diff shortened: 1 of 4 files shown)


adorokhine:
Rebased.

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -214,7 +217,7 @@ def connect(self, port, addr='localhost', uid=UNKNOWN_UID,
                     logging.exception("Failed to create socket connection!")
                     raise
                 time.sleep(1)
-
+        self.port = port
```
xpconanfan: maybe we should name the arg more specifically.
Is this the port on the device-side?

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -149,6 +149,10 @@ def _is_app_running(self):
 
     # Rest of the client methods.
 
+    def check_app_installed(self):
+      if not self._is_app_installed():
+            raise AppStartError('App is not installed on %s' % self._adb.serial)
```
xpconanfan: let's specify which app?

mobly/controllers/android_device.py
```diff
@@ -487,6 +486,31 @@ def root_adb(self):
         self.adb.root()
         self.adb.wait_for_device()
 
+    def load_snippets(self, name, package):
```
xpconanfan: should it be 'load_snippet' singular since it's loading one package here?

mobly/controllers/android_device_lib/jsonrpc_client_base.py
adorokhine: The port is on whatever addr is. The client doesn't know or care if the port is local or not as long as it can connect to addr:port.

xpconanfan: Ah didn't notice the 'addr' arg.
Do we need to have a default?
Perhaps it's better to force caller to be explicit about addr; also we can have the nice order of (addr, port, ...) as opposed to the reverse.

mobly/controllers/android_device_lib/jsonrpc_client_base.py
adorokhine: Done

mobly/controllers/android_device.py
adorokhine: Done

mobly/controllers/android_device_lib/sl4a_client.py
```diff
@@ -29,6 +29,10 @@
 
 class Sl4aClient(jsonrpc_client_base.JsonRpcClientBase):
 
+    def __init__(self, adb_proxy):
+      super(Sl4aClient, self).__init__(adb_proxy)
+      self.app_name = 'sl4a'
```
xpconanfan: How about upper case 'SL4A' here?
Since this token often appears at the beginning of a log message.

mobly/controllers/android_device_lib/jsonrpc_client_base.py
adorokhine: I expect in 100% of cases that addr will be localhost. Actually I was considering removing it. People will be specifying just 'port' so the ordering shouldn't matter much.

mobly/controllers/android_device_lib/sl4a_client.py
adorokhine: Done.

PR #36, opened 2016-12-15
Implement a shell client for snippets.

adorokhine:
For #27.


xpconanfan:
Is it possible to have a shell that takes care of both?
Can do it later if intelligently detecting what's installed is difficult.

adorokhine:
It's possible. What would you like to call it? And what kind of command line args do you picture?

xpconanfan:
Not high priority.
Let's get this in for now.

Filed #38 to track the next step.

PR #47, opened 2016-12-21
Remove redundant timeout logic from JsonRpcClientBase.connect.

### LINKED ISSUE #44: Snippet client is unable to make connection on mac
Because the convoluted timeout scheme of JsonRpcClientBase.connect.

xpconanfan:
* This logic is over-complicated, especially considering there is only
  one call to it in the entire code base and that call does not even
  use the timeout arg of 'connect'.
* This logic causes eternal hang on macs.

fixes #44

### DIFF
--- a/mobly/controllers/android_device_lib/jsonrpc_client_base.py
+++ b/mobly/controllers/android_device_lib/jsonrpc_client_base.py
@@ -176,12 +176,12 @@ class JsonRpcClientBase(object):
             '%s failed to start on %s.' % (self.app_name, self._adb.serial))
 
     def connect(self, port, addr='localhost', uid=UNKNOWN_UID,
-                connection_timeout=None, cmd=JsonRpcCommand.INIT):
+                cmd=JsonRpcCommand.INIT):
         """
         Opens a connection to the remote client.
 
         Opens a connection to a remote client. The connection will error out if
-        it takes longer than the connection_timeout time. Once connected if the
+        it takes longer than the _SOCKET_TIMEOUT seconds. Once connected if the
         socket takes longer than _SOCKET_TIMEOUT to respond the connection will
         be closed.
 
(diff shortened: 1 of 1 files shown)


xpconanfan:
Tested with both sl4a and snippet shells.
Working fine.

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -199,28 +199,16 @@ def connect(self, port, addr='localhost', uid=UNKNOWN_UID,
             socket.timeout: Raised when the socket waits to long for connection.
             ProtocolError: Raised when there is an error in the protocol.
         """
-        if connection_timeout:
-            timeout_time = time.time() + connection_timeout
-        else:
-            timeout_time = sys.maxsize
         self._counter = self._id_counter()
         while True:
```
dthkao: Is this loop still needed?

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -176,12 +176,12 @@ def start_app(self, wait_time=APP_START_WAIT_TIME):
...
 
     def connect(self, port, addr='localhost', uid=UNKNOWN_UID,
-                connection_timeout=None, cmd=JsonRpcCommand.INIT):
+                cmd=JsonRpcCommand.INIT):
         """
         Opens a connection to the remote client.
 
         Opens a connection to a remote client. The connection will error out if
-        it takes longer than the connection_timeout time. Once connected if the
+        it takes longer than the _SOCKET_TIMEOUT seconds. Once connected if the
```
dthkao: This phrasing is a little confusing. Can you clarify that the second timeout applies to any other rpc?

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -176,12 +176,12 @@ def start_app(self, wait_time=APP_START_WAIT_TIME):
             '%s failed to start on %s.' % (self.app_name, self._adb.serial))
 
     def connect(self, port, addr='localhost', uid=UNKNOWN_UID,
-                connection_timeout=None, cmd=JsonRpcCommand.INIT):
+                cmd=JsonRpcCommand.INIT):
```
dthkao: Was the removed param not used anywhere?

xpconanfan: We only make one call to this function and that call dose not use this param.
So there's really no point in having it for practical purposes.

mobly/controllers/android_device_lib/jsonrpc_client_base.py
xpconanfan: Done

xpconanfan:
It was added here
https://android-review.googlesource.com/#/c/219831/
The new JsonRpcClientBase.start_app already handles the problems of sl4a not running before connection is attempted.

PR #58, opened 2016-12-22
Fix Py3 compatibility issues in rpc clients.

### LINKED ISSUE #45: Snippet client is not py3 compatible.
Because of the byte string handling.
A couple examples:
1. >>> ad.load_snippet(name='spt', package='com.google.android.mobly.snippet.example1')
WARNING:root:[AndroidDevice|84B7N15A28012457] a bytes-like object is required, not 'str'

2. >>> ad.stop_services()
  File "/Users/angli/Developer/mobly/mobly/controllers/android_device.py", line 417, in stop_services
    self._terminate_jsonrpc_client(client)
  File "/Users/angli/Developer/mobly/mobly/controllers/android_device.py", line 576, in _terminate_jsonrpc_client
    client.stop_app()
  File "/Users/angli/Developer/mobly/mobly/controllers/android_device_lib/snippet_client.py", line 68, in stop_app
    if 'OK (0 tests)' not in out:
TypeError: a bytes-like object is required, not 'str'

xpconanfan:
So we don't get errors like this in Py3:
>>> ad.stop_services()
  File "/Users/angli/Developer/mobly/mobly/controllers/android_device.py", line 417, in stop_services
    self._terminate_jsonrpc_client(client)
  File "/Users/angli/Developer/mobly/mobly/controllers/android_device.py", line 576, in _terminate_jsonrpc_client
    client.stop_app()
  File "/Users/angli/Developer/mobly/mobly/controllers/android_device_lib/snippet_client.py", line 68, in stop_app
    if 'OK (0 tests)' not in out:
TypeError: a bytes-like object is required, not 'str'

fixes #45

### DIFF
--- a/mobly/controllers/android_device_lib/sl4a_client.py
+++ b/mobly/controllers/android_device_lib/sl4a_client.py
@@ -13,7 +13,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 """JSON RPC interface to android scripting engine."""
 
 from mobly.controllers.android_device_lib import adb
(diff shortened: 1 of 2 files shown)


xpconanfan:
Py3 is actually pretty nice, especially when I was debugging my new snippet apk today.
Py2 was hiding some stacktrace, but in Py3, I was able to figure out what happened with the full stacktrace ;)

adorokhine:
Could you add a logging.exception in the place where the cause was getting lost to make sure we get the complete stack trace in some form under both pythons?

xpconanfan:
I'll try. But I don't remember exactly where it was as I was focused on getting the apk to work.

PR #60, opened 2016-12-22
Snippet shell improvements

xpconanfan:
* Improve argparse usage in snippet shell to use positional arg.
* Add a check to verify the device is actually connected.
* Change log level to be less verbose in console.

### DIFF
--- a/mobly/controllers/android_device_lib/snippet_client.py
+++ b/mobly/controllers/android_device_lib/snippet_client.py
@@ -53,13 +53,16 @@ class SnippetClient(jsonrpc_client_base.JsonRpcClientBase):
     def _do_start_app(self):
         """Overrides superclass."""
         cmd = _LAUNCH_CMD.format(self._port, self._package)
-        logging.debug('Launching snippet apk with: %s', cmd)
+        # Use info here so people know exactly what's happening here, which is
+        # helpful since they need to create their own instrumentations and
+        # anifest.
+        logging.info('Launching snippet apk with: %s', cmd)
         self._adb.shell(_LAUNCH_CMD.format(self._port, self._package))
 
     def stop_app(self):
         """Overrides superclass."""
         cmd = _STOP_CMD.format(self._package)
-        logging.debug('Stopping snippet apk with: %s', cmd)
+        logging.info('Stopping snippet apk with: %s', cmd)
         out = self._adb.shell(_STOP_CMD.format(self._package)).decode('utf-8')
         if 'OK (0 tests)' not in out:
             raise Error('Failed to stop existing apk. Unexpected output: ' +
--- a/mobly/controllers/android_device_lib/snippet_client.py
+++ b/mobly/controllers/android_device_lib/snippet_client.py
@@ -55,7 +55,7 @@ class SnippetClient(jsonrpc_client_base.JsonRpcClientBase):
         cmd = _LAUNCH_CMD.format(self._port, self._package)
         # Use info here so people know exactly what's happening here, which is
         # helpful since they need to create their own instrumentations and
-        # anifest.
+        # manifest.
         logging.info('Launching snippet apk with: %s', cmd)
         self._adb.shell(_LAUNCH_CMD.format(self._port, self._package))
 
--- a/mobly/controllers/android_device_lib/snippet_client.py
+++ b/mobly/controllers/android_device_lib/snippet_client.py
@@ -53,13 +53,13 @@ class SnippetClient(jsonrpc_client_base.JsonRpcClientBase):
     def _do_start_app(self):
         """Overrides superclass."""
         cmd = _LAUNCH_CMD.format(self._port, self._package)
-        logging.info('Launching snippet apk with: %s', cmd)
+        logging.debug('Launching snippet apk with: %s', cmd)
         self._adb.shell(_LAUNCH_CMD.format(self._port, self._package))
 
     def stop_app(self):
         """Overrides superclass."""
         cmd = _STOP_CMD.format(self._package)
-        logging.info('Stopping snippet apk with: %s', cmd)
+        logging.debug('Stopping snippet apk with: %s', cmd)
         out = self._adb.shell(_STOP_CMD.format(self._package)).decode('utf-8')
         if 'OK (0 tests)' not in out:
             raise Error('Failed to stop existing apk. Unexpected output: ' +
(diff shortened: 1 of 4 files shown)


mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -53,13 +53,13 @@ def __init__(self, package, port, adb_proxy):
     def _do_start_app(self):
         """Overrides superclass."""
         cmd = _LAUNCH_CMD.format(self._port, self._package)
-        logging.info('Launching snippet apk with: %s', cmd)
+        logging.debug('Launching snippet apk with: %s', cmd)
```
adorokhine: I want this log on console. Since people have to create their own instrumentations and manifest, seeing exactly what the command is doing helps a lot with troubleshooting issues like missing instrumentation.

tools/sl4a_shell.py
```diff
@@ -48,12 +48,11 @@ def _start_services(self, console_env):
         console_env['ed'] = self._ad.ed
 
     def _get_banner(self, serial):
-        return """
-Connected to {}. Call methods against:
-    ad (android_device.AndroidDevice)
-    sl4a or s (SL4A)
-    ed (EventDispatcher)
-    """.format(serial)
+        return """Connected to {}. Call methods against:
```
adorokhine: Can you run this and make sure it looks the same? I'm pretty sure the subsequent lines will end up hugely indented in the output this way.

tools/snippet_shell.py
```diff
@@ -45,24 +45,24 @@ def _start_services(self, console_env):
...
-    logging.basicConfig(level=logging.INFO)
-    SnippetShell(argv[0]).main(args.serial)
+        '-s',
+        '--serial',
+        help='Device serial to connect to (if more than one device is connected)'
+    )
+    parser.add_argument('package', metavar='PACKAGE_NAME', type=str, nargs=1,
+                        help='The pacakge name of the snippet to use.')
+    args = parser.parse_args()
+    logger.setup_test_logger('/tmp/logs/mobly_snippet_shell',
```
adorokhine: Just curious why you want this?
Do we really need to save the logs from interactive sessions in a hardcoded path like this?

tools/snippet_shell.py
```diff
@@ -45,24 +45,24 @@ def _start_services(self, console_env):
...
-        print('ERROR: Snippet package not specified on command line',
-              file=sys.stderr)
-        sys.exit(1)
-    logging.basicConfig(level=logging.INFO)
-    SnippetShell(argv[0]).main(args.serial)
+        '-s',
+        '--serial',
+        help='Device serial to connect to (if more than one device is connected)'
+    )
+    parser.add_argument('package', metavar='PACKAGE_NAME', type=str, nargs=1,
```
adorokhine: What was wrong with this as a positional arg?

xpconanfan: This is still a positional arg.
This is how argparse handles positional args.

mobly/controllers/android_device_lib/snippet_client.py
xpconanfan: Done

tools/sl4a_shell.py
xpconanfan: Done

tools/snippet_shell.py
xpconanfan: It shows a nice [SnippetShell] prefix with timestamp in log lines, instead of DEBUG::Xxx:: xyz.
It also helps to have a file tracing what happened especially if people want to use logging in the shell and do stuff?
But we don't have to make this change. No strong preference.

xpconanfan: With mobly logger:
[SnippetShell] 12-22 16:04:22.153 INFO xxx
Without:
INFO:root:

Maybe we should keep the former?...

tools/snippet_shell.py
```diff
@@ -45,24 +45,24 @@ def _start_services(self, console_env):
...
+                        help='The pacakge name of the snippet to use.')
```
dthkao: typo

xpconanfan: Done

xpconanfan:
Ignore travis here.
The failure is fixed in master.

mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -53,6 +53,9 @@ def __init__(self, package, port, adb_proxy):
     def _do_start_app(self):
         """Overrides superclass."""
         cmd = _LAUNCH_CMD.format(self._port, self._package)
+        # Use info here so people know exactly what's happening here, which is
+        # helpful since they need to create their own instrumentations and
+        # anifest.
```
adorokhine: manifest

tools/sl4a_shell.py
adorokhine: What does Done mean? Won't it be hugely indented?

tools/snippet_shell.py
```diff
@@ -45,24 +45,24 @@ def _start_services(self, console_env):
...
-Connected to {}. Call methods against:
-    ad (android_device.AndroidDevice)
-    snippet or s (Snippet)
-    """.format(serial)
-
+        lines = ['Connected to %s.' % serial,
+                 'Call methods against:',
+                 '    ad (android_device.AndroidDevice)',
+                 '    snippet or s (Snippet)']
+        return '\n'.join(lines)
```
adorokhine: Is this really clearer to read? What was wrong with the old way? Also the new line breaks are different; is that intentional?

tools/snippet_shell.py
adorokhine: I'm ok with the new log format but dumping the log of an interactive session to a file, especially at a hardcoded path like this, seems strange and unnecessary.

tools/snippet_shell.py
xpconanfan: Yes, this way the code doesn't have to have weird indentation that doesn't follow the regular style just to get around space issues in string; and it's still pretty easy to figure out what the final string looks like.

I changed the breaks a little bit. Can change it back if you prefer the old way

tools/sl4a_shell.py
xpconanfan: I changed it, as shown below.

mobly/controllers/android_device_lib/snippet_client.py
xpconanfan: Done

tools/sl4a_shell.py
adorokhine: I think you only changed the one in snippet shell. The one in sl4a_shell is still unchanged from the first patch, and now they're different.

xpconanfan: Done

PR #61, opened 2016-12-22
Improve snippet client management.

### LINKED ISSUE #55: Snippet client reference is not removed from ad object after teardown
'load_snippet' sets a snippet client reference attribute on the AndroidDevice object.
We should remove the attribute after snippet is stopped.

### LINKED ISSUE #56: load_snippet allows registering different packages under the same alias.
We should probably throw an error if attempts to register a different package under an existing alias is made?
```
ad.load_snippet('snippet', 'my.package')
ad.load_snippet('snippet', 'my.other.package') # Raise error
```

xpconanfan:
* Better handling of attempts to load snippet with duplicate info.
* Remove client attributes in stop_services.
* Add unit tests
fixes #55 
fixes #56

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -40,6 +40,9 @@ ANDROID_DEVICE_ADB_LOGCAT_PARAM_KEY = "adb_logcat_param"
 ANDROID_DEVICE_EMPTY_CONFIG_MSG = "Configuration is empty, abort!"
 ANDROID_DEVICE_NOT_LIST_CONFIG_MSG = "Configuration should be a list, abort!"
 
+SNIPPET_KEY_CLIENT = 'client'
+SNIPPET_KEY_PACKAGE = 'name'
+
 # Keys for attributes in configs that alternate device behavior
 KEY_SKIP_SL4A = "skip_sl4a"
 KEY_DEVICE_REQUIRED = "required"
@@ -47,6 +50,10 @@ KEY_DEVICE_REQUIRED = "required"
 
 class Error(signals.ControllerError):
     pass
+
+
+class SnippetError(signals.ControllerError):
+    """Raised when somethig wrong with snippet happens."""
 
 
 class DoesNotExistError(Error):
@@ -366,7 +373,8 @@ class AndroidDevice(object):
         self.fastboot = fastboot.FastbootProxy(serial)
         if not self.is_bootloader and self.is_rootable:
             self.root_adb()
-        self._snippet_clients = []
+        # A dict for tracking snippet clients
+        self._snippet_clients = {}
 
     def set_logger_prefix_tag(self, tag):
         """Set a tag for the log line prefix of this instance.
@@ -413,9 +421,11 @@ class AndroidDevice(object):
         if self._adb_logcat_process:
             self.stop_adb_logcat()
         self._terminate_sl4a()
-        for client in self._snippet_clients:
+        for name in list(self._snippet_clients.keys()):
+            client = self._snippet_clients[name][SNIPPET_KEY_CLIENT]
             self._terminate_jsonrpc_client(client)
-        self._snippet_clients = []
+            delattr(self, name)
+        self._snippet_clients = {}
 
     @property
     def build_info(self):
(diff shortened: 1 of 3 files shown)


mobly/controllers/android_device.py
```diff
@@ -509,26 +518,49 @@ def root_adb(self):
...
+                raise SnippetError(
+                    'Snippet package "%s" has already been loaded under name "%s".'
+                    % (package, client_name))
         host_port = utils.get_available_host_port()
         # TODO(adorokhine): Don't assume that a free host-side port is free on
         # the device as well. Both sides should allocate a unique port.
         device_port = host_port
         client = snippet_client.SnippetClient(
             package=package, port=host_port, adb_proxy=self.adb)
         self._start_jsonrpc_client(client, host_port, device_port)
```
dthkao: why a nested dict? we could access client.app_name.

xpconanfan: Ah, didn't know that

xpconanfan: Done

mobly/controllers/android_device.py
```diff
@@ -413,9 +418,11 @@ def stop_services(self):
         if self._adb_logcat_process:
             self.stop_adb_logcat()
         self._terminate_sl4a()
-        for client in self._snippet_clients:
+        for name in list(self._snippet_clients.keys()):
+            client = self._snippet_clients[name]
```
adorokhine: Seems like you should be using items() here?

xpconanfan: Done

mobly/controllers/android_device.py
```diff
@@ -366,7 +370,8 @@ def __init__(self, serial=""):
         self.fastboot = fastboot.FastbootProxy(serial)
         if not self.is_bootloader and self.is_rootable:
             self.root_adb()
-        self._snippet_clients = []
+        # A dict for tracking snippet clients
```
dthkao: Mind adding a comment on the structure? {attr_name : client}

xpconanfan: Done

mobly/controllers/android_device.py
```diff
@@ -509,26 +514,46 @@ def root_adb(self):
...
+
+        Raises:
+            SnippetError is raised if illegal load operations are attempted.
         """
+        # Should not load snippet with the same attribute more than once.
+        if name in self._snippet_clients:
+            raise SnippetError(
+                ('Attribute "%s" is already registered with '
+                 'package "%s", it cannot be used again.') % (
+                     name, self._snippet_clients[name].app_name))
```
dthkao: Just realized that app_name can (and is) overwritten, which breas usage as a comparison to package. sl4a uses "SL4A".

Any thoughts on making client._package public?

xpconanfan: I don't understand the concern, can you clarify

dthkao: Sorry this should apply to the if package==client.app_name below. Is it possible someone could reload sl4a using load_snippet()?

xpconanfan: Nothing stops them from doing so, they just won't get ed started properly.
Are you saying we should guard against it?

dthkao: Disregard. Sl4a wouldn't load as a snippet anyway.

mobly/controllers/android_device.py
```diff
@@ -509,26 +515,46 @@ def root_adb(self):
...
+        # Should not load snippet with an existing attribute.
+        if hasattr(self, name):
+            raise SnippetError(
+                'Attribute "%s" already exists, please use a different name.' %
+                name)
+        # Should not load the same snippet package more than once.
+        for client_name, client in self._snippet_clients.items():
+            if package == client.app_name:
```
adorokhine: You're making the assumption here that the snippet's app_name will always equal the package. That's not really guaranteed; it's an arbitrary string intended for user debug messages.

Do you want to expose client._package as client.package and assert on that instead?

xpconanfan: I see.

xpconanfan: Done.
Great use case of @property :P

Issue #57, opened 2016-12-22
Fix test logger_test.test_epoch_to_log_line_timestamp


xpconanfan:
This test only works in PST timezone, which is not fit for continuous integration.
It needs to be fixed to be timezone agnostic.

xpconanfan:
To reprod, change your system time to a timezone that's not PST and run this test:
```
$ cd <mobly>/tests/mobly
$ python logger_test.py
```

PR #62, opened 2016-12-23
Update time zone improvement

ygw11223:
fixed it to be timezone agnostic

### DIFF
--- a/mobly/logger.py
+++ b/mobly/logger.py
@@ -90,15 +90,16 @@ def epoch_to_log_line_timestamp(epoch_time, time_zone=None):
 
     Args:
         epoch_time: integer, an epoch timestamp in ms.
-        time_zone: instance of tzinfo, time zone information. Using pytz rather than
-        python 3.2 time_zone implementation for python 2 compatibility reasons
+        time_zone: instance of tzinfo, time zone information. 
+                   Using pytz rather than python 3.2 time_zone implementation
+                   for python 2 compatibility reasons.
 
     Returns:
         A string that is the corresponding timestamp in log line timestamp
         format.
     """
     s, ms = divmod(epoch_time, 1000)
-    d = datetime.datetime.fromtimestamp(s, tz = time_zone)
+    d = datetime.datetime.fromtimestamp(s, tz=time_zone)
     return d.strftime("%m-%d %H:%M:%S.") + str(ms)
 
 
--- a/mobly/logger.py
+++ b/mobly/logger.py
@@ -84,7 +84,7 @@ def _get_timestamp(time_format, delta=None):
     return t.strftime(time_format)[:-3]
 
 
-def epoch_to_log_line_timestamp(epoch_time, time_zone):
+def epoch_to_log_line_timestamp(epoch_time, time_zone=None):
     """Converts an epoch timestamp in ms to log line timestamp format, which
     is readible for humans.
 
--- a/mobly/logger.py
+++ b/mobly/logger.py
@@ -90,7 +90,7 @@ def epoch_to_log_line_timestamp(epoch_time, time_zone=None):
 
     Args:
         epoch_time: integer, an epoch timestamp in ms.
-        time_zone: instance of tzinfo, time zone information. 
+        time_zone: instance of tzinfo, time zone information.
                    Using pytz rather than python 3.2 time_zone implementation
                    for python 2 compatibility reasons.
 
--- a/mobly/logger.py
+++ b/mobly/logger.py
@@ -84,19 +84,21 @@ def _get_timestamp(time_format, delta=None):
     return t.strftime(time_format)[:-3]
 
 
-def epoch_to_log_line_timestamp(epoch_time):
+def epoch_to_log_line_timestamp(epoch_time, time_zone):
     """Converts an epoch timestamp in ms to log line timestamp format, which
     is readible for humans.
 
     Args:
         epoch_time: integer, an epoch timestamp in ms.
+        time_zone: instance of tzinfo, time zone information. Using pytz rather than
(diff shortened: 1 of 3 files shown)


ygw11223:
I signed CLA.

tests/mobly/logger_test.py
```diff
@@ -23,8 +24,10 @@ class LoggerTest(unittest.TestCase):
     """Verifies code in mobly.logger module.
     """
     def test_epoch_to_log_line_timestamp(self):
-        # actual_stamp = logger.epoch_to_log_line_timestamp(1469134262116)
-        # self.assertEqual("07-21 13:51:02.116", actual_stamp)
+        test_time = time.localtime(1469134262)
+        correct_stamp = time.strftime("%m-%d %H:%M:%S.116", test_time)
```
adorokhine: Seems strange to encode '116' right into the format string. Stack Overflow suggests to get milliseconds into the formatted string like so:
http://stackoverflow.com/a/18406412

Perhaps you could use this instead?

tests/mobly/logger_test.py
```diff
@@ -23,8 +24,12 @@ class LoggerTest(unittest.TestCase):
     """Verifies code in mobly.logger module.
     """
     def test_epoch_to_log_line_timestamp(self):
-        # actual_stamp = logger.epoch_to_log_line_timestamp(1469134262116)
-        # self.assertEqual("07-21 13:51:02.116", actual_stamp)
+        correct_stamp = (
+            datetime.fromtimestamp(1469134262.1161)
```
adorokhine: Why the extra '1' at the end versus the other timestamp? Typo?

adorokhine: Just saw the PR comment. It definitely needs a code comment at least. 

epoch_to_log_line_timestamp has the same issue with py3.4, and handles this by removing the milliseconds, rounding them manually, and appending them onto the end of the formatted string. You could do the same thing, but as this test becomes more flexible, it is starting to duplicate the logic of the method under test. Since it's a formatting method, I do think comparing to a hardcoded string in the right format is ultimately the right thing to do here.

fromtimestamp takes an optional timezone object. Maybe the right thing to do is to expose tz in epoch_to_log_line_timestamp (default None) and supply UTC in the unit test. Then we can do the entire test explicitly in UTC and compare to a hardcoded string.

That way we get timezone independence, but don't have a 'test' that just duplicates the formatting logic of the method under test.

What do you think?

dthkao: +1 to this.

xpconanfan:
Please also make sure you have 'fixes #57' in your final commit message when you do 'Squash and merge'

tests/mobly/logger_test.py
```diff
@@ -23,8 +24,12 @@ class LoggerTest(unittest.TestCase):
...
+            .strftime("%m-%d %H:%M:%S.%f")[:-3]
+        )
+        actual_stamp = logger.epoch_to_log_line_timestamp(1469134262116)
+        self.assertEqual(correct_stamp, actual_stamp)
         pass
```
xpconanfan: Remove 'pass'

mobly/logger.py
```diff
@@ -84,19 +84,21 @@ def _get_timestamp(time_format, delta=None):
...
 
 
-def epoch_to_log_line_timestamp(epoch_time):
+def epoch_to_log_line_timestamp(epoch_time, time_zone=None):
     """Converts an epoch timestamp in ms to log line timestamp format, which
     is readible for humans.
 
     Args:
         epoch_time: integer, an epoch timestamp in ms.
+        time_zone: instance of tzinfo, time zone information. Using pytz rather than
```
xpconanfan: Line length limit is 80 chars.
This line is too long.

mobly/logger.py
```diff
@@ -84,19 +84,21 @@ def _get_timestamp(time_format, delta=None):
...
+        python 3.2 time_zone implementation for python 2 compatibility reasons
```
xpconanfan: Indent subsequent lines to a space after ':'

mobly/logger.py
```diff
@@ -84,19 +84,21 @@ def _get_timestamp(time_format, delta=None):
...
 
     Returns:
         A string that is the corresponding timestamp in log line timestamp
         format.
     """
     s, ms = divmod(epoch_time, 1000)
-    d = datetime.datetime.fromtimestamp(s)
+    d = datetime.datetime.fromtimestamp(s, tz = time_zone)
```
xpconanfan: no extra spaces.
`tz=time_zone`

mobly/logger.py
xpconanfan: Also missing the ending period

mobly/logger.py
ygw11223: Done

tests/mobly/logger_test.py
```diff
@@ -23,9 +24,9 @@ class LoggerTest(unittest.TestCase):
     """Verifies code in mobly.logger module.
     """
     def test_epoch_to_log_line_timestamp(self):
-        # actual_stamp = logger.epoch_to_log_line_timestamp(1469134262116)
-        # self.assertEqual("07-21 13:51:02.116", actual_stamp)
-        pass
+        actual_stamp = logger.epoch_to_log_line_timestamp(1469134262116,
```
xpconanfan: It looks like there's a trailing space at the end of this line.
Please remove.
You can use linter _yapf_ to take care of formatting.
`yapf -i --style pep8 <file path>

ygw11223: Done. Thank you.

xpconanfan:
fixes #57

PR #63, opened 2016-12-27
Fix a bug in AndroidDevice.reboot

xpconanfan:
AndroidDevice.reboot tries to start sl4a even if skip_sl4a is set to True.

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -132,7 +132,7 @@ def _start_services_on_ads(ads):
     for ad in ads:
         running_ads.append(ad)
         try:
-            ad.start_services(getattr(ad, KEY_SKIP_SL4A, False))
+            ad.start_services(skip_sl4a=getattr(ad, KEY_SKIP_SL4A, False))
         except Exception as e:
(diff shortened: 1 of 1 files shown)


mobly/controllers/android_device.py
```diff
@@ -408,6 +408,8 @@ def start_services(self, skip_sl4a=False):
         except:
             self.log.exception("Failed to start adb logcat!")
             raise
+        if skip_sl4a is None:
+            skip_sl4a=getattr(ad, KEY_SKIP_SL4A, False)
```
dthkao: Why not just fetch it here and remove it as a param?

xpconanfan: What's the benefit in removing it?
Also, start_services could be called multiple times in the lifetime of ad.

dthkao: I meant remove it as a function arg. Is start_services ever called when the device attribute is True but we pass in False?

xpconanfan: I see.
Now I realize that it is a bit odd to affect the function's behavior by setting attribute on an object...

PR #69, opened 2016-12-28
Make ad.is_rootable more robust.

### LINKED ISSUE #66: ad.is_rootable fails to identify the build properly if adb server isn't running already
There is an edge case where `ad.is_rootable` incorrectly returns 'True', in which case we issue `adb root` call, causing some NBU devices to stop responding to adb commands.

The edge case is: adb server is not running prior to starting the test.
In this situation, the following steps occur:

1. Starting the test
2. `AndroidDevice` controller obj issues adb call to determine build type
3. adb server gets started, printing extra strings to stdout
4. is_rootable tries to match the string 'user' to the stdout, which evaluates to 'False' due to the extra output.
5. The user build device is deemed 'rootable'.

Sample output from such a case:
```
# Beginning of the test
$ adb -s <serial> shell getprop ro.build.type
# Results in the following stdout:
# * daemon not running. starting it now on port 5037 *
# * daemon started successfully *
# user^M
```

`ad.adb.getprop` is the first adb call made during the lifecycle of an `AndroidDevice` obj, so it's critical to handle this check correctly.
Other adb-based check may also fail if adb server somehow gets restarted and stdout is polluted. As a less urgent item, we should consider making checks based on adb's stdout more robust.

xpconanfan:
Also remove a duplicate 'if' statement.

fixes #66

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -463,8 +463,15 @@ class AndroidDevice(object):
 
     @property
     def is_rootable(self):
-        build_type = self.adb.getprop("ro.build.type").lower()
-        return build_type != 'user'
+        """If the build type is 'user', the device is not rootable.
+
+        Other possible build types are 'userdebug' and 'eng', both are rootable.
+        We are checking the last four chars of the clean stdout because the
+        stdout of the adb command could be polluted with other info like adb
+        server startup message.
+        """
+        build_type_output = self.adb.getprop("ro.build.type").lower()
+        return build_type_output[-4:] != 'user'
 
     @property
     def model(self):
--- a/tests/lib/mock_android_device.py
+++ b/tests/lib/mock_android_device.py
@@ -75,8 +75,6 @@ class MockAdbProxy(object):
             return "FakeModel"
         elif params == "sys.boot_completed":
             return "1"
-        elif params == "ro.build.type":
-            return "userdebug"
 
     def bugreport(self, params):
         expected = os.path.join(logging.log_path,


xpconanfan:
Sure, we can do something like a 'before any adb cmd' check.

But it doesn't necessarily solve the issue well since:
1. adb server could die at any point of the execution and this error could pop up again.
2. Users do not necessarily follow the same execution order. We are looking at AndroidDevice class only, but what about the other functions that use adb proxy in different ways, e.g. list_occupied_adb_ports.

It is still good to guard each adb cmd as much as possible in addition to a check before any adb calls.
As for adding a mechanism that is guaranteed to check adb server status before any adb call is issued, I think it's a nice-to-have thing. I'm not against it, just don't think it's something urgent.

adorokhine:
So does that mean we will be modifying every adb command in ACTS with the same logic? If so can we build it into the shell() wrapper to avoid every test writer having to implement this check?

xpconanfan:
Just saying we need to be smarter in handling outputs from adb, and there may not be a one-fit-all solution for some adb problems.
Server starting msg is one thing, there could be other junk that adb throws into stdout/stderr, which adb also misuses often.

I'm not sure which check you're referring to.
I did not add an extra 'check' in this PR; I simply made the existing condition more explicit.
I think for the adb APIs we provide (such as is_rootable), we need to make them as reliable as possible. But for direct adb calls, we should leave it as authentic as possible. What if some user intentionally wants some adb output that are not valuable to our tests, but valuable to theirs? (one person's trash could be another person's treasure). It would be inappropriate for us to filter out info from direct AdbProxy calls.

PR #75, opened 2017-01-06
Improve a warning message in TestRunner.

xpconanfan:
Also add another type to gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ Gemfile.lock
 _site
 build/
 dist/
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,7 @@
-/.cache/
+build/
+.cache/
+dist/
+.DS_Store
 *.egg-info/
 *.egg/
 *.eggs/
@@ -7,6 +10,3 @@ Gemfile.lock
 *.pyc
 .sass-cache
 _site
-build/
-dist/
-.DS_Store
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -355,8 +355,8 @@ class TestRunner(object):
             self.results.add_controller_info(module_config_name,
                                              controller_info)
         else:
-            self.log.warning(("No optional debug info found for controller %s, "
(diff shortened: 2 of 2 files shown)


mobly/test_runner.py
```diff
@@ -355,8 +355,8 @@ def get_info(objects):
             self.results.add_controller_info(module_config_name,
                                              controller_info)
         else:
-            self.log.warning("No controller info obtained for %s",
-                             module_config_name)
+            self.log.warning(("No optional debug info found for controller %s, "
+                              "carry on."), module_config_name)
```
adorokhine: Why the extra set of brackets?

.gitignore
```diff
@@ -9,3 +9,4 @@ Gemfile.lock
 _site
 build/
 dist/
+.DS_Store
```
adorokhine: Mind putting this up top to keep this generally in alphabetical order?

mobly/test_runner.py
```diff
@@ -355,8 +355,8 @@ def get_info(objects):
             self.results.add_controller_info(module_config_name,
                                              controller_info)
         else:
-            self.log.warning("No controller info obtained for %s",
-                             module_config_name)
+            self.log.warning(("No optional debug info found for controller %s, "
```
adorokhine: This still isn't that clear. Can you mention that the controller module doesn't implement get_info?

mobly/test_runner.py
xpconanfan: Isn't it needed to make the string break work?

.gitignore
```diff
@@ -1,4 +1,7 @@
-/.cache/
+build/
+.cache/
```
adorokhine: I believe periods sort before letters right?

xpconanfan: It is not how the original file sorted:

> *.eggs/
> Gemfile.lock
> .jekyll-metadata

And the official github python gitignore doesn't seem to follow any order.
https://github.com/github/gitignore/blob/master/Python.gitignore

PR #76, opened 2017-01-07
String formatting clean up.

xpconanfan:
* Use single quote by default.
* Remove usage of '.format'.

### DIFF
--- a/mobly/config_parser.py
+++ b/mobly/config_parser.py
@@ -37,7 +37,7 @@ def _validate_test_config(test_config):
     """
     for k in keys.Config.reserved_keys.value:
         if k not in test_config:
-            raise MoblyConfigError("Required key %s missing in test config." % k)
+            raise MoblyConfigError('Required key %s missing in test config.' % k)
 
 
 def _validate_testbed_name(name):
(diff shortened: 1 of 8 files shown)


mobly/config_parser.py
```diff
@@ -80,15 +80,15 @@ def _validate_testbed_configs(testbed_configs):
         _validate_testbed_name(name)
         # Test bed names should be unique.
         if name in seen_names:
-            raise MoblyConfigError("Duplicate testbed name %s found." % name)
+            raise MoblyConfigError('Duplicate testbed name %s found.' % name)
         seen_names.add(name)
 
 
 def _verify_test_class_name(test_cls_name):
```
adorokhine: Is this method still used? I thought it was part of the old autodiscover codepath. (please don't update this in this PR though)

xpconanfan: Do we want to enforce test class naming convention going forward?
Or do we just want to rely on class inheritance relationship

adorokhine: I don't see any reason to enforce a test class naming convention. Any advantages?

xpconanfan: Yeah, not much advantage other than making it easier to id which one is test for humans.
Example:
```
class Hahaha(Hehehe):
...
class Hehehe(base_test.BaseTestClass):
...
```
Not super critical

PR #79, opened 2017-01-19
Restore snippet clients after reboot in ad.reboot.

### LINKED ISSUE #64: Properly handle snippet in ad.reboot
Right now AndroidDevice.reboot does not restore any snippet.
It should reload the snippet loaded before reboot and make their aliases available.

xpconanfan:
Fixes #64

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -795,6 +795,20 @@ class AndroidDevice(object):
             time.sleep(5)
         raise Error('Device %s booting process timed out.' % self.serial)
 
+    def _get_active_snippet_info(self):
+        """Collects information on currently active snippet clients.
+
+        The info is used for restoring the snippet clients after rebooting the
+        device.
+
+        Returns:
+            A list of tuples, each tuple's first element is the 
+        """
+        snippet_info = []
+        for attr_name, client in self._snippet_clients.items():
+            snippet_info.append((attr_name, client.package))
+        return snippet_info
+
     def reboot(self):
         """Reboots the device.
 
@@ -812,12 +826,16 @@ class AndroidDevice(object):
         if self.is_bootloader:
             self.fastboot.reboot()
             return
+        snippet_info = self._get_active_snippet_info()
         self.stop_services()
         self.adb.reboot()
         self.wait_for_boot_completion()
         if self.is_rootable:
             self.root_adb()
-        self.start_services(skip_sl4a=getattr(self, KEY_SKIP_SL4A, False))
+        skip_sl4a = getattr(self, KEY_SKIP_SL4A, False) or self.sl4a is None
+        self.start_services(skip_sl4a=skip_sl4a)
+        for attr_name, package_name in snippet_info:
+            self.load_snippet(attr_name, package_name)
 
 
 class AndroidDeviceLoggerAdapter(logging.LoggerAdapter):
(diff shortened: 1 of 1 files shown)


mobly/controllers/android_device.py
```diff
@@ -812,12 +826,16 @@ def reboot(self):
         if self.is_bootloader:
             self.fastboot.reboot()
             return
+        snippet_info = self._get_active_snippet_info()
```
adorokhine: Just curious why you have to capture these? Isn't it enough to iterate through _snippet_clients after reboot and reconnect them?

mobly/controllers/android_device.py
```diff
@@ -812,12 +826,16 @@ def reboot(self):
...
         self.stop_services()
         self.adb.reboot()
         self.wait_for_boot_completion()
         if self.is_rootable:
             self.root_adb()
-        self.start_services(skip_sl4a=getattr(self, KEY_SKIP_SL4A, False))
+        skip_sl4a = getattr(self, KEY_SKIP_SL4A, False) or self.sl4a is None
```
adorokhine: Curious what was wrong with the old behaviour? Under what circumstances would skip_sl4a be true but self.sl4a be None?

mobly/controllers/android_device.py
xpconanfan: Because stop_services clears those info as intended.

mobly/controllers/android_device.py
xpconanfan: sl4a is required by default is an incorrect assumption as our usage is expanded beyond Nexus devices. In the open source world or NBU use cases, sl4a is not there by default. Also in console mode, sl4a is not started by default either. So overall this is a better check than the original one. The old logic causes errors in reboot in python console if sl4a is not installed on the device at all.

We might want to consider making sl4a a service that needs to be explicitly requested like a snippet as opposed to assuming it's there by default.

dthkao: +1 to separate services (and an api for adding/requesting them).

xpconanfan: Can file a separate bug for that. But i'll require all internal tests that use sl4a to explicitly request sl4a in test bed config.

mobly/controllers/android_device.py
```diff
@@ -795,6 +795,20 @@ def wait_for_boot_completion(self):
...
         raise Error('Device %s booting process timed out.' % self.serial)
 
+    def _get_active_snippet_info(self):
+        """Collects information on currently active snippet clients.
+
+        The info is used for restoring the snippet clients after rebooting the
+        device.
+
+        Returns:
+            A list of tuples, each tuple's first element is the
```
adorokhine: "first element is the" ?

mobly/controllers/android_device.py
adorokhine: a) Do we need both conditions?
b) Does "self.sl4a is None" work if sl4a has been skipped? Wouldn't it just cause a crash saying unknown attribute sl4a?

xpconanfan: We need both since teardown sl4a will clear self.sl4a
self.sl4a is set to None in constructor.

mobly/controllers/android_device.py
xpconanfan: Done

PR #90, opened 2017-01-27
Rework snippet and sl4a startup process.

### LINKED ISSUE #88: Snippet client's _is_app_running() doesn't work properly, leading to race conditions
Slow loading snippets fail because _is_app_running() returns True even when nothing is on the other side of the socket.
This leads to race conditions between the snippet app really starting and the connection attempt.

```
>>> sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
>>> print self._port
8822
>>> result = sock.connect_ex(('127.0.0.1', self._port))
>>> print result
0
And yet...
$ nc 127.0.0.1 8822
$
```

adorokhine:
_is_app_running() didn't work before because the socket was 'live' as soon
as it was bound through adb, even though nothing had started on the other
end. This led to race conditions between the snippet app starting and the
client connecting.

I couldn't find a way to make sure something was alive on the other end
other than actually sending data through the socket and getting a response,
so I've changed the is_app_running() process to actually call connect().
Tested with both snippet and sl4a.

Fixes #88.

### DIFF
--- a/mobly/controllers/android_device_lib/jsonrpc_client_base.py
+++ b/mobly/controllers/android_device_lib/jsonrpc_client_base.py
@@ -41,7 +41,6 @@ from builtins import str
 import json
 import logging
 import socket
-import sys
 import threading
 import time
 
@@ -94,15 +93,21 @@ class JsonRpcClientBase(object):
     of communication.
 
     Attributes:
+        host_port: (int) The host port of this RPC client.
+        device_port: (int) The device port of this RPC client. Must be set by
+          the superclass.
         uid: (int) The uid of this session.
         app_name: (str) The user-visible name of the app being communicated
           with. Must be set by the superclass.
     """
-    def __init__(self, adb_proxy):
+
+    def __init__(self, host_port, adb_proxy):
         """
         Args:
+            port: (int) the host-side port this server will listen on
             adb_proxy: adb.AdbProxy, The adb proxy to use to start the app
         """
+        self.host_port = host_port
         self.uid = None
         self._adb = adb_proxy
         self._client = None  # prevent close errors on connect failure
@@ -139,16 +144,6 @@ class JsonRpcClientBase(object):
         """
         raise NotImplementedError()
 
-    def _is_app_running(self):
-        """Checks if the app is currently running on an android device.
-
-        Must be implemented by subclasses.
-
-        Returns:
-            True if the app is running, False otherwise.
-        """
-        raise NotImplementedError()
-
     # Rest of the client methods.
 
     def check_app_installed(self):
@@ -175,8 +170,7 @@ class JsonRpcClientBase(object):
         raise AppStartError(
             '%s failed to start on %s.' % (self.app_name, self._adb.serial))
 
-    def connect(self, port, addr='localhost', uid=UNKNOWN_UID,
-                cmd=JsonRpcCommand.INIT):
+    def connect(self, uid=UNKNOWN_UID, cmd=JsonRpcCommand.INIT):
         """Opens a connection to a JSON RPC server.
 
         Opens a connection to a remote client. The connection attempt will time
@@ -185,12 +179,8 @@ class JsonRpcClientBase(object):
         as well.
 
         Args:
-            port: int, The port this client should connect to.
-            addr: str, The address this client should connect to.
             uid: int, The uid of the session to join, or UNKNOWN_UID to start a
                  new session.
-            connection_timeout: int, The time to wait for the connection to come
-                 up.
             cmd: JsonRpcCommand, The command to use for creating the connection.
 
         Raises:
@@ -200,13 +190,12 @@ class JsonRpcClientBase(object):
         """
         self._counter = self._id_counter()
         try:
-            self._conn = socket.create_connection((addr, port),
+            self._conn = socket.create_connection(('127.0.0.1', self.host_port),
                                                   _SOCKET_TIMEOUT)
             self._conn.settimeout(_SOCKET_TIMEOUT)
         except (socket.timeout, socket.error, IOError):
             logging.exception("Failed to create socket connection!")
             raise
-        self.port = port
         self._client = self._conn.makefile(mode="brw")
 
         resp = self._cmd(cmd, uid)
@@ -270,6 +259,20 @@ class JsonRpcClientBase(object):
         if result['id'] != apiid:
             raise ProtocolError(ProtocolError.MISMATCHED_API_ID)
         return result['result']
+
+    def _is_app_running(self):
+        """Checks if the app is currently running on an android device.
+
+        May be overridden by subclasses with custom sanity checks.
+        """
+        running = False
+        try:
+          self.connect()
+          running = True
+        finally:
+          self.close()
+          # This 'return' squashes exceptions fron connect()
+          return running
 
     def __getattr__(self, name):
         """Wrapper for python magic to turn method calls into RPC calls."""
--- a/mobly/controllers/android_device_lib/jsonrpc_client_base.py
+++ b/mobly/controllers/android_device_lib/jsonrpc_client_base.py
@@ -94,20 +94,24 @@ class JsonRpcClientBase(object):
 
     Attributes:
         host_port: (int) The host port of this RPC client.
-        device_port: (int) The device port of this RPC client. Must be set by
-          the superclass.
-        uid: (int) The uid of this session.
+        device_port: (int) The device port of this RPC client.
         app_name: (str) The user-visible name of the app being communicated
-          with. Must be set by the superclass.
+                  with.
+        uid: (int) The uid of this session.
     """
 
-    def __init__(self, host_port, adb_proxy):
+    def __init__(self, host_port, device_port, app_name, adb_proxy):
         """
         Args:
-            port: (int) the host-side port this server will listen on
-            adb_proxy: adb.AdbProxy, The adb proxy to use to start the app
+            host_port: (int) The host port of this RPC client.
+            device_port: (int) The device port of this RPC client.
+            app_name: (str) The user-visible name of the app being communicated
+                      with.
+            adb_proxy: (adb.AdbProxy) The adb proxy to use to start the app.
         """
         self.host_port = host_port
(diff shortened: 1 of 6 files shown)


mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -94,15 +93,21 @@ class JsonRpcClientBase(object):
     of communication.
 
     Attributes:
+        host_port: (int) The host port of this RPC client.
+        device_port: (int) The device port of this RPC client. Must be set by
+          the superclass.
```
xpconanfan: i'm actually not sure what 'Must be set by the superclass' mean here.
Isn't this the superclass? And I don't see a way to set this. Same with app_name..

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -200,13 +190,12 @@ def connect(self, port, addr='localhost', uid=UNKNOWN_UID,
         """
         self._counter = self._id_counter()
         try:
-            self._conn = socket.create_connection((addr, port),
+            self._conn = socket.create_connection(('127.0.0.1', self.host_port),
```
xpconanfan: why are we changing to this representation of localhost? As it turns out mac and linux interprets these two with subtle differences. Just curious.

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -271,6 +260,20 @@ def _rpc(self, method, *args):
             raise ProtocolError(ProtocolError.MISMATCHED_API_ID)
         return result['result']
 
+    def _is_app_running(self):
```
xpconanfan: The name of this function kinda indicates that we could call this at any point to check if the app  is running. Is this true? Would attempting another connection when there's an ongoing connection mess things up?

mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -32,42 +31,40 @@ class Error(Exception):
...
+    def __init__(self, package, host_port, adb_proxy):
         """Initialzies a SnippetClient.
   
         Args:
           package: (str) The package name of the apk where the snippets are
             defined.
-          port: (int) The port at which to start the snippet client. Note that
-            the same port will currently be used for both the device and host
-            side of the connection.
+          host_port: (int) The port at which to start the snippet client. Note
```
xpconanfan: can we indent the comments to after ':'?

mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -32,42 +31,40 @@ class Error(Exception):
 
 
 class SnippetClient(jsonrpc_client_base.JsonRpcClientBase):
-    def __init__(self, package, port, adb_proxy):
+    def __init__(self, package, host_port, adb_proxy):
         """Initialzies a SnippetClient.
   
         Args:
```
xpconanfan: should these be attributes instead of being in Args docstring?

mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -32,42 +31,40 @@ class Error(Exception):
...
-        self._package = package
-        self._port = port
-
-    @property
-    def package(self):
-        return self._package
+        # TODO(adorokhine): Don't assume that a free host-side port is free on
+        # the device as well. Both sides should allocate a unique port.
+        self.device_port = host_port
+        self.package = package
```
xpconanfan: Do we want to prevent self.package from being changed by making it a @property?

mobly/controllers/android_device_lib/jsonrpc_client_base.py
adorokhine: Sorry, I meant subclass. I've simplified this and just passed this stuff into the constructor.

mobly/controllers/android_device_lib/jsonrpc_client_base.py
adorokhine: Just trying to skip an extra dns lookup... curious what the difference is between mac and linux?

mobly/controllers/android_device_lib/jsonrpc_client_base.py
adorokhine: Yeah it would mess things up, since it would disconnect _conn. It's a private function so hopefully it won't be called outside of the expected lifecycle but I can wrap it in a check for self._conn == null if you like. lmk what you want to see.

mobly/controllers/android_device_lib/snippet_client.py
adorokhine: They should be in both... updated

mobly/controllers/android_device_lib/snippet_client.py
adorokhine: Done.

mobly/controllers/android_device_lib/snippet_client.py
adorokhine: This is python; anything can be changed...

properties are not really used for access control. They are intended for when a value has nontrivial computation, needs to be cached, etc. 

Here we don't need to compute this value since we already have it, so it should just be a simple public attribute.

http://stackoverflow.com/questions/1554546/when-and-how-to-use-the-builtin-function-property-in-python
http://blaag.haard.se/What-s-the-point-of-properties-in-Python/

mobly/controllers/android_device_lib/jsonrpc_client_base.py
xpconanfan: Should we introduce an 'internal heartbeat Rpc' to check if the server side is ready instead of relying on this?

mobly/controllers/android_device_lib/jsonrpc_client_base.py
xpconanfan: NBD.
It was a long explanation that I don't remember...

mobly/controllers/android_device_lib/jsonrpc_client_base.py
adorokhine: It's hard to change the RPC protocol because of sl4a. If we add a new message the framework relies on, we'd have to port it to sl4a and then backport it to all supported branches. :(

mobly/controllers/android_device_lib/snippet_client.py
xpconanfan: I know the point of nontrivial computation, which was the standard argument, in py2.
In py3, setting a property is not allowed. So it's still better to wrap it in property if it's something we don't want users to be able to accidentally change?

```
>>> class A:
...   def __init__(self):
...     self._var = 40
...   @property
...   def var(self):
...     return self._var
...
>>> a = A()
>>> a.var = 42
AttributeError: can't set attribute
```

adorokhine: There's no point to trying to do access control in python because you can 'accidentally' change literally anything. You can override another object's functions, imports, modules, even the setattr function itself. You can change self._package which underlies this property.

What makes self.package more important to protect than, say, self.stop_app?

I just can't get behind trying to access control one or two random attributes in python at the expense of more code, when there's hundreds of other ways anyone could accidentally or purposefully break our object, especially when it's a property the user controls anyway (since they pass it into the constructor).

It doesn't buy us anything but more verbose code; there's just no point to it.

mobly/controllers/android_device_lib/jsonrpc_client_base.py
xpconanfan: It seems odd that we can't make improvements to snippet lib because the two clients share a base class with sl4a client. If the situation demands, wouldn't it make sense to just stop sharing a base class if the clients are different enough?

We are not trying to keep the protocol identical with sl4a forever right? At one point we'll have some features that sl4a doesn't have and we shouldn't shoot down new features just to be compatible with sl4a? SL4A never had a good way to check if the connection to server is alive and well, we'd want to fix such a problem at one point, right?

But in this particular case, to preserve the inheritance, can we move `is_app_running` down to subclasses and have a different implementation? Also perhaps `is_app_running` should be named sth like `is_server_ready`?

mobly/controllers/android_device_lib/snippet_client.py
xpconanfan: I'm ambivalent to this, so either way is fine...

mobly/controllers/android_device_lib/jsonrpc_client_base.py
adorokhine: Sure we can fork it at some point, but you mentioned this specific bug was one you'd seen in sl4a too right?

Also sl4a isn't the only thing; snippetlib 1.0.0 is out now. This change would make us incompatible with 1.0.0 which people might be using. So we have to make protocol-breaking changes very carefully and with lots of notice to users. :)

In this case I think it's easy enough to work around by reusing one of the existing RPCs that it's not worth trying to change the protocol. Connect was just the easiest way but we could implement it as some kind of erroroneous RPC if you want.

xpconanfan: I was hoping we could fix this in snippet, and we can't fix sl4a anyway.
It is better to fix this now than waiting for later?

And at one point we would have to make breaking changes, so the client should be able to figure out a way to check which version of the lib is being used and behave accordingly?

Sounds like we need another 'internal rpc' for lib version checking....

adorokhine: The current implementation fixes sl4a too actually.
I agree having a version check would be nice going forward. Could you file an FR?

In this particular case, since we're still active users of sl4a and you've seen this issue, I'd like to fix it too :)

I don't think we need to start changing the protocol for this particular case since it's so easy to actually check... we just need ANY output from the socket. This could be a connection attempt, an invalid message that causes it to spit out an error, anything. So I'd like to push off the pain of forking/duplicating/versioning/cherrypicking until there's no easy workaround available. :)

xpconanfan: But we can address that in a separate PR

PR #96, opened 2017-01-31
Fixes in snippet client.

### LINKED ISSUE #89: [Snippet Client] Misleading error message in _is_app_installed()
If a app is installed but not instrumented properly, SnippetClient._is_app_installed() will throw an error saying the app is not installed.
The error message should correctly reflect the situation.

xpconanfan:
* Fix an inaccurate error message.
* Fix an incorrect call for retrieving serial number.

fixes #89


mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -74,12 +74,26 @@ def stop_app(self):
...
                 'pm list instrumentation | grep ^instrumentation:%s/' %
                 self.package).decode('utf-8')
-            return bool(out)
+            is_instrumented = bool(out)
         except adb.AdbError as e:
             if (e.ret_code == 1) and (not e.stdout) and (not e.stderr):
-                return False
+                is_instrumented = False
             raise
+        if not is_instrumented:
```
adorokhine: You're starting to conflate is_app_installed and check_app_installed. If is_app_installed isn't used without check_app_installed, could you delete it and merge them?

mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -74,12 +74,26 @@ def stop_app(self):
 
     def _is_app_installed(self):
         """Overrides superclass."""
+        # Check if app is installed
+        try:
```
adorokhine: We seem to have this pattern in a lot of places, so can you stick this in a shared place somewhere?

mobly/controllers/android_device_lib/snippet_client.py
xpconanfan: I was gonna merge them then I thought you might had these two separated intentionally to keep the external interface from the internal implementation.
What was the rationale to have these two separated?

adorokhine: No strong rationale, just reusing the checking and printing code.
Go ahead and make check_app_installed() the method that has to be implemented by subclasses.

mobly/controllers/android_device_lib/snippet_client.py
xpconanfan: I thought about that, but it seemed like coming up with a proper design and usage for this 'wrapper' would be more complicated than just writing it out.

Since we don't know what will go into the try block, we'd have a function that takes a function?
```
def adb_stuff():
  out = self.adb.blah()
  # do sth with out.

wrapper(adb_stuff)
```

or if you're thinking about a one line wrapper that returns the `out`, I was actually thinking about adding a param in Adb proxy itself for that.

Either way, it seems like a bigger change that deserves its own CL?

adorokhine: Are all the usages we have currently for checking a grep-like call? (meaning pass if 0 or output non-empty, fail if non-zero or output empty).
If so let's just make a wrapper for that specific case that accepts shell args.

mobly/controllers/android_device_lib/snippet_client.py
xpconanfan: Do we want to go with `is_` or `check_`.
Since there's also `_is_app_running`, it might be good to be consistent?

adorokhine: I'm asking you to remove _is_app_running.

adorokhine: Er wait, no, I'm asking you to remove is_app_installed.

Don't quite understand your comment then. You want to change is_app_running to check_app_running and in the retry loop catch and handle the exception instead of handling boolean output?

xpconanfan: Huh? not at all..
I'm asking if we should rename `check_app_installed` to `is_app_installed` while removing `_is_app_installed`.
Since `is_app_installed` is a better name and is consistent with the name `_is_app_running`

mobly/controllers/android_device_lib/snippet_client.py
xpconanfan: Done

mobly/controllers/android_device_lib/snippet_client.py
xpconanfan: Done

adorokhine: That would be mixing conventions:
is_* = returns boolean
check_* = asserts something and dies if false

That's why I'm suggesting to just make check_app_installed the public-facing function. That way you can throw a better error message.

xpconanfan: Done

mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -72,14 +72,16 @@ def stop_app(self):
...
-            out = self._adb.shell(
-                'pm list instrumentation | grep ^instrumentation:%s/' %
-                self.package).decode('utf-8')
-            return bool(out)
-        except adb.AdbError as e:
-            if (e.ret_code == 1) and (not e.stdout) and (not e.stderr):
-                return False
-            raise
+        if not self._adb_grep_wrapper(
+            'pm list package | grep %s' % self.package):
```
adorokhine: Can you make the grep more specific? ie grep 'package:%s$'
The snippet's package is usually a superset of the main app's package
eg
com.my.foo
com.my.foo.snippet

So there's opportunities for the grep to detect an app as 'installed' even if it really isn't.

mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -72,14 +72,16 @@ def stop_app(self):
...
+            raise jsonrpc_client_base.AppStartError(
+                '%s is not installed on %s' % (
+                self.app_name, self._adb.getprop('ro.boot.serialno')))
+        if not self._adb_grep_wrapper(
+            'pm list instrumentation | grep ^instrumentation:%s/' %
```
adorokhine: Can you add another grep call after this to make sure the instrumentation runner is actually SnippetRunner?

mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -72,14 +72,16 @@ def stop_app(self):
...
+            self.package):
```
adorokhine: Can you add another grep call after this to make sure the target of the instrumentation is actually installed?

mobly/controllers/android_device_lib/snippet_client.py
xpconanfan: Done

mobly/controllers/android_device_lib/snippet_client.py
xpconanfan: Don't we already check if it's installed before this call?

xpconanfan: Done

mobly/controllers/android_device_lib/snippet_client.py
xpconanfan: Done

mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -65,21 +67,36 @@ def _do_start_app(self):
...
+        cmd = _STOP_CMD % self.package
         logging.info('Stopping snippet apk with: %s', cmd)
-        out = self._adb.shell(_STOP_CMD.format(self.package)).decode('utf-8')
+        out = self._adb.shell(_STOP_CMD % self.package).decode('utf-8')
         if 'OK (0 tests)' not in out:
             raise Error('Failed to stop existing apk. Unexpected output: %s' %
                         out)
 
-    def _is_app_installed(self):
+    def check_app_installed(self):
```
adorokhine: This is starting to get complicated... should we add a test for this? (just for the output parsing logic)

xpconanfan: I was gonna file an issue to you for adding unit tests for all of these client stuff :P

adorokhine: jsonrpc_client_base has unit tests. Since you're expanding the installation checking logic in this PR, this is where the unit test should be added too. :)

PR #100, opened 2017-02-02
Fix a device detection bug in shell base.

xpconanfan:


mobly/controllers/android_device_lib/jsonrpc_shell_base.py
```diff
@@ -50,10 +50,15 @@ def load_device(self, serial=None):
         only one.
         """
         serials = android_device.list_adb_devices()
-        if (not serial) and len(serials) != 1:
-            raise Error('Expected 1 phone, but %d found. Use the -s flag.' %
+        if not serials:
+            raise Error("No adb device found!")
```
adorokhine: We decided on single quotes right?

PR #127, opened 2017-02-23
Adding mobly to rtd_requirements.txt

yusufm:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/127)

### DIFF
--- a/docs/rtd_requirements.txt
+++ b/docs/rtd_requirements.txt
@@ -3,3 +3,4 @@ CommonMark==0.5.4
 recommonmark==0.4.0
 Sphinx==1.5.2
 sphinx-autobuild==0.6.0
+mobly


adorokhine:
- Is there a way to test this prior to submission to make sure we've got everything?
- Won't this cause pip to install the latest release of mobly instead of what we're actually testing?

xpconanfan:
regarding 2: do we want to pull docs from HEAD?
perhaps it's better to version the docs with releases, like flask does.

adorokhine:
Our tutorials all render at HEAD right? So what the user would see when following the tutorial wouldn't match what the user saw in the api docs right?

xpconanfan:
Tutorials change much less frequently than API docs and requires much less update than API docs.
I think this is why readthedocs actually indexes md files as well, so they can be versioned, now it makes sense...
Perhaps it's better to point users to readthedocs by default so everything is versioned...

adorokhine:
You mean you can get to docs/tutorial.md on readthedocs.org, and it points to the latest release? If so, sgtm, can we add a link to that on the readme?

xpconanfan:
Yes, we can have readthedocs grab md files per release.
Will add after it's ready. Yusuf would need to make some changes to include them.

PR #140, opened 2017-03-09
[Snippet] Fix a bug in Rpc event timeout.

### LINKED ISSUE #138: Rpc socket times out before EventHandler.waitAndGet
`json_rpc_base._SOCKET_TIMEOUT` is 60s. Yet `EventHandler.waitAndGet` can take an arbitrary timeout length.

`EventHandler.waitAndGet` makes a call to `EventSnippet.eventWaitAndGet` through the socket, which doesn't return until the event is received or the method itself times out.

So if somebody calls `EventHandler.waitAndGet` with a timeout longer than `json_rpc_base._SOCKET_TIMEOUT`, we'd get a socket timeout error.

We should:
1. Make socket timeout longer
2. Add a guard in `EventHandler.waitAndGet` for a max timeout value.

xpconanfan:
* Increase the socket's max wait time for a response message.
* Add a guard for EventHandler's input timeout.

Fixes #138

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/140)

### DIFF
--- a/mobly/controllers/android_device_lib/callback_handler.py
+++ b/mobly/controllers/android_device_lib/callback_handler.py
@@ -21,7 +21,7 @@ from mobly.controllers.android_device_lib import snippet_event
 # The max timeout cannot be larger than the max time the socket waits for a
 # response message. Otherwise, the socket would timeout before the Rpc call
 # does, leaving both server and client in unknown states.
-MAX_TIMEOUT = 60 * 6
+MAX_TIMEOUT = 60 * 10
 
 
 class Error(Exception):
--- a/mobly/controllers/android_device_lib/callback_handler.py
+++ b/mobly/controllers/android_device_lib/callback_handler.py
@@ -16,7 +16,13 @@
 
 import time
 
+from mobly.controllers.android_device_lib import jsonrpc_client_base
 from mobly.controllers.android_device_lib import snippet_event
+
+# The max timeout cannot be larger than the max time the socket waits for a
+# response message. Otherwise, the socket would timeout before the Rpc call
+# does, leaving both server and client in unknown states.
+MAX_TIMEOUT = jsonrpc_client_base.SOCKET_READ_TIMEOUT
 
 
 class Error(Exception):
@@ -68,9 +74,15 @@ class CallbackHandler(object):
             SnippetEvent, the oldest entry of the specified event.
 
         Raises:
+            Error: If the specified timeout is longer than the max timeout
+                   supported.
             TimeoutError: The expected event does not occur within time limit.
         """
         if timeout:
+            if timeout > MAX_TIMEOUT:
+                raise Error(
+                    'Specified timeout %s is longer than max timeout %s.' %
+                    (timeout, MAX_TIMEOUT))
             timeout *= 1000  # convert to milliseconds for java side
         try:
             raw_event = self._event_client.eventWaitAndGet(self._id,
@@ -78,8 +90,8 @@ class CallbackHandler(object):
         except Exception as e:
             if 'EventSnippetException: timeout.' in str(e):
                 raise TimeoutError(
-                    'Timeout waiting for event "%s" triggered by %s (%s).'
-                    % (event_name, self._method_name, self._id))
+                    'Timeout waiting for event "%s" triggered by %s (%s).' %
+                    (event_name, self._method_name, self._id))
             raise
         return snippet_event.from_dict(raw_event)
 
--- a/mobly/controllers/android_device_lib/callback_handler.py
+++ b/mobly/controllers/android_device_lib/callback_handler.py
@@ -16,13 +16,12 @@
 
 import time
 
-from mobly.controllers.android_device_lib import jsonrpc_client_base
 from mobly.controllers.android_device_lib import snippet_event
 
 # The max timeout cannot be larger than the max time the socket waits for a
 # response message. Otherwise, the socket would timeout before the Rpc call
 # does, leaving both server and client in unknown states.
-MAX_TIMEOUT = jsonrpc_client_base.SOCKET_READ_TIMEOUT
+MAX_TIMEOUT = 60 * 6
 
 
 class Error(Exception):
--- a/mobly/controllers/android_device_lib/jsonrpc_client_base.py
+++ b/mobly/controllers/android_device_lib/jsonrpc_client_base.py
@@ -56,7 +56,7 @@ UNKNOWN_UID = -1
 _SOCKET_CONNECTION_TIMEOUT = 60
 
 # Maximum time to wait for a response message on the socket.
-SOCKET_READ_TIMEOUT = 60 * 6
+_SOCKET_READ_TIMEOUT = 60 * 6
 
 
 class Error(Exception):
@@ -73,9 +73,9 @@ class ApiError(Error):
 
 class ProtocolError(Error):
     """Raised when there is some error in exchanging data with server."""
-    NO_RESPONSE_FROM_HANDSHAKE = "No response from handshake."
-    NO_RESPONSE_FROM_SERVER = "No response from server."
-    MISMATCHED_API_ID = "Mismatched API id."
+    NO_RESPONSE_FROM_HANDSHAKE = 'No response from handshake.'
+    NO_RESPONSE_FROM_SERVER = 'No response from server.'
+    MISMATCHED_API_ID = 'Mismatched API id.'
 
 
 class JsonRpcCommand(object):
@@ -206,7 +206,7 @@ class JsonRpcClientBase(object):
         self._counter = self._id_counter()
         self._conn = socket.create_connection(('127.0.0.1', self.host_port),
                                               _SOCKET_CONNECTION_TIMEOUT)
-        self._conn.settimeout(SOCKET_READ_TIMEOUT)
+        self._conn.settimeout(_SOCKET_READ_TIMEOUT)
         self._client = self._conn.makefile(mode="brw")
 
         resp = self._cmd(cmd, uid)
--- a/mobly/controllers/android_device_lib/jsonrpc_client_base.py
+++ b/mobly/controllers/android_device_lib/jsonrpc_client_base.py
@@ -56,7 +56,7 @@ UNKNOWN_UID = -1
 _SOCKET_CONNECTION_TIMEOUT = 60
 
 # Maximum time to wait for a response message on the socket.
-_SOCKET_READ_TIMEOUT = 60 * 6
+_SOCKET_READ_TIMEOUT = 60 * 10
 
 
 class Error(Exception):
--- a/mobly/controllers/android_device_lib/jsonrpc_client_base.py
+++ b/mobly/controllers/android_device_lib/jsonrpc_client_base.py
@@ -56,7 +56,7 @@ UNKNOWN_UID = -1
 _SOCKET_CONNECTION_TIMEOUT = 60
 
 # Maximum time to wait for a response message on the socket.
-_SOCKET_READ_TIMEOUT = 60 * 10
+_SOCKET_READ_TIMEOUT = callback_handler.MAX_TIMEOUT
 
 
 class Error(Exception):
--- a/mobly/controllers/android_device_lib/jsonrpc_client_base.py
+++ b/mobly/controllers/android_device_lib/jsonrpc_client_base.py
@@ -53,7 +53,10 @@ APP_START_WAIT_TIME = 15
 UNKNOWN_UID = -1
 
 # Maximum time to wait for the socket to open on the device.
-_SOCKET_TIMEOUT = 60
+_SOCKET_CONNECTION_TIMEOUT = 60
+
+# Maximum time to wait for a response message on the socket.
+SOCKET_READ_TIMEOUT = 60 * 6
 
 
 class Error(Exception):
@@ -186,9 +189,9 @@ class JsonRpcClientBase(object):
         """Opens a connection to a JSON RPC server.
 
         Opens a connection to a remote client. The connection attempt will time
-        out if it takes longer than _SOCKET_TIMEOUT seconds. Each subsequent
-        operation over this socket will time out after _SOCKET_TIMEOUT seconds
-        as well.
+        out if it takes longer than _SOCKET_CONNECTION_TIMEOUT seconds. Each
+        subsequent operation over this socket will time out after
+        _SOCKET_READ_TIMEOUT seconds as well.
 
         Args:
             uid: int, The uid of the session to join, or UNKNOWN_UID to start a
@@ -202,9 +205,9 @@ class JsonRpcClientBase(object):
         """
         self._counter = self._id_counter()
         self._conn = socket.create_connection(('127.0.0.1', self.host_port),
-                                              _SOCKET_TIMEOUT)
-        self._conn.settimeout(_SOCKET_TIMEOUT)
-        self._client = self._conn.makefile(mode='brw')
+                                              _SOCKET_CONNECTION_TIMEOUT)
+        self._conn.settimeout(SOCKET_READ_TIMEOUT)
+        self._client = self._conn.makefile(mode="brw")
 
         resp = self._cmd(cmd, uid)
         if not resp:
--- a/mobly/controllers/android_device_lib/jsonrpc_client_base.py
+++ b/mobly/controllers/android_device_lib/jsonrpc_client_base.py
@@ -207,7 +207,7 @@ class JsonRpcClientBase(object):
         self._conn = socket.create_connection(('127.0.0.1', self.host_port),
                                               _SOCKET_CONNECTION_TIMEOUT)
         self._conn.settimeout(_SOCKET_READ_TIMEOUT)
-        self._client = self._conn.makefile(mode="brw")
+        self._client = self._conn.makefile(mode='brw')
 
         resp = self._cmd(cmd, uid)
         if not resp:
--- a/tests/mobly/controllers/android_device_lib/callback_handler_test.py
+++ b/tests/mobly/controllers/android_device_lib/callback_handler_test.py
@@ -37,6 +37,10 @@ class CallbackHandlerTest(unittest.TestCase):
     """Unit tests for mobly.controllers.android_device_lib.callback_handler.
     """
 
+    def test_timeout_value(self):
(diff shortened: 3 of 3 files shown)


adorokhine:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/callback_handler.py, line 25 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9Pslq_nRF_cech8C9:-Kf9Pslq_nRF_cech8CA:bv62v3v) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/callback_handler.py#L25)):*
> ```Python
> # does, leaving both server and client in unknown states.
> MAX_TIMEOUT = 60 * 6
> ```

According to the docs, if you don't call settimeout(), the socket won't time out. If we are making the assumption that the connection will break if the server dies regardless of the timeout, seems like we should delete settimeout() completely rather than setting it to huge values and then increasing it every time we have a longer async call?

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 210 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9QHqXr_hgRjHfk6TN:-Kf9QHqXr_hgRjHfk6TO:berfly9) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L210)):*
> ```Python
>                                               _SOCKET_CONNECTION_TIMEOUT)
>         self._conn.settimeout(_SOCKET_READ_TIMEOUT)
>         self._client = self._conn.makefile(mode="brw")
> ```

Changed from single quotes to double quotes? Isn't that backwards?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

xpconanfan:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/callback_handler.py, line 25 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9Pslq_nRF_cech8C9:-Kf9QzSGk1hPwddnbuyN:b-l73rtc) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/callback_handler.py#L25)):*

According to the docs, if you don't call settimeout(), the socket won't time out. If we are making the assumption that the connection will break if the server dies regardless of the timeout, seems like we should delete settimeout() completely rather than setting it to huge values and then increasing it every time we have a longer async call?
</blockquote></details>

But that would also make it possible for an Rpc call to block forever and never returns? Are we ok with that?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

adorokhine:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/callback_handler.py, line 25 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9Pslq_nRF_cech8C9:-Kf9RRqbRpOaFt-rPLoB:bfwammr) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/callback_handler.py#L25)):*

But that would also make it possible for an Rpc call to block forever and never returns? Are we ok with that?
</blockquote></details>

Can you describe the use case you're picturing where there is a practical and important difference between waiting forever and waiting for 6 minutes?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

xpconanfan:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/callback_handler.py, line 25 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9Pslq_nRF_cech8C9:-Kf9SmBLQqJpNUQlc8tn:b-3l121y) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/callback_handler.py#L25)):*

Can you describe the use case you're picturing where there is a practical and important difference between waiting forever and waiting for 6 minutes?
</blockquote></details>

Waiting for 6 min means the test eventually finishes, whereas waiting forever doesn't?
Seems a bit odd to wait forever, especially if the test is not run by a separate harness, we'd just let it hang forever?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

adorokhine:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/callback_handler.py, line 25 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9Pslq_nRF_cech8C9:-Kf9TYLqrVvTydUqZpbH:b1aqxx4) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/callback_handler.py#L25)):*

Waiting for 6 min means the test eventually finishes, whereas waiting forever doesn't?
Seems a bit odd to wait forever, especially if the test is not run by a separate harness, we'd just let it hang forever?
</blockquote></details>

6 minutes seems like a pretty arbitrary value.... is it driven by the current longest use case that we know of? Is the plan to just keep increasing this value whenever a longer use case comes in? Maybe we can get ahead of this train and set it to a nice large round number like 15min?

I think the situation you're describing is a continuous build (because manual users wouldn't wait that long) which doesn't have a timeout on the subcommands that are being run? Do you know of one?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

xpconanfan:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/callback_handler.py, line 25 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9Pslq_nRF_cech8C9:-Kf9Vb55Id3C2vPsZP3d:b-5n6gm3) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/callback_handler.py#L25)):*

6 minutes seems like a pretty arbitrary value.... is it driven by the current longest use case that we know of? Is the plan to just keep increasing this value whenever a longer use case comes in? Maybe we can get ahead of this train and set it to a nice large round number like 15min?

I think the situation you're describing is a continuous build (because manual users wouldn't wait that long) which doesn't have a timeout on the subcommands that are being run? Do you know of one?
</blockquote></details>

I don't see how 15 is less arbitrary than 6, but sure.

Whether the test is run manually or not does not make any difference. I don't think we should assume that all tests are run in some kind of integration system, especially when it is totally valid to run a test manually during debug phase.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

adorokhine:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/callback_handler.py, line 25 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9Pslq_nRF_cech8C9:-Kf9XK_t5njm2VeaaHKj:b-wyz1vl) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/callback_handler.py#L25)):*

I don't see how 15 is less arbitrary than 6, but sure.

Whether the test is run manually or not does not make any difference. I don't think we should assume that all tests are run in some kind of integration system, especially when it is totally valid to run a test manually during debug phase.
</blockquote></details>

15 doesn't seem like a reaction to cover some particular existing use case but a nice round default intended to cover general use cases going forward :) same reason that a speed limit of 75 looks ok but a speed limit of 67.5 makes you go hmm...

Manual users wouldn't wait around for 6 minutes while the phone was doing nothing, surely? They'd just kill it if something went wrong and fix the test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

xpconanfan:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/callback_handler.py, line 25 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9Pslq_nRF_cech8C9:-Kf9XzJhPj5sWKEtduXV:bcfa0fc) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/callback_handler.py#L25)):*

15 doesn't seem like a reaction to cover some particular existing use case but a nice round default intended to cover general use cases going forward :) same reason that a speed limit of 75 looks ok but a speed limit of 67.5 makes you go hmm...

Manual users wouldn't wait around for 6 minutes while the phone was doing nothing, surely? They'd just kill it if something went wrong and fix the test.
</blockquote></details>

If they had to manually interrupt, teardown doesn't happen, which could leave the devices in unknown states.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

xpconanfan:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/callback_handler.py, line 25 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9Pslq_nRF_cech8C9:-Kf9YGlHkTMf2THJUlrT:b-hfzy3f) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/callback_handler.py#L25)):*

If they had to manually interrupt, teardown doesn't happen, which could leave the devices in unknown states.
</blockquote></details>

The same goes for automated run.
It is better for the test to timeout itself than being forcefully killed, in which case we lose outputs and debug info.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

adorokhine:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/callback_handler.py, line 25 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9Pslq_nRF_cech8C9:-Kf9_5btP6YZUI_luBEM:b-x62h04) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/callback_handler.py#L25)):*

The same goes for automated run.
It is better for the test to timeout itself than being forcefully killed, in which case we lose outputs and debug info.
</blockquote></details>

Fair point; eventWaitAndGet() has a timeout built in right? Any reason we shouldn't be relying on that user-configurable, call-specific timeout to make sure we don't wait forever?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

xpconanfan:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/callback_handler.py, line 25 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9Pslq_nRF_cech8C9:-Kf9bZ4cyrOW1tQZWMzC:b7go405) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/callback_handler.py#L25)):*

Fair point; eventWaitAndGet() has a timeout built in right? Any reason we shouldn't be relying on that user-configurable, call-specific timeout to make sure we don't wait forever?
</blockquote></details>

Huh? for that one we can, but not every Rpc has a timeout built-in. So this one on the socket is needed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

adorokhine:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/callback_handler.py, line 25 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9Pslq_nRF_cech8C9:-Kf9dMtRi-ZgkWGm6xYI:b-4lduov) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/callback_handler.py#L25)):*

Huh? for that one we can, but not every Rpc has a timeout built-in. So this one on the socket is needed.
</blockquote></details>

You mean for blocking RPCs? sg, let's change it to something round like 5, 10 or 15 so it doesn't look so random and call it a day.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

xpconanfan:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/callback_handler.py, line 25 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9Pslq_nRF_cech8C9:-Kf9hcMMZ6SDXIX_WgaH:b-896fix) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/callback_handler.py#L25)):*

You mean for blocking RPCs? sg, let's change it to something round like 5, 10 or 15 so it doesn't look so random and call it a day.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

xpconanfan:
Review status: 1 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 210 at r1](https://reviewable.io:443/reviews/google/mobly/140#-Kf9QHqXr_hgRjHfk6TN:-Kf9lkyHWr5hm3rWvBr2:b-896fix) ([raw file](https://github.com/google/mobly/blob/2367a171beaeb180b62c380a75fa2761393de11b/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L210)):*

Changed from single quotes to double quotes? Isn't that backwards?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

adorokhine:
Reviewed 2 of 2 files at r2.

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 58 at r2](https://reviewable.io:443/reviews/google/mobly/140#-Kf9mp6YTTbjVq2iS_oQ:-Kf9mp6YTTbjVq2iS_oR:b-yqr34u) ([raw file](https://github.com/google/mobly/blob/d5a068269c406e7c3df289129911598ff2ca55c7/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L58)):*
> ```Python
> _SOCKET_CONNECTION_TIMEOUT = 60
> 
> # Maximum time to wait for a response message on the socket.
> ```

Should we assign this from callback_handler.MAX_TIMEOUT so we don't have to update two places? Or alternatively should we add comments saying to update the other place if one is changed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

xpconanfan:

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 58 at r2](https://reviewable.io:443/reviews/google/mobly/140#-Kf9mp6YTTbjVq2iS_oQ:-Kf9noz9EWHPa7rMpV9e:b-pi3z0p) ([raw file](https://github.com/google/mobly/blob/d5a068269c406e7c3df289129911598ff2ca55c7/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L58)):*

Should we assign this from callback_handler.MAX_TIMEOUT so we don't have to update two places? Or alternatively should we add comments saying to update the other place if one is changed?
</blockquote></details>

Unfortunately we can't due to circular import.
I've added a unit test guarding this :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

xpconanfan:

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 58 at r2](https://reviewable.io:443/reviews/google/mobly/140#-Kf9mp6YTTbjVq2iS_oQ:-Kf9nxcQ5J05kzNs8h8T:b-b0x0id) ([raw file](https://github.com/google/mobly/blob/d5a068269c406e7c3df289129911598ff2ca55c7/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L58)):*

Unfortunately we can't due to circular import.
I've added a unit test guarding this :)
</blockquote></details>

oh, wait, i could do the other way around, hmmm.
It's a bit odd having the socket timeout dictated by callback handler though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

adorokhine:

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 58 at r2](https://reviewable.io:443/reviews/google/mobly/140#-Kf9mp6YTTbjVq2iS_oQ:-Kf9o7ukjdMW-JVpJVfU:b-a13t02) ([raw file](https://github.com/google/mobly/blob/d5a068269c406e7c3df289129911598ff2ca55c7/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L58)):*

oh, wait, i could do the other way around, hmmm.
It's a bit odd having the socket timeout dictated by callback handler though.
</blockquote></details>

It's a little bit odd but not too bad. At least we don't have to update two places. We could also move the timeout to another place like constants.py and import that from both?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

xpconanfan:

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 58 at r2](https://reviewable.io:443/reviews/google/mobly/140#-Kf9mp6YTTbjVq2iS_oQ:-Kf9ooiwmLutOm8s_XVO:b-896fix) ([raw file](https://github.com/google/mobly/blob/d5a068269c406e7c3df289129911598ff2ca55c7/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L58)):*

It's a little bit odd but not too bad. At least we don't have to update two places. We could also move the timeout to another place like constants.py and import that from both?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140)*

adorokhine:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/140#-:-Kf9pFmV2hOaQE957yzS:bnfp4nl)*

PR #146, opened 2017-03-14
Fix a bug where string is parsed as dict for AndroidDevice config

### LINKED ISSUE #144: controllers/android_device.py throws an exception on android configuration with just string
Seems like an earlier checkin to switch to using base64 string to load configuration might have introduced a bug in mobly/controllers/android_device.create(configs).  I was hitting an issue where I specified the android device serial in string format in my configuration file but was failing the isinstance check with str, so instead it tries to interpret my android config as a dict format and fails.  Changing the isinstance check with basestring instead of str seems to have solved the problem.

Here is the snippet of the exception thrown.
mobly/controllers/android_device.py", line 81, in create
mobly/controllers/android_device.py", line 219, in get_instances_with_configs
MoblyTest [ln]     serial = c.pop('serial')
MoblyTest [ln] AttributeError: 'str' object has no attribute 'pop'

xpconanfan:
Because checking if a token is a str has inconsistencies between py2 and py3, we should flip the logic to check for dict instead.

Fixes #144 

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/146)

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -86,12 +86,12 @@ def create(configs):
         # Configs is a list of strings representing serials.
         ads = get_instances(configs)
     else:
-        raise Error("No valid config found in: %s" configs)
+        raise Error("No valid config found in: %s" % configs)
     connected_ads = list_adb_devices()
 
     for ad in ads:
         if ad.serial not in connected_ads:
-            raise DeviceError(ad, 'Android device is specified in config but '
+            raise DeviceError(ad, 'Android device is specified in config but'
                               ' is not attached.')
     _start_services_on_ads(ads)
     return ads
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -78,12 +78,12 @@ def create(configs):
         ads = get_all_instances()
     elif not isinstance(configs, list):
         raise Error(ANDROID_DEVICE_NOT_LIST_CONFIG_MSG)
-    elif isinstance(configs[0], str):
-        # Configs is a list of serials.
-        ads = get_instances(configs)
-    else:
+    elif isinstance(configs[0], dict):
         # Configs is a list of dicts.
         ads = get_instances_with_configs(configs)
+    else:
+        # Configs is a list of strings representing serials.
+        ads = get_instances(configs)
     connected_ads = list_adb_devices()
 
     for ad in ads:
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -16,6 +16,7 @@
 
 from builtins import str
 from builtins import open
+from past.builtins import basestring
 
 import contextlib
 import logging
@@ -81,9 +82,11 @@ def create(configs):
     elif isinstance(configs[0], dict):
         # Configs is a list of dicts.
         ads = get_instances_with_configs(configs)
-    else:
+    elif isinstance(configs[0], basestring):
         # Configs is a list of strings representing serials.
         ads = get_instances(configs)
+    else:
+        raise Error("No valid config found in: %s" configs)
     connected_ads = list_adb_devices()
 
     for ad in ads:
--- a/tests/lib/mock_android_device.py
+++ b/tests/lib/mock_android_device.py
@@ -33,13 +33,25 @@ def get_mock_ads(num):
     """
     ads = []
     for i in range(num):
-        ad = mock.MagicMock(name="AndroidDevice", serial=i, h_port=None)
+        ad = mock.MagicMock(name="AndroidDevice", serial=str(i), h_port=None)
         ads.append(ad)
     return ads
 
 
 def get_all_instances():
     return get_mock_ads(5)
+
+
+def get_instances(serials):
+    ads = []
+    for serial in serials:
+        ad = mock.MagicMock(name="AndroidDevice", serial=serial, h_port=None)
+        ads.append(ad)
+    return ads
+
+
+def get_instances_with_configs(dicts):
+    return get_instances([d['serial'] for d in dicts])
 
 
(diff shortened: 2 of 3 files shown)


adorokhine:

---

*[mobly/controllers/android_device.py, line 85 at r1](https://reviewable.io:443/reviews/google/mobly/146#-KfE6o_hpll25LNdo5My:-KfE6o_iNDJWTEohNYRi:b7wxyyw) ([raw file](https://github.com/google/mobly/blob/646cad3e4bf0166c43c9870fa35976a03c0c63d8/mobly/controllers/android_device.py#L85)):*
> ```Python
>         ads = get_instances_with_configs(configs)
>     else:
>         # Configs is a list of strings representing serials.
> ```

Instead of 'else', should we add a check here that it really is of string type and die completely if it's unrecognized?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/146)*

xpconanfan:

---

*[mobly/controllers/android_device.py, line 85 at r1](https://reviewable.io:443/reviews/google/mobly/146#-KfE6o_hpll25LNdo5My:-KfE8Tfl_mUwXX-C8W4-:b-rh7m8s) ([raw file](https://github.com/google/mobly/blob/646cad3e4bf0166c43c9870fa35976a03c0c63d8/mobly/controllers/android_device.py#L85)):*

Instead of 'else', should we add a check here that it really is of string type and die completely if it's unrecognized?
</blockquote></details>

The whole point of this change is to avoid checking if something is of string type, since it is not reliable in py2.
Specifically, the string type check `isinstance(configs[0], str)` returns false even when something is a string.
So adding that check here would cause false negative failures.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/146)*

adorokhine:

---

*[mobly/controllers/android_device.py, line 85 at r1](https://reviewable.io:443/reviews/google/mobly/146#-KfE6o_hpll25LNdo5My:-KfE9cjXCGbTQZKDmZh8:b-khzmmy) ([raw file](https://github.com/google/mobly/blob/646cad3e4bf0166c43c9870fa35976a03c0c63d8/mobly/controllers/android_device.py#L85)):*

The whole point of this change is to avoid checking if something is of string type, since it is not reliable in py2.
Specifically, the string type check `isinstance(configs[0], str)` returns false even when something is a string.
So adding that check here would cause false negative failures.
</blockquote></details>

Per the [py2/3 compatibility cheat sheet](http://python-future.org/compatible_idioms.html#basestring), checking for string should look like this:
```
from past.builtins import basestring
string1 = u'hi'
string2 = b'hi'
print(isinstance(string1, basestring))
print(isinstance(string2, basestring))
```

```
True
True
```

under both py2 and py3.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/146)*

xpconanfan:

---

*[mobly/controllers/android_device.py, line 85 at r1](https://reviewable.io:443/reviews/google/mobly/146#-KfE6o_hpll25LNdo5My:-KfEAXarNf5sMyncRcm9:b-896fix) ([raw file](https://github.com/google/mobly/blob/646cad3e4bf0166c43c9870fa35976a03c0c63d8/mobly/controllers/android_device.py#L85)):*

Per the [py2/3 compatibility cheat sheet](http://python-future.org/compatible_idioms.html#basestring), checking for string should look like this:
```
from past.builtins import basestring
string1 = u'hi'
string2 = b'hi'
print(isinstance(string1, basestring))
print(isinstance(string2, basestring))
```

```
True
True
```

under both py2 and py3.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/146)*

adorokhine:
Can we add a unit test for this codepath?

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/146#-:-KfEDl_EAL1VBxm77msN:b-serfsh)*

adorokhine:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[mobly/controllers/android_device.py, line 89 at r2](https://reviewable.io:443/reviews/google/mobly/146#-KfEDpoMSq13d_RU3kPv:-KfEDpoMSq13d_RU3kPw:bm1k50w) ([raw file](https://github.com/google/mobly/blob/4e7f1b682dd59f599183574b204dd09546689cd1/mobly/controllers/android_device.py#L89)):*
> ```Python
>         ads = get_instances(configs)
>     else:
>         raise Error("No valid config found in: %s" configs)
> ```

No formatting means this probably won't work/parse. Is this tested?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/146)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 89 at r2](https://reviewable.io:443/reviews/google/mobly/146#-KfEDpoMSq13d_RU3kPv:-KfEIP-WwDSSUhrm_VY5:b-896fix) ([raw file](https://github.com/google/mobly/blob/4e7f1b682dd59f599183574b204dd09546689cd1/mobly/controllers/android_device.py#L89)):*

No formatting means this probably won't work/parse. Is this tested?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/146)*

adorokhine:
:lgtm:

---

Reviewed 3 of 3 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/146#-:-KfEOZQOOClWLItFpxob:bnfp4nl)*

PR #161, opened 2017-03-23
Support older versions of psutil.

adorokhine:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/161)

### DIFF
--- a/mobly/utils.py
+++ b/mobly/utils.py
@@ -376,13 +376,13 @@ def stop_standing_subprocess(proc, kill_signal=signal.SIGTERM):
     for child in children:
         try:
             child.kill()
-            child.wait()
+            child.wait(timeout=10)
         except:
             success = False
             logging.exception('Failed to kill standing subprocess %d', child.pid)
     try:
         process.kill()
-        process.wait()
+        process.wait(timeout=10)
     except:
         success = False
         logging.exception('Failed to kill standing subprocess %d', pid)
--- a/mobly/utils.py
+++ b/mobly/utils.py
@@ -371,7 +371,7 @@ def stop_standing_subprocess(proc, kill_signal=signal.SIGTERM):
     try:
       children = process.children(recursive=True)
     except AttributeError:
-      # Handle older versions of psutil
+      # Handle versions <3.0.0 of psutil.
       children = process.get_children(recursive=True)
     for child in children:
         try:
--- a/mobly/utils.py
+++ b/mobly/utils.py
@@ -376,11 +376,13 @@ def stop_standing_subprocess(proc, kill_signal=signal.SIGTERM):
     for child in children:
         try:
             child.kill()
+            child.wait()
         except:
             success = False
             logging.exception('Failed to kill standing subprocess %d', child.pid)
     try:
         process.kill()
+        process.wait()
     except:
         success = False
         logging.exception('Failed to kill standing subprocess %d', pid)
--- a/mobly/utils.py
+++ b/mobly/utils.py
@@ -368,7 +368,12 @@ def stop_standing_subprocess(proc, kill_signal=signal.SIGTERM):
     _assert_subprocess_running(proc)
     process = psutil.Process(pid)
     success = True
-    for child in process.children(recursive=True):
+    try:
+      children = process.children(recursive=True)
+    except AttributeError:
+      # Handle older versions of psutil
+      children = process.get_children(recursive=True)
+    for child in children:
         try:
             child.kill()
         except:
--- a/mobly/utils.py
+++ b/mobly/utils.py
@@ -369,10 +369,10 @@ def stop_standing_subprocess(proc, kill_signal=signal.SIGTERM):
     process = psutil.Process(pid)
     success = True
     try:
-      children = process.children(recursive=True)
+        children = process.children(recursive=True)
     except AttributeError:
-      # Handle versions <3.0.0 of psutil.
-      children = process.get_children(recursive=True)
+        # Handle versions <3.0.0 of psutil.
+        children = process.get_children(recursive=True)
     for child in children:
         try:
             child.kill()
--- a/tests/mobly/utils_test.py
+++ b/tests/mobly/utils_test.py
@@ -33,7 +33,6 @@ class UtilsTest(unittest.TestCase):
 
     def test_stop_standing_subproc(self):
         p = utils.start_standing_subprocess('sleep 5')
-        time.sleep(0.1)
         utils.stop_standing_subprocess(p)
         self.assertIsNotNone(p.poll())
 
--- a/tests/mobly/utils_test.py
+++ b/tests/mobly/utils_test.py
@@ -32,6 +32,12 @@ class UtilsTest(unittest.TestCase):
             utils.start_standing_subprocess("sleep 0", check_health_delay=0.1)
 
     def test_stop_standing_subproc(self):
+        p = utils.start_standing_subprocess('sleep 5')
+        time.sleep(0.1)
+        utils.stop_standing_subprocess(p)
+        self.assertIsNotNone(p.poll())
+
+    def test_stop_standing_subproc_already_dead(self):
         p = utils.start_standing_subprocess("sleep 0")
         time.sleep(0.1)
         with self.assertRaisesRegexp(utils.Error,


xpconanfan:

---

*[mobly/utils.py, line 372 at r1](https://reviewable.io:443/reviews/google/mobly/161#-Kft3ZzZP8Y4042U6lnn:-Kft3ZzZP8Y4042U6lno:bq06rs1) ([raw file](https://github.com/google/mobly/blob/a6dfa381114ae8b117df25e3e5e366eec5fb7ac9/mobly/utils.py#L372)):*
> ```Python
>     success = True
>     try:
>       children = process.children(recursive=True)
> ```

can we make the version requirement in `setup.py` like we do with `mock` instead of doing this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/161)*

adorokhine:

---

*[mobly/utils.py, line 372 at r1](https://reviewable.io:443/reviews/google/mobly/161#-Kft3ZzZP8Y4042U6lnn:-Kft4T0V5zOLi4pI_d3w:b-5yo8uq) ([raw file](https://github.com/google/mobly/blob/a6dfa381114ae8b117df25e3e5e366eec5fb7ac9/mobly/utils.py#L372)):*

can we make the version requirement in `setup.py` like we do with `mock` instead of doing this?
</blockquote></details>

The problem here is that all python modules are global, and it's likely psutil would have already been installed by something else. For example, it's a mandatory dependency on ubuntu (cannot be removed) and ships by default with many windows python distributions including activestate (cannot be removed). Ubuntu 10.04 for example ships with psutil 1.2.1.

If we add a version requirement on such a system, setup.py will fail:
`pkg_resources.VersionConflict: (psutil 1.2.1 (/usr/lib/python2.7/dist-packages), Requirement.parse('psutil>=5.2.0'))`

But there's no way for the user to fix this because the version that comes on their system cannot be replaced or upgraded in many cases.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/161)*

xpconanfan:

---

*[mobly/utils.py, line 372 at r1](https://reviewable.io:443/reviews/google/mobly/161#-Kft3ZzZP8Y4042U6lnn:-Kft8ofd9JnuKCHFpvPa:bdhcsnz) ([raw file](https://github.com/google/mobly/blob/a6dfa381114ae8b117df25e3e5e366eec5fb7ac9/mobly/utils.py#L372)):*

The problem here is that all python modules are global, and it's likely psutil would have already been installed by something else. For example, it's a mandatory dependency on ubuntu (cannot be removed) and ships by default with many windows python distributions including activestate (cannot be removed). Ubuntu 10.04 for example ships with psutil 1.2.1.

If we add a version requirement on such a system, setup.py will fail:
`pkg_resources.VersionConflict: (psutil 1.2.1 (/usr/lib/python2.7/dist-packages), Requirement.parse('psutil>=5.2.0'))`

But there's no way for the user to fix this because the version that comes on their system cannot be replaced or upgraded in many cases.
</blockquote></details>

Maybe we can require a lower version to lower the impact?
I think a user can still remove existing version of psutil then install the new version.

But if not, let's specify the exact cut off version in the comment line below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/161)*

adorokhine:

---

*[mobly/utils.py, line 372 at r1](https://reviewable.io:443/reviews/google/mobly/161#-Kft3ZzZP8Y4042U6lnn:-KftA3U7mZUSB_M9WsOe:bi7afh8) ([raw file](https://github.com/google/mobly/blob/a6dfa381114ae8b117df25e3e5e366eec5fb7ac9/mobly/utils.py#L372)):*
> Maybe we can require a lower version to lower the impact?
I think the impact will be the same since we can't allow versions as old as `1.2.1`, which is what ships with goobuntu.

> I think a user can still remove existing version of psutil
How? Trying to do `apt-get remove python-psutil` will fail because it would try to remove `ubuntu-essential` which cannot be removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/161)*

xpconanfan:

---

*[mobly/utils.py, line 372 at r1](https://reviewable.io:443/reviews/google/mobly/161#-Kft3ZzZP8Y4042U6lnn:-KftB2loBVe-Mz-gTOTP:b-q1j9fa) ([raw file](https://github.com/google/mobly/blob/a6dfa381114ae8b117df25e3e5e366eec5fb7ac9/mobly/utils.py#L372)):*

> Maybe we can require a lower version to lower the impact?
I think the impact will be the same since we can't allow versions as old as `1.2.1`, which is what ships with goobuntu.

> I think a user can still remove existing version of psutil
How? Trying to do `apt-get remove python-psutil` will fail because it would try to remove `ubuntu-essential` which cannot be removed.
</blockquote></details>

Removing the pacakage files directly from dist-packages is usually how I remove python libs.
`apt-get remove` does not work well with python packages...

But if you think it's too big a pain, let's just specify the cut off version in comment. Either way is fine.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/161)*

adorokhine:

---

*[mobly/utils.py, line 372 at r1](https://reviewable.io:443/reviews/google/mobly/161#-Kft3ZzZP8Y4042U6lnn:-KftBdaq5Y1S-PSd7PNk:boe3w40) ([raw file](https://github.com/google/mobly/blob/a6dfa381114ae8b117df25e3e5e366eec5fb7ac9/mobly/utils.py#L372)):*
> Removing the pacakage files directly from dist-packages is usually how I remove python libs.

1. This requires root access. Some deployments don't have root access, eg many corporate Windows deployments.
2. Wouldn't this break whatever in ubuntu-essential was actually using psutil? Because now the core system modules are trying to call `get_children()` but not finding it.

So I don't think that's a good solution.

> apt-get remove does not work well with python packages...

Sure it does, why not? It just doesn't work with this one because it's a mandatory core system dependency.

> But if you think it's too big a pain, let's just specify the cut off version in comment.

With this implementation there is no cutoff version; we'll work on any version of psutil that exists currently.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/161)*

dthkao:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/utils.py, line 374 at r1](https://reviewable.io:443/reviews/google/mobly/161#-KftIgVYdlUU-BVHPb_9:-KftIgVYdlUU-BVHPb_A:b98dm3y) ([raw file](https://github.com/google/mobly/blob/a6dfa381114ae8b117df25e3e5e366eec5fb7ac9/mobly/utils.py#L374)):*
> older versions
Do you happen to know the cutoff? Can we document it here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/161)*

adorokhine:

---

*[mobly/utils.py, line 374 at r1](https://reviewable.io:443/reviews/google/mobly/161#-KftIgVYdlUU-BVHPb_9:-KftKFYqErc-y7DGdX8Y:b-896fix) ([raw file](https://github.com/google/mobly/blob/a6dfa381114ae8b117df25e3e5e366eec5fb7ac9/mobly/utils.py#L374)):*

> older versions
Do you happen to know the cutoff? Can we document it here?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/161)*

xpconanfan:

---

*[mobly/utils.py, line 379 at r2](https://reviewable.io:443/reviews/google/mobly/161#-Kg0t4FOWM4gedVMMi2H:-Kg0t4FOWM4gedVMMi2I:bp22bm8) ([raw file](https://github.com/google/mobly/blob/ea79f9ca7706f4ebe3eb1a92a38f669845c35d99/mobly/utils.py#L379)):*
> ```Python
>         try:
>             child.kill()
>             child.wait()
> ```

If we want to wait, we should have a timeout and an extra block handling `psutil.TimeoutExpired`?
Same for the one below.

---

*[tests/mobly/utils_test.py, line 36 at r2](https://reviewable.io:443/reviews/google/mobly/161#-Kg0slv-inZI6jegtd8M:-Kg0slv-inZI6jegtd8N:b-h8aqh2) ([raw file](https://github.com/google/mobly/blob/ea79f9ca7706f4ebe3eb1a92a38f669845c35d99/tests/mobly/utils_test.py#L36)):*
> ```Python
>     def test_stop_standing_subproc(self):
>         p = utils.start_standing_subprocess('sleep 5')
>         time.sleep(0.1)
> ```

We don't need this sleep do we?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/161)*

adorokhine:
Reviewed 1 of 1 files at r1, 2 of 2 files at r2.
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[mobly/utils.py, line 379 at r2](https://reviewable.io:443/reviews/google/mobly/161#-Kg0t4FOWM4gedVMMi2H:-Kg1HPTt3SyFMODQSHcp:b-2y5j3u) ([raw file](https://github.com/google/mobly/blob/ea79f9ca7706f4ebe3eb1a92a38f669845c35d99/mobly/utils.py#L379)):*

If we want to wait, we should have a timeout and an extra block handling `psutil.TimeoutExpired`?
Same for the one below.
</blockquote></details>

Added timeout. I don't think the handling for psutil.TimeoutExpired should be any different than the current except block right?

---

*[tests/mobly/utils_test.py, line 36 at r2](https://reviewable.io:443/reviews/google/mobly/161#-Kg0slv-inZI6jegtd8M:-Kg1HlAH1NmmprT7xDjf:b-896fix) ([raw file](https://github.com/google/mobly/blob/ea79f9ca7706f4ebe3eb1a92a38f669845c35d99/tests/mobly/utils_test.py#L36)):*

We don't need this sleep do we?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/161)*

xpconanfan:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/161#-:-KgFr8eEaO_8alpuQ_cb:bnfp4nl)*

adorokhine:
Reviewed 2 of 2 files at r3, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/161)*

PR #180, opened 2017-04-07
Fix snippet_client so that it can be used with default device.

artikz:
The problem was introduced in [1], where snippet_client always assumes
that device has a serial number, which may not always be the case.

[1] https://github.com/google/mobly/commit/e5a553c6cb238c354b6402e6de1876c3605beb77

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/180)


dthkao:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/180#-:-Kh7t-ZW0d_RzZ-gkbgA:bnfp4nl)*

PR #194, opened 2017-04-28
Fix a bug where the standing adb logcat process fails to start.

### LINKED ISSUE #190: logcat is empty for some NBU devices
When starting adb logcat, we use `-b all` to catch all buffers.

But some NBU devices (like HTC Desire 526g+) don't have the buffer option "all", and their adb logcat would only have one line saying "Unable to open log device '/dev/log/all': No such file or directory".

We need to handle this properly so we get the logs from the buffers the device does have, like 'main', 'system', 'radio', and 'events'.

xpconanfan:
List the individual buffers useful for debugging instead of using "all".
Fixes #190

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/194)

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -756,27 +756,13 @@ class AndroidDevice(object):
         f_name = 'adblog,%s,%s.txt' % (self.model, self.serial)
         utils.create_dir(self.log_path)
         logcat_file_path = os.path.join(self.log_path, f_name)
-        use_default_param = False
         try:
             extra_params = self.adb_logcat_param
         except AttributeError:
-            use_default_param = True
-            extra_params = '-b all'
+            extra_params = ''
         cmd = 'adb -s %s logcat -v threadtime %s >> %s' % (
             self.serial, extra_params, logcat_file_path)
-        try:
-            process = utils.start_standing_subprocess(
-                cmd, check_health_delay=0.1, shell=True)
-        except utils.Error:
-            # adb process did not start succesfully.
-            # If the user specificed the params, raise the error.
-            if not use_default_param:
-                raise
-            # Otherwise retry without params.
-            cmd = 'adb -s %s logcat -v threadtime >> %s' % (self.serial,
-                                                            logcat_file_path)
-            process = utils.start_standing_subprocess(
-                cmd, check_health_delay=0.1, shell=True)
+        process = utils.start_standing_subprocess(cmd, shell=True)
         self._adb_logcat_process = process
         self.adb_logcat_file_path = logcat_file_path
 
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -759,7 +759,7 @@ class AndroidDevice(object):
         try:
             extra_params = self.adb_logcat_param
         except AttributeError:
-            extra_params = '-b all'
+            extra_params = '-b main -b crash -b system -b radio'
         cmd = 'adb -s %s logcat -v threadtime %s >> %s' % (
             self.serial, extra_params, logcat_file_path)
         self._adb_logcat_process = utils.start_standing_subprocess(
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -756,14 +756,28 @@ class AndroidDevice(object):
         f_name = 'adblog,%s,%s.txt' % (self.model, self.serial)
         utils.create_dir(self.log_path)
         logcat_file_path = os.path.join(self.log_path, f_name)
+        use_default_param = False
         try:
             extra_params = self.adb_logcat_param
         except AttributeError:
-            extra_params = '-b main -b crash -b system -b radio'
+            use_default_param = True
+            extra_params = '-b all'
         cmd = 'adb -s %s logcat -v threadtime %s >> %s' % (
             self.serial, extra_params, logcat_file_path)
-        self._adb_logcat_process = utils.start_standing_subprocess(
-            cmd, shell=True)
+        try:
+            process = utils.start_standing_subprocess(
+                cmd, check_health_delay=0.1, shell=True)
+        except utils.Error:
+            # adb process did not start succesfully.
+            # If the user specificed the params, raise the error.
+            if not use_default_param:
+                raise
+            # Otherwise retry without params.
+            cmd = 'adb -s %s logcat -v threadtime >> %s' % (self.serial,
+                                                            logcat_file_path)
+            process = utils.start_standing_subprocess(
+                cmd, check_health_delay=0.1, shell=True)
+        self._adb_logcat_process = process
         self.adb_logcat_file_path = logcat_file_path
 
     def stop_adb_logcat(self):
--- a/tests/mobly/controllers/android_device_test.py
+++ b/tests/mobly/controllers/android_device_test.py
@@ -280,10 +280,9 @@ class AndroidDeviceTest(unittest.TestCase):
                                          "AndroidDevice%s" % ad.serial,
                                          "adblog,fakemodel,%s.txt" % ad.serial)
         creat_dir_mock.assert_called_with(os.path.dirname(expected_log_path))
-        adb_cmd = 'adb -s %s logcat -v threadtime -b all >> %s'
+        adb_cmd = 'adb -s %s logcat -v threadtime  >> %s'
         start_proc_mock.assert_called_with(
-              adb_cmd % (ad.serial, expected_log_path), check_health_delay=0.1,
-                         shell=True)
+              adb_cmd % (ad.serial, expected_log_path), shell=True)
         self.assertEqual(ad.adb_logcat_file_path, expected_log_path)
         expected_msg = ('Logcat thread is already running, cannot start another'
                         ' one.')
@@ -327,8 +326,7 @@ class AndroidDeviceTest(unittest.TestCase):
         creat_dir_mock.assert_called_with(os.path.dirname(expected_log_path))
         adb_cmd = 'adb -s %s logcat -v threadtime -b radio >> %s'
         start_proc_mock.assert_called_with(
-              adb_cmd % (ad.serial, expected_log_path), check_health_delay=0.1,
-                         shell=True)
+              adb_cmd % (ad.serial, expected_log_path), shell=True)
         self.assertEqual(ad.adb_logcat_file_path, expected_log_path)
 
     @mock.patch('mobly.controllers.android_device_lib.adb.AdbProxy', return_value=mock_android_device.MockAdbProxy(1))
--- a/tests/mobly/controllers/android_device_test.py
+++ b/tests/mobly/controllers/android_device_test.py
@@ -282,7 +282,8 @@ class AndroidDeviceTest(unittest.TestCase):
         creat_dir_mock.assert_called_with(os.path.dirname(expected_log_path))
         adb_cmd = 'adb -s %s logcat -v threadtime -b all >> %s'
         start_proc_mock.assert_called_with(
-              adb_cmd % (ad.serial, expected_log_path), shell=True)
+              adb_cmd % (ad.serial, expected_log_path), check_health_delay=0.1,
+                         shell=True)
         self.assertEqual(ad.adb_logcat_file_path, expected_log_path)
         expected_msg = ('Logcat thread is already running, cannot start another'
                         ' one.')
@@ -326,7 +327,8 @@ class AndroidDeviceTest(unittest.TestCase):
         creat_dir_mock.assert_called_with(os.path.dirname(expected_log_path))
         adb_cmd = 'adb -s %s logcat -v threadtime -b radio >> %s'
         start_proc_mock.assert_called_with(
-              adb_cmd % (ad.serial, expected_log_path), shell=True)
+              adb_cmd % (ad.serial, expected_log_path), check_health_delay=0.1,
+                         shell=True)
         self.assertEqual(ad.adb_logcat_file_path, expected_log_path)
 
     @mock.patch('mobly.controllers.android_device_lib.adb.AdbProxy', return_value=mock_android_device.MockAdbProxy(1))


adorokhine:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 762 at r1](https://reviewable.io:443/reviews/google/mobly/194#-Kin5wTIpvyhcpSOfSyt:-Kin5wTJNBtMJIyFu1bh:b-twcozc) ([raw file](https://github.com/google/mobly/blob/7d4fb4ef105e2e00a9657f4c9dc45c4dc953a501/mobly/controllers/android_device.py#L762)):*
> ```Python
>             extra_params = self.adb_logcat_param
>         except AttributeError:
>             extra_params = '-b main -b crash -b system -b radio'
> ```

Are we leaving out 'events' on purpose?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194)*

xpconanfan:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 762 at r1](https://reviewable.io:443/reviews/google/mobly/194#-Kin5wTIpvyhcpSOfSyt:-KinHEozjaAsJESifOf6:bjhi5gh) ([raw file](https://github.com/google/mobly/blob/7d4fb4ef105e2e00a9657f4c9dc45c4dc953a501/mobly/controllers/android_device.py#L762)):*

Are we leaving out 'events' on purpose?
</blockquote></details>

I just realized that `-b radio` will not work for tablets...
Perhaps we should drop the custom '-b' values for good and use default?...
But that means all internal users would need to supply `-b radio` to get radio logs...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194)*

adorokhine:

---

*[mobly/controllers/android_device.py, line 769 at r2](https://reviewable.io:443/reviews/google/mobly/194#-Kj5Th6VbQv3k2tRBlhw:-Kj5Th6VbQv3k2tRBlhx:b-2ypkyj) ([raw file](https://github.com/google/mobly/blob/f348704d60b0f72ce656b0376e3feaceaa03284a/mobly/controllers/android_device.py#L769)):*
> ```Python
>         try:
>             process = utils.start_standing_subprocess(
>                 cmd, check_health_delay=0.1, shell=True)
> ```

Instead of assuming 100ms is enough for logcat to start up on a potentially overloaded machine and potentially slow phone, and dropping all log buffers, is there a way to detect which buffers are available so we can solve both problems? Does it perhaps show up in logcat --help?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194)*

xpconanfan:

---

*[mobly/controllers/android_device.py, line 769 at r2](https://reviewable.io:443/reviews/google/mobly/194#-Kj5Th6VbQv3k2tRBlhw:-Kj5UlCJxa19whyjUC1-:b-2w7mwb) ([raw file](https://github.com/google/mobly/blob/f348704d60b0f72ce656b0376e3feaceaa03284a/mobly/controllers/android_device.py#L769)):*

Instead of assuming 100ms is enough for logcat to start up on a potentially overloaded machine and potentially slow phone, and dropping all log buffers, is there a way to detect which buffers are available so we can solve both problems? Does it perhaps show up in logcat --help?
</blockquote></details>

there is no reliable way.
parsing the help menu is not reliable bc each manufacturer may have different text for it.

We don't need to wait for logcat to start up though, the wait is to make sure the process didn't die.

Or could you think of other ways.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194)*

adorokhine:

---

*[mobly/controllers/android_device.py, line 769 at r2](https://reviewable.io:443/reviews/google/mobly/194#-Kj5Th6VbQv3k2tRBlhw:-Kj5VWTp0QWe-4A_lrD3:bwaslzj) ([raw file](https://github.com/google/mobly/blob/f348704d60b0f72ce656b0376e3feaceaa03284a/mobly/controllers/android_device.py#L769)):*

there is no reliable way.
parsing the help menu is not reliable bc each manufacturer may have different text for it.

We don't need to wait for logcat to start up though, the wait is to make sure the process didn't die.

Or could you think of other ways.
</blockquote></details>

Can we add the buffers to the `logcat -c` call and wait for it to complete? According to the documentation, `logcat -c` takes the same set of buffers so it should be possible to use its success as a gauge for what buffers are supported right?
https://developer.android.com/studio/command-line/logcat.html

I want to get away from time-based checks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194)*

xpconanfan:

---

*[mobly/controllers/android_device.py, line 769 at r2](https://reviewable.io:443/reviews/google/mobly/194#-Kj5Th6VbQv3k2tRBlhw:-Kj5X8lbdIYnw5sMxRDw:b-u395ce) ([raw file](https://github.com/google/mobly/blob/f348704d60b0f72ce656b0376e3feaceaa03284a/mobly/controllers/android_device.py#L769)):*

Can we add the buffers to the `logcat -c` call and wait for it to complete? According to the documentation, `logcat -c` takes the same set of buffers so it should be possible to use its success as a gauge for what buffers are supported right?
https://developer.android.com/studio/command-line/logcat.html

I want to get away from time-based checks.
</blockquote></details>

We actually can't bc the failure does not cause ret code to be non-zero........

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194)*

adorokhine:

---

*[mobly/controllers/android_device.py, line 769 at r2](https://reviewable.io:443/reviews/google/mobly/194#-Kj5Th6VbQv3k2tRBlhw:-Kj6-Y3nqYU7kijXUXdC:beebema) ([raw file](https://github.com/google/mobly/blob/f348704d60b0f72ce656b0376e3feaceaa03284a/mobly/controllers/android_device.py#L769)):*

We actually can't bc the failure does not cause ret code to be non-zero........
</blockquote></details>

What does it print in the correct and incorrect case? Anything we can key off of? I really want to move away from blanket catching all errors and doing something else. In my experience, especially when it comes to Android, the thing you "expect" went wrong is actually totally unrelated to what actually went wrong. :) 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194)*

xpconanfan:

---

*[mobly/controllers/android_device.py, line 769 at r2](https://reviewable.io:443/reviews/google/mobly/194#-Kj5Th6VbQv3k2tRBlhw:-Kj9LLZZp_DzFKO9mzNl:b-7n246i) ([raw file](https://github.com/google/mobly/blob/f348704d60b0f72ce656b0376e3feaceaa03284a/mobly/controllers/android_device.py#L769)):*

What does it print in the correct and incorrect case? Anything we can key off of? I really want to move away from blanket catching all errors and doing something else. In my experience, especially when it comes to Android, the thing you "expect" went wrong is actually totally unrelated to what actually went wrong. :) 
</blockquote></details>

It prints the help menu, which we shouldn't really rely on.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194)*

adorokhine:

---

*[mobly/controllers/android_device.py, line 769 at r2](https://reviewable.io:443/reviews/google/mobly/194#-Kj5Th6VbQv3k2tRBlhw:-KjA1vWp7SbQlXNZHUPd:basxykf) ([raw file](https://github.com/google/mobly/blob/f348704d60b0f72ce656b0376e3feaceaa03284a/mobly/controllers/android_device.py#L769)):*

It prints the help menu, which we shouldn't really rely on.
</blockquote></details>

And what does it print if it does work correctly?

My concern with this approach is twofold:
a) 100ms is not necessarily enough time to make sure logcat started on an arbitrarily slow system and phone; recall our unit test flake
b) we shouldn't blanket catching all errors and assuming the cause was anything in particular without some type of validation. I've seen cases already where our blanket catch was squashing exceptions for the wrong reason.

So I would like to find something we can do to make sure the issue is what we think it is. I would much rather have a solution that works 100% of the time on 90% of the phones (because we can always improve the check) than a solution that works 90% of the time on 100% of the phones (because people will abandon our framework).

If there's absolutely no way to do this automatically we could just use `-b all` by default and expose a config param to completely override it; let people manage it manually in case they have a non-compliant phone.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194)*

xpconanfan:

---

*[mobly/controllers/android_device.py, line 769 at r2](https://reviewable.io:443/reviews/google/mobly/194#-Kj5Th6VbQv3k2tRBlhw:-KjA36Jgtf8-r1sqc9C5:b-w4xiel) ([raw file](https://github.com/google/mobly/blob/f348704d60b0f72ce656b0376e3feaceaa03284a/mobly/controllers/android_device.py#L769)):*

And what does it print if it does work correctly?

My concern with this approach is twofold:
a) 100ms is not necessarily enough time to make sure logcat started on an arbitrarily slow system and phone; recall our unit test flake
b) we shouldn't blanket catching all errors and assuming the cause was anything in particular without some type of validation. I've seen cases already where our blanket catch was squashing exceptions for the wrong reason.

So I would like to find something we can do to make sure the issue is what we think it is. I would much rather have a solution that works 100% of the time on 90% of the phones (because we can always improve the check) than a solution that works 90% of the time on 100% of the phones (because people will abandon our framework).

If there's absolutely no way to do this automatically we could just use `-b all` by default and expose a config param to completely override it; let people manage it manually in case they have a non-compliant phone.
</blockquote></details>

It doesn't print anything on my phone if it works fine, but this is no guarantee that it won't print on other NBU devices.

Maybe we should not use '-b all' as a default in the first place.
If somebody has a special need for extra buffer like 'radio', they should use the param option to enable radio buffer collection.
For the default behavior, the default logcat option should be used, which is to not pass '-b' at all.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194)*

adorokhine:

---

*[mobly/controllers/android_device.py, line 769 at r2](https://reviewable.io:443/reviews/google/mobly/194#-Kj5Th6VbQv3k2tRBlhw:-KjA4yUN5r_m01yPfUdZ:b9lhgqa) ([raw file](https://github.com/google/mobly/blob/f348704d60b0f72ce656b0376e3feaceaa03284a/mobly/controllers/android_device.py#L769)):*

It doesn't print anything on my phone if it works fine, but this is no guarantee that it won't print on other NBU devices.

Maybe we should not use '-b all' as a default in the first place.
If somebody has a special need for extra buffer like 'radio', they should use the param option to enable radio buffer collection.
For the default behavior, the default logcat option should be used, which is to not pass '-b' at all.
</blockquote></details>

Sure, default logcat settings unless manually overridden sgtm as long as it's ok by David.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194)*

xpconanfan:

---

*[mobly/controllers/android_device.py, line 762 at r1](https://reviewable.io:443/reviews/google/mobly/194#-Kin5wTIpvyhcpSOfSyt:-KjBGRqolitKbYi38S7Q:b-896fix) ([raw file](https://github.com/google/mobly/blob/7d4fb4ef105e2e00a9657f4c9dc45c4dc953a501/mobly/controllers/android_device.py#L762)):*

I just realized that `-b radio` will not work for tablets...
Perhaps we should drop the custom '-b' values for good and use default?...
But that means all internal users would need to supply `-b radio` to get radio logs...
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 769 at r2](https://reviewable.io:443/reviews/google/mobly/194#-Kj5Th6VbQv3k2tRBlhw:-KjBGQeDCIceDFvcK4Zt:b-896fix) ([raw file](https://github.com/google/mobly/blob/f348704d60b0f72ce656b0376e3feaceaa03284a/mobly/controllers/android_device.py#L769)):*

Sure, default logcat settings unless manually overridden sgtm as long as it's ok by David.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194)*

adorokhine:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[tests/mobly/controllers/android_device_test.py, line 285 at r3](https://reviewable.io:443/reviews/google/mobly/194#-KjEyV2jFI9E3PM6o0u8:-KjEyV2jFI9E3PM6o0u9:b-4sw3pm) ([raw file](https://github.com/google/mobly/blob/6004a10e87dace8f994445a6c823e071b2b03307/tests/mobly/controllers/android_device_test.py#L285)):*
> ```Python
>         adb_cmd = 'adb -s %s logcat -v threadtime -b all >> %s'
>         start_proc_mock.assert_called_with(
>               adb_cmd % (ad.serial, expected_log_path), check_health_delay=0.1,
> ```

The continuous build seems to be failing

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194)*

xpconanfan:
Verified that this fixes the issue on the HTC device :)

---

---

*[tests/mobly/controllers/android_device_test.py, line 285 at r3](https://reviewable.io:443/reviews/google/mobly/194#-KjEyV2jFI9E3PM6o0u8:-KjFp6f1jxMochlsov8f:b-896fix) ([raw file](https://github.com/google/mobly/blob/6004a10e87dace8f994445a6c823e071b2b03307/tests/mobly/controllers/android_device_test.py#L285)):*

The continuous build seems to be failing
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194#-:-KjFp7RWqhR7K9RsHbTD:b-azqjdd)*

adorokhine:
:lgtm: Looks good. Can you make sure David is ok with this too?

---

Reviewed 1 of 1 files at r3, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194#-:-KjHDAdUPYjqkn96z8ol:b-jak5b3)*

xpconanfan:
David said notify him when this syncs.

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/194#-:-KjHFZ9N-yWvl7sXlDv_:bvrbfhz)*

Issue #83, opened 2017-01-25
Redo generated tests (parameterized tests)


xpconanfan:
Today's generated test trigger function share the same code path as a regular test, which makes accounting, reporting, and logging difficult. It also makes it hard to support selecting specific generated test cases via --test_cases.

It is time to redesign and implement the parameterized test mechanism properly.

PR #197, opened 2017-05-04
Generated tests 2.0.

xpconanfan:
* Simplified and more concise usage.
* Generated test cases can now be selected via --test_case cli arg.

Related #83 

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/197)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import collections
+import functools
 import inspect
 import logging
 import os
@@ -89,6 +91,7 @@ class BaseTestClass(object):
         self.register_controller = configs.register_controller
         self.results = records.TestResult()
         self.current_test_name = None
+        self._generated_test_table = collections.OrderedDict()
 
     def __enter__(self):
         return self
@@ -145,6 +148,17 @@ class BaseTestClass(object):
             else:
                 logging.warning('Missing optional user param "%s" in '
                                 'configuration, continue.', name)
+
+    def pre_setup(self):
+        """Preprocesses that need to be done before setup_class.
+
+        This phase is used to do pre-test processes like generating test cases.
+        This is the only place `self.generate_tests` should be called.
+
+        If this function throws an error, the test class will be marked failure
+        and the "Requested" field will be 0 because the number of tests
+        requested is unknown at this point.
+        """
 
     def _setup_class(self):
         """Proxy function to guarantee the base implementation of setup_class
@@ -375,6 +389,50 @@ class BaseTestClass(object):
             if not is_generate_trigger:
                 self.results.add_record(tr_record)
 
+    def _assert_caller_function_name(self, expected_caller_func_name):
+        """Asserts that a function only be called within a certain function.
+        """
+        current_frame = inspect.currentframe()
+        caller_frame = inspect.getouterframes(current_frame, 2)
+        if caller_frame[2][3] != expected_caller_func_name:
+            raise Error('"%s" cannot be called outside of %s' %
+                        (caller_frame[1][3], expected_caller_func_name))
+
+    def generate_tests(self, test_logic, name_func, arg_sets):
+        """Generates test cases in the test class.
+
+        This function has to be called inside a test class's `self.pre_setup`
+        function.
+
+        Generated test cases are not written down as functions, but as a list
+        of parameter sets. This way we reduce code repetition and improve
+        test case scalability.
+
+        Args:
+            test_logic: function, the common logic shared by all the generated
+                        test cases.
+            name_func: function, generate a test name according to a set of
+                       test arguments. This function should take the same
+                       arguments as the test logic function. The test name
+                       should be shorter than utils.MAX_FILENAME_LEN. Names
+                       over the limit will be truncated.
+            arg_sets: a list of tuples, each tuple is a set of arguments to be
+                      passed to the test logic function and name function.
+        """
+        self._assert_caller_function_name('pre_setup')
+        for args in arg_sets:
+            test_name = name_func(*args)
+            if not test_name.startswith('test_'):
+                raise Error(
+                    'Test name "%s" is invalid, because it does not start with'
+                    ' "test_".' % test_name)
+            if test_name in self._get_all_test_names():
+                raise Error(
+                    'Test name "%s" already exists, cannot be duplicated.' %
+                    test_name)
+            test_func = functools.partial(test_logic, *args)
+            self._generated_test_table[test_name] = test_func
+
     def run_generated_testcases(self,
                                 test_func,
                                 settings,
@@ -383,6 +441,8 @@ class BaseTestClass(object):
                                 tag='',
                                 name_func=None):
         """Runs generated test cases.
+
+        !DEPRECATED! Use self.generate_tests instead.
 
         Generated test cases are not written down as functions, but as a list
         of parameter sets. This way we reduce code repetition and improve
@@ -460,7 +520,7 @@ class BaseTestClass(object):
         for name, _ in inspect.getmembers(self, inspect.ismethod):
             if name.startswith('test_'):
                 test_names.append(name)
-        return test_names
+        return test_names + list(self._generated_test_table.keys())
 
     def _get_test_funcs(self, test_names):
         """Obtain the actual functions of test cases based on test names.
@@ -482,16 +542,18 @@ class BaseTestClass(object):
             if not test_name.startswith('test_'):
                 raise Error('Test case name %s does not follow naming '
                             'convention test_*, abort.' % test_name)
-            try:
-                test_funcs.append((test_name, getattr(self, test_name)))
-            except AttributeError:
+            if hasattr(self, test_name):
+                test_func = getattr(self, test_name)
+            elif test_name in self._generated_test_table:
+                test_func = self._generated_test_table[test_name]
+            else:
                 raise Error('%s does not have test case %s.' % (self.TAG,
                                                                 test_name))
+            test_funcs.append((test_name, test_func))
         return test_funcs
 
     def run(self, test_names=None):
-        """Runs test cases within a test class by the order they appear in the
-        execution list.
+        """Runs test cases within a test class.
 
         One of these test cases lists will be executed, shown here in priority
         order:
@@ -509,6 +571,16 @@ class BaseTestClass(object):
         Returns:
             The test results object of this class.
         """
+        # Executes pre-setup procedures, like generating test cases.
+        try:
+            self.pre_setup()
+        except Exception as e:
+            logging.exception('Pre-setup processes failed for %s.', self.TAG)
+            class_record = records.TestResultRecord('pre_setup', self.TAG)
+            class_record.test_begin()
+            class_record.test_fail(e)
+            self.results.fail_class(class_record)
+            return self.results
         logging.info('==========> %s <==========', self.TAG)
         # Devise the actual test cases to run in the test class.
         if not test_names:
@@ -520,14 +592,13 @@ class BaseTestClass(object):
                 test_names = self._get_all_test_names()
         self.results.requested = test_names
         tests = self._get_test_funcs(test_names)
-        # A TestResultRecord used for when setup_class fails.
-        class_record = records.TestResultRecord('setup_class', self.TAG)
-        class_record.test_begin()
         # Setup for the class.
         try:
             self._setup_class()
         except Exception as e:
             logging.exception('Failed to setup %s.', self.TAG)
+            class_record = records.TestResultRecord('setup_class', self.TAG)
+            class_record.test_begin()
             class_record.test_fail(e)
             self._exec_procedure_func(self._on_fail, class_record)
             self._safe_exec_func(self.teardown_class)
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -433,6 +433,12 @@ class BaseTestClass(object):
             test_func = functools.partial(test_logic, *args)
             self._generated_test_table[test_name] = test_func
 
+    def generate_tests(func, run=True):
+        def func_wrapper(name):
+            return "<p>{0}</p>".format(func(name))
+
+        return func_wrapper
+
     def run_generated_testcases(self,
                                 test_func,
                                 settings,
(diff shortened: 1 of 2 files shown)


dthkao:

---

*[mobly/base_test.py, line 152 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHQUzwzgYHO8cCFPVE:-KjHQUzwzgYHO8cCFPVF:b-u837g4) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L152)):*
> ```Python
>                                 'configuration, continue.', name)
> 
>     def pre_setup(self):
> ```

why not be specific and call this "setup_generated_tests"?

On more of a design level, are we not going with a decorator approach?

---

*[mobly/base_test.py, line 425 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHQeNGz_jdw0tw0y3K:-KjHQeNGz_jdw0tw0y3L:bie500r) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L425)):*
> ```Python
>         for args in arg_sets:
>             test_name = name_func(*args)
>             if not test_name.startswith('test_'):
> ```

did you want to add the verifier for test case name format here? or are spaces ok?

---

*[mobly/base_test.py, line 445 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHR4v4KmKd2Z8NrFCk:-KjHR4v4KmKd2Z8NrFCl:b-lg3v88) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L445)):*
> ```Python
>         """Runs generated test cases.
> 
>         !DEPRECATED! Use self.generate_tests instead.
> ```

can we add a log.warning?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/197)*

xpconanfan:

---

*[mobly/base_test.py, line 152 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHQUzwzgYHO8cCFPVE:-KjHT6uZlg-QYG3tDE9B:b-4eax0j) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L152)):*

why not be specific and call this "setup_generated_tests"?

On more of a design level, are we not going with a decorator approach?
</blockquote></details>

Because we could potentially use this for other things in the future.
I don't want to name an entire new phase after generated tests.

I don't see how could decorator help here. Can you elaborate?

---

*[mobly/base_test.py, line 425 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHQeNGz_jdw0tw0y3K:-KjHTMMxyswK4F3TqyzL:b7sgd88) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L425)):*

did you want to add the verifier for test case name format here? or are spaces ok?
</blockquote></details>

That is an issue I want to address in a separate PR with a more generic solution that also applies to other places that need test name validation like `config_parser`.

---

*[mobly/base_test.py, line 445 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHR4v4KmKd2Z8NrFCk:-KjHTdx5y_0iodT_Uc80:b-y598pt) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L445)):*

can we add a log.warning?
</blockquote></details>

I checked the usage, seems like we could remove old code paths for good?
I want to do the removal in a follow up PR to keep this one small though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/197)*

dthkao:

---

*[mobly/base_test.py, line 152 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHQUzwzgYHO8cCFPVE:-KjHUYU1sRm8JROgco3e:bku6fno) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L152)):*

Because we could potentially use this for other things in the future.
I don't want to name an entire new phase after generated tests.

I don't see how could decorator help here. Can you elaborate?
</blockquote></details>

My concern is ambiguity between pre_setup and setup_class. In order to get args for some generated tests we'd need to move a lot of the logic from setup_class to pre_setup anyway. Is there any reason the try loops for setup_class can't just be pushed ahead of fetching the tests list?

For a decorator for test functions, there are some nice points in usage. In stead of an explicit call generate tests, you could use the decorator to mark which test methods generate tests (as is done already), and if setup_class occurs before calling fetching test names, then in the reflection in _get_all_test_names we can have the logic to generate test_names (basically what's inside generate_tests at the moment).

Usage could be
@generated_test(name_function, arg_sets)
def test_do_stuff(self):
    # test logic

---

*[mobly/base_test.py, line 425 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHQeNGz_jdw0tw0y3K:-KjHvCZNzd4DIIZjhLbU:bb7ohn6) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L425)):*

That is an issue I want to address in a separate PR with a more generic solution that also applies to other places that need test name validation like `config_parser`.
</blockquote></details>

sgtm

---

*[mobly/base_test.py, line 445 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHR4v4KmKd2Z8NrFCk:-KjHvG-LLGTtfPmLYVoH:b-do2ynt) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L445)):*

I checked the usage, seems like we could remove old code paths for good?
I want to do the removal in a follow up PR to keep this one small though.
</blockquote></details>

What about external users? How about adding the log.warning and keeping it there for at least one full release before removal?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/197)*

xpconanfan:

---

*[mobly/base_test.py, line 152 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHQUzwzgYHO8cCFPVE:-KjJM1fkegwin5jY0dc5:b-g0rq1j) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L152)):*

My concern is ambiguity between pre_setup and setup_class. In order to get args for some generated tests we'd need to move a lot of the logic from setup_class to pre_setup anyway. Is there any reason the try loops for setup_class can't just be pushed ahead of fetching the tests list?

For a decorator for test functions, there are some nice points in usage. In stead of an explicit call generate tests, you could use the decorator to mark which test methods generate tests (as is done already), and if setup_class occurs before calling fetching test names, then in the reflection in _get_all_test_names we can have the logic to generate test_names (basically what's inside generate_tests at the moment).

Usage could be
@generated_test(name_function, arg_sets)
def test_do_stuff(self):
    # test logic


</blockquote></details>

Decorator does not work well for this.

People may need to write some code to create the arg_sets, e.g. a list comprehension with string formatting.
This code needs to be executed somewhere, before setup kicks off, to generate new tests.
Decorator does not provide such a venue.

The order of setup_class affects how the record is tallied, changing its order requires a lot of changes.
setup_class is also the place users register controllers today, whereas all we want is to have a pre-process that decides what tests there are.

This is actually a precursor to be able to list all the tests in a test class, which is required for CTS/GTS integration. If generate_tests is mixed into setup_class, it would be impossible to do list tests without also initializing all the controllers.

Overall, I chose this approach because it is less invasive to the current test execution flow, and leaves room for future expansion.

---

*[mobly/base_test.py, line 445 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHR4v4KmKd2Z8NrFCk:-KjJO7tltyjaK-wHYrMa:bcuzqtu) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L445)):*

What about external users? How about adding the log.warning and keeping it there for at least one full release before removal?
</blockquote></details>

Added a warning line.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/197)*

dthkao:

---

*[mobly/base_test.py, line 152 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHQUzwzgYHO8cCFPVE:-KjK57MLUqk1ZTwGKp1b:b-6n3u8z) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L152)):*

Decorator does not work well for this.

People may need to write some code to create the arg_sets, e.g. a list comprehension with string formatting.
This code needs to be executed somewhere, before setup kicks off, to generate new tests.
Decorator does not provide such a venue.

The order of setup_class affects how the record is tallied, changing its order requires a lot of changes.
setup_class is also the place users register controllers today, whereas all we want is to have a pre-process that decides what tests there are.

This is actually a precursor to be able to list all the tests in a test class, which is required for CTS/GTS integration. If generate_tests is mixed into setup_class, it would be impossible to do list tests without also initializing all the controllers.

Overall, I chose this approach because it is less invasive to the current test execution flow, and leaves room for future expansion.
</blockquote></details>

-- You can do that with a decorator still, although if you need to access instance attributes it's a little more work. I think it's worth considering the API from how a user would use it both in writing the test and is specifying it from a command/test suite. 

"This is actually a precursor to be able to list all the tests in a test class, which is required for CTS/GTS integration. If generate_tests is mixed into setup_class, it would be impossible to do list tests without also initializing all the controllers."

-- Then let's be explicit and give it a non-ambiguous name. "pre_setup" means pre-what? 

'Overall, I chose this approach because it is less invasive to the current test execution flow, and leaves room for future expansion."

-- sounds like yagni to me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/197)*

xpconanfan:

---

*[mobly/base_test.py, line 152 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHQUzwzgYHO8cCFPVE:-KjK6MV5wBoUWOlszPpl:ba6tgo8) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L152)):*

-- You can do that with a decorator still, although if you need to access instance attributes it's a little more work. I think it's worth considering the API from how a user would use it both in writing the test and is specifying it from a command/test suite. 

"This is actually a precursor to be able to list all the tests in a test class, which is required for CTS/GTS integration. If generate_tests is mixed into setup_class, it would be impossible to do list tests without also initializing all the controllers."

-- Then let's be explicit and give it a non-ambiguous name. "pre_setup" means pre-what? 

'Overall, I chose this approach because it is less invasive to the current test execution flow, and leaves room for future expansion."

-- sounds like yagni to me.
</blockquote></details>

It seems more cumbersome for users to try to fit the logic that reads user param and output the arg sets into a decorator arg. Could you give an example on what you have in mind for decorator?

If the problem is with the name, then we can name it better.

Why is choosing a less invasive approach yagni? We are talking to CTS/GTS right now for integration, so we definitely need it. I'm not sure what you mean.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/197)*

xpconanfan:

---

*[mobly/base_test.py, line 152 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHQUzwzgYHO8cCFPVE:-KjiUdMuL1nxw8ofk-ab:b-jn5s6n) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L152)):*

It seems more cumbersome for users to try to fit the logic that reads user param and output the arg sets into a decorator arg. Could you give an example on what you have in mind for decorator?

If the problem is with the name, then we can name it better.

Why is choosing a less invasive approach yagni? We are talking to CTS/GTS right now for integration, so we definitely need it. I'm not sure what you mean.
</blockquote></details>

Discussed offline.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/197)*

adorokhine:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[mobly/base_test.py, line 425 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHQeNGz_jdw0tw0y3K:-Kjla_k6vHMONNLnV2-z:b-26vcu2) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L425)):*

sgtm
</blockquote></details>

We have tests named 'Call TMobile to Sprint (5s call/5s hold)`. This check would make that name invalid. Would it be possible to avoid doing partial validation in this CL until we can fully discuss this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/197)*

xpconanfan:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[mobly/base_test.py, line 425 at r1](https://reviewable.io:443/reviews/google/mobly/197#-KjHQeNGz_jdw0tw0y3K:-KjnO6Ps69J_Txg2aFpy:b-896fix) ([raw file](https://github.com/google/mobly/blob/0a179964b1992b6583e8fc3fcce673e6511339b7/mobly/base_test.py#L425)):*

We have tests named 'Call TMobile to Sprint (5s call/5s hold)`. This check would make that name invalid. Would it be possible to avoid doing partial validation in this CL until we can fully discuss this?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/197)*

adorokhine:
:lgtm: Looks good! Mind waiting for David's approval too?

---

Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/197#-:-KjoX66gRfQ2AWt8-RQk:b-y8akqh)*

dthkao:
:lgtm:

---

Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/197#-:-KjtAn-fx5Y1GbtZub2K:bnfp4nl)*

PR #203, opened 2017-05-09
Try to reuse previous snippet/sl4a session after device disconnection

l-meng:
To execute snippet PRCs while device USB is off, need to keep snippet/sl4a session alive while disconnected:
  - Do not stop apps while device got disconnected
  - Reuse previous snippet/sl4a session if available

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/203)


xpconanfan:
Unit test failed in py3, pls fix.

adorokhine:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 474 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju0yftXCB3ZWojDO8u:-Kju0yftXCB3ZWojDO8v:bxxbjke) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L474)):*
> ```Python
>         For sample usage, see self.reboot().
>         """
>         if self._adb_logcat_process:
> ```

Can we just leave logcat running?

---

*[mobly/controllers/android_device.py, line 480 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju0uXizHvIISDSVtjk:-Kju0uXizHvIISDSVtjl:b-82f8r1) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L480)):*
> ```Python
>                 self.log.exception('Failed to stop adb logcat.')
> 
>         self._stop_sl4a_dispatcher()
> ```

I'm a little bit confused why ed needs to be torn down, but sl4a itself does not?

---

*[mobly/controllers/android_device.py, line 520 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju1FY7qXhbuWBXR_gp:-Kju1FY8li1nFGfP8Ei6:b-1xthur) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L520)):*
> ```Python
>                 # try to forward the local port to prevoius device port
>                 # also restarts snippet app if failed to connect
>                 self._start_jsonrpc_client(client)
> ```

This will throw an exception if something's wrong, won't it? eg if the device has been rebooted?

---

*[mobly/controllers/android_device.py, line 525 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju1hC766xNKeqV39ZP:-Kju1hC766xNKeqV39ZQ:b-kwlmfb) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L525)):*
> ```Python
>         return False
> 
>     def restore_sl4a(self):
> ```

There's a lot of custom handling of different services right now... maybe we can have jsonrpc_client have a restore() method and just have this thing call restore on everything?
Ang, any thoughts?

---

*[mobly/controllers/android_device.py, line 878 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju1cye_ZP4stVq1Kf6:-Kju1cyf4SGlar6rYxvf:bzfp90a) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L878)):*
> ```Python
>             self._stop_sl4a_dispatcher()
> 
>     def _stop_sl4a_dispatcher(self):
> ```

Should be called event_dispatcher

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

l-meng:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 474 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju0yftXCB3ZWojDO8u:-KjuIzRxtcKFZ7Sbetll:byaetim) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L474)):*

Can we just leave logcat running?
</blockquote></details>

Added comment. We may either lose some logs at reboot, or to get some duplicated logs if this disconnection is caused by USB off.

---

*[mobly/controllers/android_device.py, line 480 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju0uXizHvIISDSVtjk:-KjuJF4dCLlyKZMeBHa5:bekfaof) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L480)):*

I'm a little bit confused why ed needs to be torn down, but sl4a itself does not?
</blockquote></details>

Added comment: dispatcher continuously polling device, so need to stop it before disconnection

---

*[mobly/controllers/android_device.py, line 520 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju1FY7qXhbuWBXR_gp:-KjuJV-kUNxSUhgdRMWS:bzamzep) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L520)):*

This will throw an exception if something's wrong, won't it? eg if the device has been rebooted?
</blockquote></details>

The exception is handled in _start_jsonrpc_client(). It will try to reconnect from new local port to previous old port. If failed, it will create a new server session.

---

*[mobly/controllers/android_device.py, line 878 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju1cye_ZP4stVq1Kf6:-KjuK6OwDhPCer-Z6hDT:b-896fix) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L878)):*

Should be called event_dispatcher
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

l-meng:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 474 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju0yftXCB3ZWojDO8u:-KkSV-RTmMNh9O_1FgPR:b-896fix) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L474)):*
<details><summary><i>Previously, l-meng wrote…</i></summary><blockquote>

Added comment. We may either lose some logs at reboot, or to get some duplicated logs if this disconnection is caused by USB off.
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 480 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju0uXizHvIISDSVtjk:-KkSV2Heiqm_SZNRO97M:b-896fix) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L480)):*
<details><summary><i>Previously, l-meng wrote…</i></summary><blockquote>

Added comment: dispatcher continuously polling device, so need to stop it before disconnection
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 520 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju1FY7qXhbuWBXR_gp:-KkSV5Lb_ZM-YG_XIe8F:b-896fix) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L520)):*
<details><summary><i>Previously, l-meng wrote…</i></summary><blockquote>

The exception is handled in _start_jsonrpc_client(). It will try to reconnect from new local port to previous old port. If failed, it will create a new server session.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
This PR seems to have an extremely complicated commit/merge history, could we simplify that?

adorokhine:
What's the impact of that? It'll be squashed upon submit. Is it giving you trouble following the changes in reviewable?

---

Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203#-:-KkSXtVRp9eX389wKhsC:bfg33zo)*

xpconanfan:
I'm worried that the diffs may be based on out of date code since each patch seems to have a different base.
Not a huge problem, but it would be nice to straighten things out.

---

Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 525 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju1hC766xNKeqV39ZP:-KkSYifRjioRWTFScVno:bdx2bf4) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L525)):*

There's a lot of custom handling of different services right now... maybe we can have jsonrpc_client have a restore() method and just have this thing call restore on everything?
Ang, any thoughts?
</blockquote></details>

Working on a AndroidDevice service interface to rein these in ;)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203#-:-KkSYrbO1MliMILoun2d:bnshgqa)*

adorokhine:
The base is actually configured in the PR, and it's master. Github and reviewable will show you the latest commit in common with this PR's commit history as the base. So it should work ok.

---

Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203#-:-KkS_4Sn5SH0HyJ3Vwwg:b5rxzbh)*

adorokhine:
Ang: what's the plan for this PR? Are we waiting for a service interface or can it go in?

---

---

*[mobly/controllers/android_device.py, line 510 at r3](https://reviewable.io:443/reviews/google/mobly/203#-Kkr6s2EGR2sMmUtuhdQ:-Kkr6s2F9dmLb_wiCJ2C:b-9vbb16) ([raw file](https://github.com/google/mobly/blob/2f7eb6bd85ef5f2ed4c2ade67c9d54122eb761c9/mobly/controllers/android_device.py#L510)):*
> ```Python
>             self.restore_sl4a()
> 
>     def restore_snippet(self, name, package):
> ```

Is this public API intended to be called by tests? If not, should start with underscore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203#-:-Kkr7Stvt9M961UNuc6m:b6w7dow)*

l-meng:

---

*[mobly/controllers/android_device.py, line 525 at r1](https://reviewable.io:443/reviews/google/mobly/203#-Kju1hC766xNKeqV39ZP:-Kl6lPRPqF5OCJgX0h1n:b8t2cg4) ([raw file](https://github.com/google/mobly/blob/f5211eb9bf96b2899e5023b1df237c3b8e7c52b3/mobly/controllers/android_device.py#L525)):*

Working on a AndroidDevice service interface to rein these in ;)
</blockquote></details>

already merged the change and updated the code

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 403 at r4](https://reviewable.io:443/reviews/google/mobly/203#-KlRpKnYjeKZEYtSXAWp:-KlRpKnYjeKZEYtSXAWq:b-o1rn9d) ([raw file](https://github.com/google/mobly/blob/f2873f81ffe4c04179a424d84a6db777f75316bb/mobly/controllers/android_device.py#L403)):*
> ```Python
>         # names, values are the clients: {<attr name string>: <client object>}.
>         self._snippet_clients = {}
>         # Clears cached adb content.
> ```

I don't think this is added in this PR yet it's showing in diff.
Could we do a proper rebase of this PR so it's easier to review?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

adorokhine:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 403 at r4](https://reviewable.io:443/reviews/google/mobly/203#-KlRpKnYjeKZEYtSXAWp:-KlRtM3eun_KY0pzF2CP:b-l9f37p) ([raw file](https://github.com/google/mobly/blob/f2873f81ffe4c04179a424d84a6db777f75316bb/mobly/controllers/android_device.py#L403)):*

I don't think this is added in this PR yet it's showing in diff.
Could we do a proper rebase of this PR so it's easier to review?
</blockquote></details>

It was indeed added (er, moved), in this PR: https://github.com/l-meng/mobly/commit/71601cdbd4ed149a6ad37008277308f39a6e997d

As far as I can see the reviewable view is still accurate.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 403 at r4](https://reviewable.io:443/reviews/google/mobly/203#-KlRpKnYjeKZEYtSXAWp:-KlRwnDKSaUNVMSZ0-d4:bxuujj6) ([raw file](https://github.com/google/mobly/blob/f2873f81ffe4c04179a424d84a6db777f75316bb/mobly/controllers/android_device.py#L403)):*

It was indeed added (er, moved), in this PR: https://github.com/l-meng/mobly/commit/71601cdbd4ed149a6ad37008277308f39a6e997d

As far as I can see the reviewable view is still accurate.
</blockquote></details>

Ah I see.

Why are we moving this line from `start_adb_logcat` to two separate places?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 127 at r4](https://reviewable.io:443/reviews/google/mobly/203#-KlRykXrSbHyVblDQQ2W:-KlRykXrSbHyVblDQQ2X:b6th123) ([raw file](https://github.com/google/mobly/blob/f2873f81ffe4c04179a424d84a6db777f75316bb/mobly/controllers/android_device_lib/snippet_client.py#L127)):*
> ```Python
>                        self.package, time.time() - start_time, self.host_port)
> 
>     def restore_app(self, port=None):
> ```

Seems like this should have a limited scope like `restore_app_connection`?

I thought the purpose of this PR was to enable cases where device temporarily go offline during power measurement, but not when device gets rebooted, which is a different case?

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 148 at r4](https://reviewable.io:443/reviews/google/mobly/203#-KlRyRmcgxQeiFUf2BUR:-KlRyRmcgxQeiFUf2BUS:b-ort4kl) ([raw file](https://github.com/google/mobly/blob/f2873f81ffe4c04179a424d84a6db777f75316bb/mobly/controllers/android_device_lib/snippet_client.py#L148)):*
> ```Python
>                 self._connect_to_v1()
>         except:
>             # Failed to connect to app, restart app
> ```

If the app is dead, it seems like we should fail.
The purpose of this method is to "restore" the connection, if the app is not there to be restored, something wrong has happened and we should fail instead of silently relaunch the app?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

l-meng:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 510 at r3](https://reviewable.io:443/reviews/google/mobly/203#-Kkr6s2EGR2sMmUtuhdQ:-KlVDpNO1z3IGqKFw9Pg:b-896fix) ([raw file](https://github.com/google/mobly/blob/2f7eb6bd85ef5f2ed4c2ade67c9d54122eb761c9/mobly/controllers/android_device.py#L510)):*

Is this public API intended to be called by tests? If not, should start with underscore.
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 403 at r4](https://reviewable.io:443/reviews/google/mobly/203#-KlRpKnYjeKZEYtSXAWp:-KlVCyS-DdssD0ChSFxS:b-7imzuh) ([raw file](https://github.com/google/mobly/blob/f2873f81ffe4c04179a424d84a6db777f75316bb/mobly/controllers/android_device.py#L403)):*
> start_adb_logcat
To avoid duplicated log lines after device reconnected, we need to clear log in two occasions:
1. At the very first time when device initialized
2. At the time when device reconnected
We discussed about this in the very first discussion

However, that means, if an user calls start/stop logcat, test will get duplicated logs
Do we still need this line also for start_adb_logcat() ? should user be accessible to this function?

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 127 at r4](https://reviewable.io:443/reviews/google/mobly/203#-KlRykXrSbHyVblDQQ2W:-KlVE5XSAYkvNLvholNL:bcexnmb) ([raw file](https://github.com/google/mobly/blob/f2873f81ffe4c04179a424d84a6db777f75316bb/mobly/controllers/android_device_lib/snippet_client.py#L127)):*

Seems like this should have a limited scope like `restore_app_connection`?

I thought the purpose of this PR was to enable cases where device temporarily go offline during power measurement, but not when device gets rebooted, which is a different case?
</blockquote></details>

Yes, that's right, Done.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 148 at r4](https://reviewable.io:443/reviews/google/mobly/203#-KlRyRmcgxQeiFUf2BUR:-KlVErFC-dKnay58r8d_:b-896fix) ([raw file](https://github.com/google/mobly/blob/f2873f81ffe4c04179a424d84a6db777f75316bb/mobly/controllers/android_device_lib/snippet_client.py#L148)):*

If the app is dead, it seems like we should fail.
The purpose of this method is to "restore" the connection, if the app is not there to be restored, something wrong has happened and we should fail instead of silently relaunch the app?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 403 at r4](https://reviewable.io:443/reviews/google/mobly/203#-KlRpKnYjeKZEYtSXAWp:-KlWBmop9VCP7DZ_rRK-:b-xp740a) ([raw file](https://github.com/google/mobly/blob/f2873f81ffe4c04179a424d84a6db777f75316bb/mobly/controllers/android_device.py#L403)):*
<details><summary><i>Previously, l-meng wrote…</i></summary><blockquote>

> start_adb_logcat
To avoid duplicated log lines after device reconnected, we need to clear log in two occasions:
1. At the very first time when device initialized
2. At the time when device reconnected
We discussed about this in the very first discussion

However, that means, if an user calls start/stop logcat, test will get duplicated logs
Do we still need this line also for start_adb_logcat() ? should user be accessible to this function?
</blockquote></details>

In the old code, start log cat method calls '-c', which already handles both cases?

---

*[mobly/controllers/android_device.py, line 470 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWBAXe-9ABPPpQadO5:-KlWBAXe-9ABPPpQadO6:b-lmq51b) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device.py#L470)):*
> ```Python
> 
>     def _terminate_snippet(self, attr_name, client):
>         """Stops a long running service on the Android device."""
> ```

this stops a snippet, not a "long running service" right?

---

*[mobly/controllers/android_device.py, line 486 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWBaSD_dpxfy9fhF1x:-KlWBaSD_dpxfy9fhF1y:b-einzei) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device.py#L486)):*
> ```Python
>         For sample usage, see self.reboot().
>         """
>         if self._adb_logcat_process:
> ```

Why are we duplicating things from stop_services in this method?

The problem you're having is that what you need is different from what `handle_device_disconnect` was made to do.

`handle_device_disconnect` was meant for device reboot situation (it is poorly named), whereas what you want is device disconnect but not because of reboot.

This PR breaks the `AndroidDevice.reboot` usage as it will crash upon restoring service because a rebooted device does not have the snippets running so connection restoration would fail.

For changes in `AndroidDevice`, I'd suggest making separate methods for "disconnect" and "reboot" cases.
After that, you could see what pieces can be reused and consolidate the private methods.

The steps for "disconnect/reconnect" is a subset of the steps for "reboot" (no tearing down snippet app, no relaunching snippet app). So I don't think you would need to dup a lot of code.

---

*[mobly/controllers/android_device.py, line 514 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWArfDrUPkqtgpY2v9:-KlWArfDrUPkqtgpY2vA:b8sci6u) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device.py#L514)):*
> ```Python
>         # Restore snippets.
>         for attr_name, client in self._snippet_clients.items():
>             # Select a new available host port, and leaving previous device port unchanged.
> ```

why is there a comment regarding port? seems unrelated to the code here.

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 162 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWAJtwAx8UHV01uq-8:-KlWAJtxyVC_saqa133d:b-lefy57) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L162)):*
> AppResoreConnectionError
do you mean `AppResoreConnectionError`?

---

*[mobly/controllers/android_device_lib/sl4a_client.py, line 100 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWALhgbZ03BInmVSt7:-KlWALhgbZ03BInmVSt8:bk39hx8) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device_lib/sl4a_client.py#L100)):*
> ```Python
> 
>         Raises:
>             AppStartError: When the app was not able to be started.
> ```

do you mean `AppResoreConnectionError`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

l-meng:
Review status: 0 of 4 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 403 at r4](https://reviewable.io:443/reviews/google/mobly/203#-KlRpKnYjeKZEYtSXAWp:-Kl_FtUFFapAe3zsOjr_:b-j6saps) ([raw file](https://github.com/google/mobly/blob/f2873f81ffe4c04179a424d84a6db777f75316bb/mobly/controllers/android_device.py#L403)):*

In the old code, start log cat method calls '-c', which already handles both cases?
</blockquote></details>

Nope. If clear log is in start_adb_logcat(), once device is reconnected, and we call start_adb_logcat(), the log during the disconnection will be lost

---

*[mobly/controllers/android_device.py, line 486 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWBaSD_dpxfy9fhF1x:-Kl_GpeXbe3Wzg-7olmA:baie0ki) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device.py#L486)):*
>  I don't think you would need to dup a lot of code.
In the previous commits, I was thinking to handle reboot and disconnection all together, using the same method handle_device_disconnect() because they are both "disconnection" from the name of this function. In that case, this method won't break reboot.

A comment from last commit regarding "restore_app" -> "restore_app_connection", and "do not restart app if failed to restore" which potentially broke this. Will update. Before fixing this, just want to get some details:

- Do we handle reboot and disconnection together 
  - e.g., reboot is one special type (subset) of disconnection (that's my understanding, other types e.g., USB off)
  - The steps for "disconnect/reconnect" is a subset of the steps for "reboot" (according to the comment above), which is in opposite to my understanding)

- Or separate in two methods:
  - To have original handle_device_disconnect() unchanged (admittedly the name is misleading in that case)
  - And to have a different handle_device_usb_disconnect() for USB disconnection purpose?

Get to get better understanding, what's the concern/benefit to deal with reboot and USB disconnection separately, while we can deal with them with the same logic/function?

Thanks. I will fixed if we strongly think we need to deal with it separately.
(for now, I just recovered the previous version of restore_app_connection() to make it work for reboot case

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 486 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWBaSD_dpxfy9fhF1x:-Kl_WokcN6WiBS3fjA_p:b-b64r0) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device.py#L486)):*
<details><summary><i>Previously, l-meng wrote…</i></summary><blockquote>

>  I don't think you would need to dup a lot of code.
In the previous commits, I was thinking to handle reboot and disconnection all together, using the same method handle_device_disconnect() because they are both "disconnection" from the name of this function. In that case, this method won't break reboot.

A comment from last commit regarding "restore_app" -> "restore_app_connection", and "do not restart app if failed to restore" which potentially broke this. Will update. Before fixing this, just want to get some details:

- Do we handle reboot and disconnection together 
  - e.g., reboot is one special type (subset) of disconnection (that's my understanding, other types e.g., USB off)
  - The steps for "disconnect/reconnect" is a subset of the steps for "reboot" (according to the comment above), which is in opposite to my understanding)

- Or separate in two methods:
  - To have original handle_device_disconnect() unchanged (admittedly the name is misleading in that case)
  - And to have a different handle_device_usb_disconnect() for USB disconnection purpose?

Get to get better understanding, what's the concern/benefit to deal with reboot and USB disconnection separately, while we can deal with them with the same logic/function?

Thanks. I will fixed if we strongly think we need to deal with it separately.
(for now, I just recovered the previous version of restore_app_connection() to make it work for reboot case




</blockquote></details>

How would user specify which behavior is needed if they are mixed together?
Reboot and disconnect require different verification, like in disconnect, error is thrown if snippet crashed during disconnect, so users need to explicitly specify which behavior is intended by using different methods.

So it's better to have separate calls than `if` statements everywhere in the code path.
It's easier to use, test, and maintain.

To user, it would look like:
```
with ad.handle_disconnect:
   # monsoon measurement

with ad.handle_reboot:
  # reboot the device.
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

l-meng:
Review status: 0 of 4 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 470 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWBAXe-9ABPPpQadO5:-Kl_aS6GtCNZTOmUxCJv:b-896fix) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device.py#L470)):*

this stops a snippet, not a "long running service" right?
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 486 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWBaSD_dpxfy9fhF1x:-Kl_tSwppGlr3kcmMgsH:b-896fix) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device.py#L486)):*

How would user specify which behavior is needed if they are mixed together?
Reboot and disconnect require different verification, like in disconnect, error is thrown if snippet crashed during disconnect, so users need to explicitly specify which behavior is intended by using different methods.

So it's better to have separate calls than `if` statements everywhere in the code path.
It's easier to use, test, and maintain.

To user, it would look like:
```
with ad.handle_disconnect:
   # monsoon measurement

with ad.handle_reboot:
  # reboot the device.
```

</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 514 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWArfDrUPkqtgpY2v9:-Kl_ahCtkIQbf_r2cRP3:b-896fix) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device.py#L514)):*

why is there a comment regarding port? seems unrelated to the code here.
</blockquote></details>

Done.

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 162 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWAJtwAx8UHV01uq-8:-Kl_aihR-SyfmYTG3n47:b-896fix) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L162)):*

> AppResoreConnectionError
do you mean `AppResoreConnectionError`?
</blockquote></details>

Done.

---

*[mobly/controllers/android_device_lib/sl4a_client.py, line 100 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWALhgbZ03BInmVSt7:-Kl_apriVnByjQUx2N1I:b-896fix) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device_lib/sl4a_client.py#L100)):*

do you mean `AppResoreConnectionError`?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 403 at r4](https://reviewable.io:443/reviews/google/mobly/203#-KlRpKnYjeKZEYtSXAWp:-KlvABfbUpMlLoLg2xSW:b-qfsm5b) ([raw file](https://github.com/google/mobly/blob/f2873f81ffe4c04179a424d84a6db777f75316bb/mobly/controllers/android_device.py#L403)):*
<details><summary><i>Previously, l-meng wrote…</i></summary><blockquote>

Nope. If clear log is in start_adb_logcat(), once device is reconnected, and we call start_adb_logcat(), the log during the disconnection will be lost
</blockquote></details>

I don't think it is a good approach to alternate the device during construction. Then the user has no way to disable this behavior. And adding flags to constructor for this is also not elegant. Nothing in the constructor should alter the device.

Is there a better place for this?

---

*[mobly/controllers/android_device.py, line 471 at r6](https://reviewable.io:443/reviews/google/mobly/203#-KlvAuSrDAEEHUnR6FZH:-KlvAuSrDAEEHUnR6FZI:bylpi4k) ([raw file](https://github.com/google/mobly/blob/1721c4edcddc4e0bd906f64dd16c3727b0c22b67/mobly/controllers/android_device.py#L471)):*
> ```Python
>         return service_info
> 
>     def _on_usb_disconnect(self):
> ```

seems like this should just be called something similar to "stop adb logcat"?

---

*[mobly/controllers/android_device.py, line 538 at r6](https://reviewable.io:443/reviews/google/mobly/203#-Klv9DhX3kagvAG-VhVS:-Klv9DhX3kagvAG-VhVT:b-7giir) ([raw file](https://github.com/google/mobly/blob/1721c4edcddc4e0bd906f64dd16c3727b0c22b67/mobly/controllers/android_device.py#L538)):*
> ```Python
>             self.load_sl4a()
> 
>     def _connect_to_services(self):
> ```

Seems like a better name could be `_reconnect_to_services`?

---

*[mobly/controllers/android_device.py, line 539 at r6](https://reviewable.io:443/reviews/google/mobly/203#-Klv9Vp2gj-fTkK-EFZ5:-Klv9Vp2gj-fTkK-EFZ6:bys6a2q) ([raw file](https://github.com/google/mobly/blob/1721c4edcddc4e0bd906f64dd16c3727b0c22b67/mobly/controllers/android_device.py#L539)):*
> ```Python
> 
>     def _connect_to_services(self):
>         """Prepare to reconnect to services after USB reconnected."""
> ```

Is this actually "reconnecting" to services or just "preparing" to reconnect?

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 162 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWAJtwAx8UHV01uq-8:-Klv92l7gacA6a4JEJEz:beta0i2) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L162)):*
<details><summary><i>Previously, l-meng wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Still seeing `AppStartError` in r6.

---

*[mobly/controllers/android_device_lib/sl4a_client.py, line 100 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWALhgbZ03BInmVSt7:-Klv8tk1MEVjTgs2WPOO:beta0i2) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device_lib/sl4a_client.py#L100)):*
<details><summary><i>Previously, l-meng wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Still seeing `AppStartError` in r6.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

l-meng:
Review status: 0 of 4 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 403 at r4](https://reviewable.io:443/reviews/google/mobly/203#-KlRpKnYjeKZEYtSXAWp:-KlzFb1s3Cw0K0Jp4Fri:b-yxoekh) ([raw file](https://github.com/google/mobly/blob/f2873f81ffe4c04179a424d84a6db777f75316bb/mobly/controllers/android_device.py#L403)):*

I don't think it is a good approach to alternate the device during construction. Then the user has no way to disable this behavior. And adding flags to constructor for this is also not elegant. Nothing in the constructor should alter the device.

Is there a better place for this?
</blockquote></details>

Done. Passing arguments to start_services().

---

*[mobly/controllers/android_device.py, line 471 at r6](https://reviewable.io:443/reviews/google/mobly/203#-KlvAuSrDAEEHUnR6FZH:-KlzEf9aDe64QsSRG9Yg:b-896fix) ([raw file](https://github.com/google/mobly/blob/1721c4edcddc4e0bd906f64dd16c3727b0c22b67/mobly/controllers/android_device.py#L471)):*

seems like this should just be called something similar to "stop adb logcat"?
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 538 at r6](https://reviewable.io:443/reviews/google/mobly/203#-Klv9DhX3kagvAG-VhVS:-KlzEnLaRSXKY-vcf8F-:b-896fix) ([raw file](https://github.com/google/mobly/blob/1721c4edcddc4e0bd906f64dd16c3727b0c22b67/mobly/controllers/android_device.py#L538)):*

Seems like a better name could be `_reconnect_to_services`?
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 539 at r6](https://reviewable.io:443/reviews/google/mobly/203#-Klv9Vp2gj-fTkK-EFZ5:-KlzEuzlbsE6aBtv1FUG:b-896fix) ([raw file](https://github.com/google/mobly/blob/1721c4edcddc4e0bd906f64dd16c3727b0c22b67/mobly/controllers/android_device.py#L539)):*

Is this actually "reconnecting" to services or just "preparing" to reconnect?
</blockquote></details>

Done.

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 162 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWAJtwAx8UHV01uq-8:-KlzH3zAeRSPRWe6XcW6:b-896fix) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L162)):*

Still seeing `AppStartError` in r6.
</blockquote></details>

Done.

---

*[mobly/controllers/android_device_lib/sl4a_client.py, line 100 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWALhgbZ03BInmVSt7:-KlzSNFLfFXs3OIz3fmf:b-896fix) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device_lib/sl4a_client.py#L100)):*

Still seeing `AppStartError` in r6.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 470 at r7](https://reviewable.io:443/reviews/google/mobly/203#-KlzeupfECkF_DncFj6F:-KlzeupgRNouYIZ0zGEd:bn86klz) ([raw file](https://github.com/google/mobly/blob/3684c862a81962692ca4e60ffd276ed2e8b25530/mobly/controllers/android_device.py#L470)):*
> ```Python
> 
>     def _stop_logcat_process(self):
>         """Prepare for USB disconnect."""
> ```

update the docstring accordingly

---

*[mobly/controllers/android_device.py, line 479 at r7](https://reviewable.io:443/reviews/google/mobly/203#-Klzf3o_ZP5xLpFweIRk:-Klzf3o_ZP5xLpFweIRl:b34ujq0) ([raw file](https://github.com/google/mobly/blob/3684c862a81962692ca4e60ffd276ed2e8b25530/mobly/controllers/android_device.py#L479)):*
> ```Python
>     @contextlib.contextmanager
>     def handle_reboot(self):
>         """Properly manage the service life cycle when the device needs to
> ```

Need empty lines between paragraphs.

---

*[mobly/controllers/android_device.py, line 494 at r7](https://reviewable.io:443/reviews/google/mobly/203#-Klzfdo3b_ElPxS1TC34:-Klzfdo3b_ElPxS1TC35:bnscuez) ([raw file](https://github.com/google/mobly/blob/3684c862a81962692ca4e60ffd276ed2e8b25530/mobly/controllers/android_device.py#L494)):*
> ```Python
>     @contextlib.contextmanager
>     def handle_usb_disconnect(self):
>         """Properly manage the service life cycle when the device needs to
> ```

In docstrings, could you clearly point out the difference between `handle_usb_disconnect` and `handle_reboot`.
Like how `handle_usb_disconnect` assumes the device is not rebooted.

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 162 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWAJtwAx8UHV01uq-8:-KlzfMPh0S5pcBdGf1kQ:b-d9q0d7) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L162)):*
<details><summary><i>Previously, l-meng wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Still seeing `AppStartError` in r7

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

l-meng:
Review status: 0 of 4 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 470 at r7](https://reviewable.io:443/reviews/google/mobly/203#-KlzeupfECkF_DncFj6F:-Klzh8DMww73WwwSL9ab:b-896fix) ([raw file](https://github.com/google/mobly/blob/3684c862a81962692ca4e60ffd276ed2e8b25530/mobly/controllers/android_device.py#L470)):*

update the docstring accordingly
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 479 at r7](https://reviewable.io:443/reviews/google/mobly/203#-Klzf3o_ZP5xLpFweIRk:-KlzhqUo0IaibNrZNbJp:b-896fix) ([raw file](https://github.com/google/mobly/blob/3684c862a81962692ca4e60ffd276ed2e8b25530/mobly/controllers/android_device.py#L479)):*

Need empty lines between paragraphs.
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 494 at r7](https://reviewable.io:443/reviews/google/mobly/203#-Klzfdo3b_ElPxS1TC34:-KlzirTS-_HySHXLoaQh:b-896fix) ([raw file](https://github.com/google/mobly/blob/3684c862a81962692ca4e60ffd276ed2e8b25530/mobly/controllers/android_device.py#L494)):*

In docstrings, could you clearly point out the difference between `handle_usb_disconnect` and `handle_reboot`.
Like how `handle_usb_disconnect` assumes the device is not rebooted.
</blockquote></details>

Done.

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 162 at r5](https://reviewable.io:443/reviews/google/mobly/203#-KlWAJtwAx8UHV01uq-8:-KlzjBKJg9JNnSVQ0h2n:b-896fix) ([raw file](https://github.com/google/mobly/blob/629e1ce62555774afa1e338beaf03956a6f4693d/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L162)):*

Still seeing `AppStartError` in r7
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 152 at r8](https://reviewable.io:443/reviews/google/mobly/203#-KmSL_7Z-i4bIPHtnlCH:-KmSL_7_7Vq_oLhw_4mP:b98c3gb) ([raw file](https://github.com/google/mobly/blob/0e22f41dd02aab448fa57ebf35ca2a6f6479061b/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L152)):*
> ```Python
> 
>         Instead of creating new instance of the client:
>           - Find a new host_port
> ```

it finds a new port if `port` is not set, let's mention that here too.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 184 at r7](https://reviewable.io:443/reviews/google/mobly/203#-KlzjoyJeHdwISX5gcml:-KlzjoyKslFrShbFVQzI:bl6iv1r) ([raw file](https://github.com/google/mobly/blob/3684c862a81962692ca4e60ffd276ed2e8b25530/mobly/controllers/android_device_lib/snippet_client.py#L184)):*
> ```Python
>             adb_proxy=self._adb,
>             log=self.log)
>         event_client.host_port = self.host_port
> ```

these changes don't seem to be part of this PR?
Is there a rebase problem?

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 193 at r8](https://reviewable.io:443/reviews/google/mobly/203#-Km0Umu1HUkpn4hQ8QlS:-Km0Umu2g-WBj7mQ7_et:bgs6j3n) ([raw file](https://github.com/google/mobly/blob/0e22f41dd02aab448fa57ebf35ca2a6f6479061b/mobly/controllers/android_device_lib/snippet_client.py#L193)):*
> ```Python
>         """Restores previously created event client."""
>         if not self._event_client:
>             self._event_client = self._start_event_client()
> ```

Would the callback objects already created resume working after reconnect?
Could you test to make sure?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 496 at r8](https://reviewable.io:443/reviews/google/mobly/203#-KmSs2Ty7z65KbpPd5ei:-KmSs2Ty7z65KbpPd5ej:b-d5q14f) ([raw file](https://github.com/google/mobly/blob/0e22f41dd02aab448fa57ebf35ca2a6f6479061b/mobly/controllers/android_device.py#L496)):*
> ```Python
>     @contextlib.contextmanager
>     def handle_usb_disconnect(self):
>         """Properly manage the service life cycle when USB is disconnected.
> ```

Let's also mention the cool fact that the existing callback objects would still function after reconnection.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

l-meng:
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 496 at r8](https://reviewable.io:443/reviews/google/mobly/203#-KmSs2Ty7z65KbpPd5ei:-KmT2Max3wIXrfVhIuqp:b-896fix) ([raw file](https://github.com/google/mobly/blob/0e22f41dd02aab448fa57ebf35ca2a6f6479061b/mobly/controllers/android_device.py#L496)):*

Let's also mention the cool fact that the existing callback objects would still function after reconnection.
</blockquote></details>

Done.

---

*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 152 at r8](https://reviewable.io:443/reviews/google/mobly/203#-KmSL_7Z-i4bIPHtnlCH:-KmSTfb-AsRuNkQQ6THt:b-896fix) ([raw file](https://github.com/google/mobly/blob/0e22f41dd02aab448fa57ebf35ca2a6f6479061b/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L152)):*

it finds a new port if `port` is not set, let's mention that here too.
</blockquote></details>

Done.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 184 at r7](https://reviewable.io:443/reviews/google/mobly/203#-KlzjoyJeHdwISX5gcml:-KmSdRlRN3XHC4NmgykI:b-896fix) ([raw file](https://github.com/google/mobly/blob/3684c862a81962692ca4e60ffd276ed2e8b25530/mobly/controllers/android_device_lib/snippet_client.py#L184)):*

these changes don't seem to be part of this PR?
Is there a rebase problem?
</blockquote></details>

Done.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 193 at r8](https://reviewable.io:443/reviews/google/mobly/203#-Km0Umu1HUkpn4hQ8QlS:-KmSpUcBjJFL7eJwOJbO:b6tg5kx) ([raw file](https://github.com/google/mobly/blob/0e22f41dd02aab448fa57ebf35ca2a6f6479061b/mobly/controllers/android_device_lib/snippet_client.py#L193)):*

Would the callback objects already created resume working after reconnect?
Could you test to make sure?
</blockquote></details>

verified in tests cases and unit test

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
Please resolve merge conflict and fix the failing unit tests.

---

Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 498 at r9](https://reviewable.io:443/reviews/google/mobly/203#-KmTV3tkt3-P87Bs_hC5:-KmTV3tkt3-P87Bs_hC6:b-d3185a) ([raw file](https://github.com/google/mobly/blob/9d485134f47429d084e4c5af64c98b5ccf6d06e1/mobly/controllers/android_device.py#L498)):*
> user-triggered power measurement
It can also disconnect for other reasons right? Let's phrase this to be more general and give the power measurement as an example instead of making it sound like power measurement is the only situation for this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203#-:-KmTVP9V4y5SoTG8O2FA:bfor2w9)*

l-meng:
Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[mobly/controllers/android_device.py, line 498 at r9](https://reviewable.io:443/reviews/google/mobly/203#-KmTV3tkt3-P87Bs_hC5:-KmTcpOWlS7LFXyq7i6y:b-896fix) ([raw file](https://github.com/google/mobly/blob/9d485134f47429d084e4c5af64c98b5ccf6d06e1/mobly/controllers/android_device.py#L498)):*

> user-triggered power measurement
It can also disconnect for other reasons right? Let's phrase this to be more general and give the power measurement as an example instead of making it sound like power measurement is the only situation for this.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 50 at r10](https://reviewable.io:443/reviews/google/mobly/203#-KmZcL2CUyt71A8cdYEo:-KmZcL2CUyt71A8cdYEp:b-pq8wpw) ([raw file](https://github.com/google/mobly/blob/2a9e80ce85a4f594c7ee396610517d5d3cdc0fdd/mobly/controllers/android_device.py#L50)):*
> ```Python
> 
> # Default Timeout to wait for USB ON
> DEFAULT_TIMEOUT_USB_ON = 5*60
> ```

need spaces.
did you run yapf?

---

*[mobly/controllers/android_device.py, line 504 at r10](https://reviewable.io:443/reviews/google/mobly/203#-KmZcUaejuNaiN0bCwpG:-KmZcUaejuNaiN0bCwpH:b-jwoxmv) ([raw file](https://github.com/google/mobly/blob/2a9e80ce85a4f594c7ee396610517d5d3cdc0fdd/mobly/controllers/android_device.py#L504)):*
> ```Python
> 
>         The device can temporarily lose adb connection due to user-triggered
>         USB disconnection, e.g., the following cases can be handled by this
> ```

no need for comma after `e.g.`?

---

*[mobly/controllers/android_device.py, line 929 at r10](https://reviewable.io:443/reviews/google/mobly/203#-KmZd7id8zH9tE4D6ZPS:-KmZd7id8zH9tE4D6ZPT:b-k1i95n) ([raw file](https://github.com/google/mobly/blob/2a9e80ce85a4f594c7ee396610517d5d3cdc0fdd/mobly/controllers/android_device.py#L929)):*
> ```Python
>         raise DeviceError(self, 'Booting process timed out.')
> 
>     def wait_for_usb_on(self, timeout=DEFAULT_TIMEOUT_USB_ON):
> ```

`wait_for_adb_detection`?

also this logic seems very similar to `wait_for_boot_completion`, can we reduce code dup by having these two method sharing the same timeout logic?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

l-meng:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 50 at r10](https://reviewable.io:443/reviews/google/mobly/203#-KmZcL2CUyt71A8cdYEo:-Kmc8qWkCHLvjD1gsyiE:b-896fix) ([raw file](https://github.com/google/mobly/blob/2a9e80ce85a4f594c7ee396610517d5d3cdc0fdd/mobly/controllers/android_device.py#L50)):*

need spaces.
did you run yapf?
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 504 at r10](https://reviewable.io:443/reviews/google/mobly/203#-KmZcUaejuNaiN0bCwpG:-Kmc8s3YR7Hg_ACx004v:b-896fix) ([raw file](https://github.com/google/mobly/blob/2a9e80ce85a4f594c7ee396610517d5d3cdc0fdd/mobly/controllers/android_device.py#L504)):*

no need for comma after `e.g.`?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

l-meng:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 929 at r10](https://reviewable.io:443/reviews/google/mobly/203#-KmZd7id8zH9tE4D6ZPS:-KmccTUqb3SuvlHjxRGD:b-896fix) ([raw file](https://github.com/google/mobly/blob/2a9e80ce85a4f594c7ee396610517d5d3cdc0fdd/mobly/controllers/android_device.py#L929)):*

`wait_for_adb_detection`?

also this logic seems very similar to `wait_for_boot_completion`, can we reduce code dup by having these two method sharing the same timeout logic?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
Review status: 0 of 5 files reviewed at latest revision, 7 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 942 at r12](https://reviewable.io:443/reviews/google/mobly/203#-KmdyuckrWt0zU6NcIYQ:-KmdyuckrWt0zU6NcIYR:b-6jsktr) ([raw file](https://github.com/google/mobly/blob/edbf9858ebace0e92d48bcdcdab7f6ea9aa2029e/mobly/controllers/android_device.py#L942)):*
> ```Python
>         self._wait_for_device(self._usb_is_on, timeout)
> 
>     def _boot_completed(self):
> ```

`_is_boot_completed`

---

*[mobly/controllers/android_device.py, line 950 at r12](https://reviewable.io:443/reviews/google/mobly/203#-KmdywnxsVMmMYvSZ3js:-Kmdywnykgp8akoNxQKu:b-47j8fi) ([raw file](https://github.com/google/mobly/blob/edbf9858ebace0e92d48bcdcdab7f6ea9aa2029e/mobly/controllers/android_device.py#L950)):*
> ```Python
>         return False
> 
>     def _usb_is_on(self):
> ```

`_is_adb_detectable`

---

*[mobly/controllers/android_device.py, line 959 at r12](https://reviewable.io:443/reviews/google/mobly/203#-KmdybyxTqav1KdO8x9o:-KmdybyxTqav1KdO8x9p:b-3cbeja) ([raw file](https://github.com/google/mobly/blob/edbf9858ebace0e92d48bcdcdab7f6ea9aa2029e/mobly/controllers/android_device.py#L959)):*
> ```Python
> 
>     def _wait_for_device(self, func, timeout):
>         """Retry wait for device to be ready over adb until func returns True.
> ```

Hard to tell what this function is supposed to do from the description here.
Please clarify.

---

*[mobly/controllers/android_device.py, line 982 at r12](https://reviewable.io:443/reviews/google/mobly/203#-KmdzTuM0rBnAEJqsug5:-KmdzTuM0rBnAEJqsug6:b38bqpo) ([raw file](https://github.com/google/mobly/blob/edbf9858ebace0e92d48bcdcdab7f6ea9aa2029e/mobly/controllers/android_device.py#L982)):*
> ```Python
>                 pass
>             time.sleep(5)
>         raise DeviceError(self, 'Time out while waiting for device.')
> ```

Unclear message. Waiting for device what?
Let's makes sure both boot completion and adb detection methods throw descriptive messages.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

l-meng:
Review status: 0 of 5 files reviewed at latest revision, 7 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 942 at r12](https://reviewable.io:443/reviews/google/mobly/203#-KmdyuckrWt0zU6NcIYQ:-KmhKlBTiy1XvDeAq4So:b-896fix) ([raw file](https://github.com/google/mobly/blob/edbf9858ebace0e92d48bcdcdab7f6ea9aa2029e/mobly/controllers/android_device.py#L942)):*

`_is_boot_completed`
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 950 at r12](https://reviewable.io:443/reviews/google/mobly/203#-KmdywnxsVMmMYvSZ3js:-KmhKlY-WBzG95eh-jH6:b-896fix) ([raw file](https://github.com/google/mobly/blob/edbf9858ebace0e92d48bcdcdab7f6ea9aa2029e/mobly/controllers/android_device.py#L950)):*

`_is_adb_detectable`
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 959 at r12](https://reviewable.io:443/reviews/google/mobly/203#-KmdybyxTqav1KdO8x9o:-KmhMmtMR05MRw90D1nj:b-896fix) ([raw file](https://github.com/google/mobly/blob/edbf9858ebace0e92d48bcdcdab7f6ea9aa2029e/mobly/controllers/android_device.py#L959)):*

Hard to tell what this function is supposed to do from the description here.
Please clarify.
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 982 at r12](https://reviewable.io:443/reviews/google/mobly/203#-KmdzTuM0rBnAEJqsug5:-KmhNfTY7DbX-sCSjR0X:b-dsm6zo) ([raw file](https://github.com/google/mobly/blob/edbf9858ebace0e92d48bcdcdab7f6ea9aa2029e/mobly/controllers/android_device.py#L982)):*

Unclear message. Waiting for device what?
Let's makes sure both boot completion and adb detection methods throw descriptive messages.
</blockquote></details>

Updated with function names, is that good?

Since this function wraps different functions for different purposes, it's hard to predict what the args func does exactly, and provide a detailed descriptive message.

As a wrapper, maybe this function should not know what the passed in function does, but just retries it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203)*

xpconanfan:
:lgtm:

Leaving final stamping to David to make sure sl4a client is ok.

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/203#-:-KmijJhEllltoFuPGSr-:b-13b1wi)*

Issue #172, opened 2017-03-30
Add suite_runner documentation to the tutorial


adorokhine:

xpconanfan:
Closing this as `test_runner` can be used freely without `suite_runner` now

PR #205, opened 2017-05-10
Adjust test_runner API to allow associating different configs with test classes.

adorokhine:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/205)

### DIFF
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -23,6 +23,10 @@ from mobly import signals
 from mobly import test_runner
 
 
+class Error(Exception):
+    pass
+
+
 def run_suite(test_classes, argv=None):
     """Execute the test suite in a module.
 
@@ -68,73 +72,79 @@ def run_suite(test_classes, argv=None):
     test_configs = config_parser.load_test_config_file(args.config[0])
 
     # Check the classes that were passed in
-    for cls in test_classes:
-        if not issubclass(cls, base_test.BaseTestClass):
+    for test_class in test_classes:
+        if not issubclass(test_class, base_test.BaseTestClass):
             logging.error('Test class %s does not extend '
-                          'mobly.base_test.BaseTestClass',
-                          cls)
+                          'mobly.base_test.BaseTestClass', test_class)
             sys.exit(1)
 
-    # Choose which tests to run
-    test_identifiers = _compute_test_identifiers(test_classes, args.tests)
+    # Find the full list of tests to execute
+    selected_tests = _compute_selected_tests(test_classes, args.tests)
 
     # Execute the suite
     ok = True
     for config in test_configs:
-        runner = test_runner.TestRunner(config, test_identifiers)
+        runner = test_runner.TestRunner(config.log_path, config.test_bed_name)
+        for (test_class, tests) in selected_tests.items():
+            runner.add_test_class(config, test_class, tests)
         try:
-            try:
-                runner.run(test_classes)
-                ok = runner.results.is_all_pass and ok
-            except signals.TestAbortAll:
-                pass
-            except:
-                logging.exception('Exception when executing %s.',
-                                  runner.test_configs.test_bed_name)
-                ok = False
-        finally:
-            runner.stop()
+            runner.run()
+            ok = runner.results.is_all_pass and ok
+        except signals.TestAbortAll:
+            pass
+        except:
+            logging.exception('Exception when executing %s.',
+                              config.test_bed_name)
+            ok = False
     if not ok:
         sys.exit(1)
 
 
-def _compute_test_identifiers(test_classes, selected_tests):
-    """Computes a list of test identifiers for TestRunner from list of strings.
+def _compute_selected_tests(test_classes, selected_tests):
+    """Computes class_to_test_name to run for each class from selector strings.
 
     Args:
         test_classes: (list of class) all classes that are part of this suite.
-        selected_tests: (list of string) list of tests to execute, eg:
+        selected_tests: (list of string) list of class_to_test_name to execute, eg:
              ['FooTest', 'BarTest',
               'BazTest.test_method_a', 'BazTest.test_method_b'].
-             May be empty, in which case all tests are selected.
+             May be empty, in which case all class_to_test_name are selected.
 
     Returns:
-        list(tuple(test_class_name, list(test_name))),
+        dict: test class -> list(test name):
         identifiers for TestRunner. For the above example:
-        [
-            ('FooTest', None),
-            ('BarTest', None),
-            ('BazTest', ['test_method_a', 'test_method_b']),
-        ]
+        {
+            FooTest: None,
+            BarTest: None,
+            BazTest: ['test_method_a', 'test_method_b'],
+        }
     """
-    # Create a map from test class name to list of test names
-    test_identifier_builder = collections.OrderedDict()
+    # Map from test class name to instance.
+    class_name_lookup = {cls.__name__: cls for cls in test_classes}
+
+    # Map from test class to list of class_to_test_name to execute.
+    class_to_test_name = collections.OrderedDict()
     if selected_tests:
         for test in selected_tests:
             if '.' in test:  # Has a test method
                 (test_class_name, test_name) = test.split('.')
-                if test_class_name not in test_identifier_builder:
-                    # Never seen this class before
-                    test_identifier_builder[test_class_name] = [test_name]
-                elif test_identifier_builder[test_class_name] is None:
-                    # Already running all tests in this class, so ignore this
-                    # extra test.
-                    pass
-                else:
-                    test_identifier_builder[test_class_name].append(test_name)
-            else:  # No test method; run all tests in this class.
-                test_identifier_builder[test] = None
+            else:
+                (test_class_name, test_name) = (test, None)
+            test_class = class_name_lookup.get(test_class_name)
+            if not test_class:
+                raise Error('Unknown test class %s' % test_class_name)
+            if test_name and test_class not in class_to_test_name:
+                # Never seen this class before
+                class_to_test_name[test_class] = [test_name]
+            elif test_name and class_to_test_name[test_class] is None:
+                # Already running all tests in this class, so ignore this extra
+                # test.
+                pass
+            elif test_name:
+                class_to_test_name[test_class].append(test_name)
+            else:  # No test method; run all class_to_test_name in this class.
+                class_to_test_name[test_class] = None
     else:
-        for test_class_name in test_classes:
-            test_identifier_builder[test_class_name.__name__] = None
-    return list(test_identifier_builder.items())
+        for test_class in test_classes:
+            class_to_test_name[test_class] = None
+    return class_to_test_name
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -106,7 +106,7 @@ def run_suite(test_classes, argv=None):
 def _compute_selected_tests(test_classes, selected_tests):
     """Computes tests to run for each class from selector strings.
 
-    This method transforms a list of selector strings (such as FooTest or
+    This function transforms a list of selector strings (such as FooTest or
     FooTest.test_method_a) to a dict where keys are test_name classes, and values are
     lists of selected tests in those classes. None means all tests in that class
     are selected.
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -101,17 +101,22 @@ def run_suite(test_classes, argv=None):
 
 
 def _compute_selected_tests(test_classes, selected_tests):
-    """Computes class_to_test_name to run for each class from selector strings.
+    """Computes tests to run for each class from selector strings.
+
+    This method transforms a list of selector strings (such as FooTest or
+    FooTest.test_method_a) to a dict where keys are test_name classes, and values are
+    lists of selected tests in those classes. None means all tests in that class
+    are selected.
 
     Args:
         test_classes: (list of class) all classes that are part of this suite.
-        selected_tests: (list of string) list of class_to_test_name to execute, eg:
+        selected_tests: (list of string) list of tests to execute, eg:
              ['FooTest', 'BarTest',
               'BazTest.test_method_a', 'BazTest.test_method_b'].
-             May be empty, in which case all class_to_test_name are selected.
+             May be empty, in which case all tests in the class are selected.
 
     Returns:
-        dict: test class -> list(test name):
+        dict: test_name class -> list(test_name name):
         identifiers for TestRunner. For the above example:
         {
             FooTest: None,
@@ -119,32 +124,39 @@ def _compute_selected_tests(test_classes, selected_tests):
             BazTest: ['test_method_a', 'test_method_b'],
         }
     """
-    # Map from test class name to instance.
-    class_name_lookup = {cls.__name__: cls for cls in test_classes}
+    class_to_tests = collections.OrderedDict()
+    if not selected_tests:
+        # No selection is needed; simply run all tests in all classes.
+        for test_class in test_classes:
+            class_to_tests[test_class] = None
+        return class_to_tests
 
-    # Map from test class to list of class_to_test_name to execute.
-    class_to_test_name = collections.OrderedDict()
-    if selected_tests:
-        for test in selected_tests:
-            if '.' in test:  # Has a test method
-                (test_class_name, test_name) = test.split('.')
-            else:
-                (test_class_name, test_name) = (test, None)
-            test_class = class_name_lookup.get(test_class_name)
-            if not test_class:
-                raise Error('Unknown test class %s' % test_class_name)
-            if test_name and test_class not in class_to_test_name:
+    # The user is selecting some tests to run. Parse the selectors.
+    # Dict from test_name class name to list of tests to execute (or None for all
+    # tests).
+    test_class_name_to_tests = collections.OrderedDict()
+    for test_name in selected_tests:
+        if '.' in test_name:  # Has a test method
+            (test_class_name, test_name) = test_name.split('.')
+            if test_class_name not in test_class_name_to_tests:
                 # Never seen this class before
-                class_to_test_name[test_class] = [test_name]
-            elif test_name and class_to_test_name[test_class] is None:
+                test_class_name_to_tests[test_class_name] = [test_name]
+            elif test_class_name_to_tests[test_class_name] is None:
                 # Already running all tests in this class, so ignore this extra
                 # test.
                 pass
-            elif test_name:
-                class_to_test_name[test_class].append(test_name)
-            else:  # No test method; run all class_to_test_name in this class.
-                class_to_test_name[test_class] = None
-    else:
-        for test_class in test_classes:
-            class_to_test_name[test_class] = None
-    return class_to_test_name
+            else:
+                test_class_name_to_tests[test_class_name].append(test_name)
+        else:  # No test method; run all tests in this class.
+            test_class_name_to_tests[test_name] = None
+
+    # Now transform class names to class objects.
+    # Dict from test_name class name to instance.
+    class_name_to_class = {cls.__name__: cls for cls in test_classes}
+    for test_class_name, tests in test_class_name_to_tests.items():
+        test_class = class_name_to_class.get(test_class_name)
+        if not test_class:
+            raise Error('Unknown test_name class %s' % test_class_name)
+        class_to_tests[test_class] = tests
+
+    return class_to_tests
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -14,7 +14,7 @@
 """Runner for Mobly test suites.
 
 To create a test suite, call suite_runner.run_suite() with one or more
-individual testcases. For example:
+individual test classes. For example:
 
     from mobly import suite_runner
 
@@ -22,7 +22,7 @@ individual testcases. For example:
     from my.test.lib import bar_test
     ...
     if __name__ == '__main__':
-        suite_runner.run(foo_test.FooTest, bar_test.BarTest)
+        suite_runner.run_suite(foo_test.FooTest, bar_test.BarTest)
 """
 
 import argparse
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -11,6 +11,19 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""Runner for Mobly test suites.
+
+To create a test suite, call suite_runner.run_suite() with one or more
+individual testcases. For example:
+
+    from mobly import suite_runner
+
+    from my.test.lib import foo_test
+    from my.test.lib import bar_test
+    ...
+    if __name__ == '__main__':
+        suite_runner.run(foo_test.FooTest, bar_test.BarTest)
+"""
 
 import argparse
 import collections
@@ -28,20 +41,10 @@ class Error(Exception):
 
 
 def run_suite(test_classes, argv=None):
-    """Execute the test suite in a module.
+    """Executes multiple test classes as a suite.
 
     This is the default entry point for running a test suite script file
     directly.
-
-    To make your test suite executable, add the following to your file:
-
-        from mobly import suite_runner
-
-        from my.test.lib import foo_test
-        from my.test.lib import bar_test
-        ...
-        if __name__ == '__main__':
-            suite_runner.run(foo_test.FooTest, bar_test.BarTest)
 
     Args:
         test_classes: List of python classes containing Mobly tests.
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -242,7 +242,8 @@ class TestRunner(object):
             config: config_parser.TestRunConfig, configuration to execute this
                     test class with.
             test_class: class, test class to execute.
-            tests: Optional list of test names within the class to execute.
+            tests: list of strings, optional list of test names within the
+                   class to execute.
 
         Raises:
             Error: if the provided config has a log_path or test_bed_name which
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -83,14 +83,14 @@ def main(argv=None):
     # Find the test class in the test script.
     test_class = _find_test_class()
     # Parse test specifiers if exist.
-    test_names = None
+    tests = None
     if args.tests:
-        test_names = args.tests
+        tests = args.tests
     # Execute the test class with configs.
     ok = True
     for config in test_configs:
         runner = TestRunner(config.log_path, config.test_bed_name)
-        runner.add_test_class(config, test_class, test_names)
+        runner.add_test_class(config, test_class, tests)
         try:
             runner.run()
             ok = runner.results.is_all_pass and ok
@@ -126,17 +126,57 @@ def _find_test_class():
     return test_classes[0]
 
 
-def _verify_controller_module(module):
+def verify_controller_module(module):
     """Verifies a module object follows the required interface for
     controllers.
+
+    A Mobly controller module is a Python lib that can be used to control
+    a device, service, or equipment. To be Mobly compatible, a controller
+    module needs to have the following members:
+
+        def create(configs):
+            [Required] Creates controller objects from configurations.
+            Args:
+                configs: A list of serialized data like string/dict. Each
+                         element of the list is a configuration for a
+                         controller object.
+            Returns:
+                A list of objects.
+
+        def destroy(objects):
+            [Required] Destroys controller objects created by the create
+            function. Each controller object shall be properly cleaned up
+            and all the resources held should be released, e.g. memory
+            allocation, sockets, file handlers etc.
+            Args:
+                A list of controller objects created by the create function.
+
+        def get_info(objects):
+            [Optional] Gets info from the controller objects used in a test
+            run. The info will be included in test_result_summary.json under
+            the key 'ControllerInfo'. Such information could include unique
+            ID, version, or anything that could be useful for describing the
+            test bed and debugging.
+            Args:
+                objects: A list of controller objects created by the create
+                         function.
+            Returns:
+                A list of json serializable objects, each represents the
+                info of a controller object. The order of the info object
+                should follow that of the input objects.
+
+    Registering a controller module declares a test class's dependency the
+    controller. If the module config exists and the module matches the
+    controller interface, controller objects will be instantiated with
+    corresponding configs. The module should be imported first.
 
     Args:
         module: An object that is a controller module. This is usually
                 imported with import statements or loaded by importlib.
 
     Raises:
-        ControllerError is raised if the module does not match the Mobly
-        controller interface, or one of the required members is null.
+        ControllerError: if the module does not match the Mobly controller
+            interface, or one of the required members is null.
     """
     required_attributes = ('create', 'destroy', 'MOBLY_CONTROLLER_CONFIG_NAME')
     for attr in required_attributes:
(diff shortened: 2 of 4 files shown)


xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[mobly/test_runner.py, line 93 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjkVg87bEa5qN6RKQQ:-KjjkVg87bEa5qN6RKQR:bx350qf) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L93)):*
> ```Python
>     for config in test_configs:
>         runner = TestRunner(config.log_path, config.test_bed_name)
>         runner.add_test_class(config, test_class, test_case_names)
> ```

So the structure of `config` here is a bit confusing.

Could we restructure/rename here to clarify what fields need to be passed during `TestRunner construction` and what need to be added by `add_test_class`?

It looks like log_path is the only one needed for construction?

---

*[mobly/test_runner.py, line 162 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjkUMgL9gOHUwTLMGy:-KjjkUMgL9gOHUwTLMGz:b1grlqg) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L162)):*
> ```Python
>     """
> 
>     def __init__(self, log_dir, test_bed_name):
> ```

Maybe the test bed name doesn't need to be part of the constructor?
How about make it part of the test config passed to `add_test_class`?

---

*[mobly/test_runner.py, line 173 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjiPBj-Eav5cj6obRg:-KjjiPBkEsDMQRXyIQDR:b-1qv75t) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L173)):*
> ```Python
> 
>         self.results = records.TestResult()
>         self._test_infos = []
> ```

Maybe a better name is `test_run_infos` since this contains
Can we have some documentation on what info this var supposed to have? Or maybe we could create a class for it as we did with `test_config`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[mobly/test_runner.py, line 129 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjmGH5UmyRehQeCd2X:-KjjmGH5UmyRehQeCd2Y:blltp6r) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L129)):*
> ```Python
> 
> 
> def _verify_controller_module(module):
> ```

can we make this public?
I can see a user may want to use this in their unit test to validate their custom module.

Also, how about moving the interface docstring from `register_controller` to this function? Seems a bit cleaner since after all this function is the guardian of the "interface"...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

adorokhine:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[mobly/test_runner.py, line 93 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjkVg87bEa5qN6RKQQ:-Kjk63VRZ4g1jB_DVOgT:b-yhr3il) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L93)):*

So the structure of `config` here is a bit confusing.

Could we restructure/rename here to clarify what fields need to be passed during `TestRunner construction` and what need to be added by `add_test_class`?

It looks like log_path is the only one needed for construction?
</blockquote></details>

log_path and test_bed_name are needed during construction. What would you like me to rename them to?

---

*[mobly/test_runner.py, line 129 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjmGH5UmyRehQeCd2X:-Kjk8JJqbCACDbADqVk4:b-896fix) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L129)):*

can we make this public?
I can see a user may want to use this in their unit test to validate their custom module.

Also, how about moving the interface docstring from `register_controller` to this function? Seems a bit cleaner since after all this function is the guardian of the "interface"...
</blockquote></details>

Done.

---

*[mobly/test_runner.py, line 162 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjkUMgL9gOHUwTLMGy:-KjjmbiN-3JNAg8T0wS0:b-9xalnq) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L162)):*

Maybe the test bed name doesn't need to be part of the constructor?
How about make it part of the test config passed to `add_test_class`?
</blockquote></details>

The testbed name is part of the logpath and a single test_runner outputs to a single logpath. So everything needed to compute the logpath needs to be part of the constructor.

---

*[mobly/test_runner.py, line 173 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjiPBj-Eav5cj6obRg:-Kjk9xijpiI4Oy2P_Ihm:b-896fix) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L173)):*

Maybe a better name is `test_run_infos` since this contains
Can we have some documentation on what info this var supposed to have? Or maybe we could create a class for it as we did with `test_config`?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[mobly/test_runner.py, line 162 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjkUMgL9gOHUwTLMGy:-KjkAKGtzKDBvJpKHg2A:b97dppk) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L162)):*

The testbed name is part of the logpath and a single test_runner outputs to a single logpath. So everything needed to compute the logpath needs to be part of the constructor.
</blockquote></details>

I see the `log_path` is computed in `run` for every iteration so i don't think test bed name needs to be in constructor?

Also, we could move the log_path calculation to `add_test_class` since there's no new info added for this calculation during run time?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

adorokhine:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[mobly/test_runner.py, line 162 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjkUMgL9gOHUwTLMGy:-KjoVW4xMsH2TpHJdnDQ:b-dvwm8k) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L162)):*
`run()` is only called one time, to run all the classes that have been added in one big batch. After that the TestRunner life cycle is completed.

The final log path calculation is done there because timestamp is part of log path, and there might be an arbitrarily large amount of time between constructing the test runner, adding classes, and finally calling run.

> we could move the log_path calculation to add_test_class

Each call to add_test_class has a different config, so it would become ambiguous which one's config would be used for logging. Rather than pick something arbitrary like the first test class's config is used for logging, I've decided to move those pieces to the constructor so there is no ambiguity.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[mobly/suite_runner.py, line 103 at r2](https://reviewable.io:443/reviews/google/mobly/205#-KjxtNxRllPVHJNR1HFj:-KjxtNxSGnd68rjcg_GC:b-7b5o26) ([raw file](https://github.com/google/mobly/blob/e49bfed417d728839e24e47ac13d2353a07f0976/mobly/suite_runner.py#L103)):*
> ```Python
> 
> 
> def _compute_selected_testcases(test_classes, selected_test_cases):
> ```

this function seems to be getting more complicated.
Could we add in docstring the high level behavior in terms of how test cases are selected by this function, and add unit tests for this?

---

*[mobly/test_runner.py, line 162 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjkUMgL9gOHUwTLMGy:-Kjy5I9GLhu4YQjUDL3Q:b-4xocuw) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L162)):*

`run()` is only called one time, to run all the classes that have been added in one big batch. After that the TestRunner life cycle is completed.

The final log path calculation is done there because timestamp is part of log path, and there might be an arbitrarily large amount of time between constructing the test runner, adding classes, and finally calling run.

> we could move the log_path calculation to add_test_class

Each call to add_test_class has a different config, so it would become ambiguous which one's config would be used for logging. Rather than pick something arbitrary like the first test class's config is used for logging, I've decided to move those pieces to the constructor so there is no ambiguity.
</blockquote></details>

There is one thing that's not clear to me in this, what's our goal with this change?
I thought we are making suite so multiple test class runs with different configs and generate a single set of files. Is that correct?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

adorokhine:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[mobly/suite_runner.py, line 103 at r2](https://reviewable.io:443/reviews/google/mobly/205#-KjxtNxRllPVHJNR1HFj:-KjyefmUjvd6yFwNogpA:b3u7fn4) ([raw file](https://github.com/google/mobly/blob/e49bfed417d728839e24e47ac13d2353a07f0976/mobly/suite_runner.py#L103)):*
> this function seems to be getting more complicated.

Tried to restructure the code a bit and add more comments to make it clearer.

> Could we add in docstring the high level behavior in terms of how test cases are selected by this function

Done.

> and add unit tests for this?

This function already has pretty extensive unit tests. Are there any testcases you can think of that don't exist yet?

---

*[mobly/test_runner.py, line 162 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjkUMgL9gOHUwTLMGy:-Kjyf5v9nTAp57ZjFYG2:be8uvd1) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L162)):*

There is one thing that's not clear to me in this, what's our goal with this change?
I thought we are making suite so multiple test class runs with different configs and generate a single set of files. Is that correct?
</blockquote></details>

Yep, that's exactly the goal (to make it possible to run multiple test classes with different configs and generate a single set of files).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[mobly/suite_runner.py, line 36 at r3](https://reviewable.io:443/reviews/google/mobly/205#-KkDKP-GCoHl3S3arw-P:-KkDKP-GCoHl3S3arw-Q:b-m2olkn) ([raw file](https://github.com/google/mobly/blob/9530cfb12978dd322978b2fbbd0a3f55ece9084f/mobly/suite_runner.py#L36)):*
> ```Python
>     directly.
> 
>     To make your test suite executable, add the following to your file:
> ```

Could we add a module level usage docstring and move this example there?

Also do we want to expose the usage for the new multi-config feature in this module?

Overall I think we should have a place to document all the usages related to suites (#172)
Doesn't have to be in this PR.

---

*[mobly/test_runner.py, line 93 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjkVg87bEa5qN6RKQQ:-KkDJ3iqOrxurPZykHZ4:by986f2) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L93)):*

log_path and test_bed_name are needed during construction. What would you like me to rename them to?
</blockquote></details>

I guess not much we could do...
Could we use kwargs in the `TestRunner` constructor here.

---

*[mobly/test_runner.py, line 162 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjkUMgL9gOHUwTLMGy:-KkDH1BByhVmNWEudX-u:bctpwjn) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L162)):*

Yep, that's exactly the goal (to make it possible to run multiple test classes with different configs and generate a single set of files).
</blockquote></details>

Ok, in that case, could we:
* In `TestRunner` docstring, clarifiy the expectation that one `TestRunner`  is associated one specific test bed, and one `TestRunner.run` generates one single set of output files for all classes and tests.
* Each config object passed to `add_test_class` may have a different `log_path` specified. If `add_test_class` gets a config with a log path different from the TestRunner's, throw an error.

---

*[mobly/test_runner.py, line 266 at r3](https://reviewable.io:443/reviews/google/mobly/205#-KkDOaf3p9OOFB5sKktZ:-KkDOaf4kgrI8X9THnMa:b-mc81o3) ([raw file](https://github.com/google/mobly/blob/9530cfb12978dd322978b2fbbd0a3f55ece9084f/mobly/test_runner.py#L266)):*
> ```Python
>                 raise e
> 
>     def run(self):
> ```

should we add a check here to throw an error if no class has been added?

---

*[tests/mobly/test_runner_test.py, line 206 at r3](https://reviewable.io:443/reviews/google/mobly/205#-KkDNJT_aQnbXRAaMMBU:-KkDNJT_aQnbXRAaMMBV:bgpzxhy) ([raw file](https://github.com/google/mobly/blob/9530cfb12978dd322978b2fbbd0a3f55ece9084f/tests/mobly/test_runner_test.py#L206)):*
> ```Python
>         }]
>         tr = test_runner.TestRunner(self.log_dir, self.test_bed_name)
>         tr.add_test_class(mock_test_config, integration2_test.Integration2Test)
> ```

can we have a test where the two configs are different and verify that each test is run with the corresponding config?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

adorokhine:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[mobly/suite_runner.py, line 36 at r3](https://reviewable.io:443/reviews/google/mobly/205#-KkDKP-GCoHl3S3arw-P:-KkDjQ1s0DGLx-m746Ql:bo4jpww) ([raw file](https://github.com/google/mobly/blob/9530cfb12978dd322978b2fbbd0a3f55ece9084f/mobly/suite_runner.py#L36)):*
> Could we add a module level usage docstring and move this example there?

Done.

> Also do we want to expose the usage for the new multi-config feature in this module?

Right now there's no way to do selection in that kind of environment (because there could be multiple `test_foo`s floating around with different configs), so I think we'd need to design what that would look like.

> Overall I think we should have a place to document all the usages related to suites (#172). Doesn't have to be in this PR.

This is another thing that might change if we want to add selection capability, so how about putting it off until the API stabilizes?

---

*[mobly/test_runner.py, line 93 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjkVg87bEa5qN6RKQQ:-KkDkPtPw752VOchqMTH:b-896fix) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L93)):*

I guess not much we could do...
Could we use kwargs in the `TestRunner` constructor here.


</blockquote></details>

Done.

---

*[mobly/test_runner.py, line 162 at r1](https://reviewable.io:443/reviews/google/mobly/205#-KjjkUMgL9gOHUwTLMGy:-KkDn4E5QW-haQS5QA7c:b-896fix) ([raw file](https://github.com/google/mobly/blob/2c2f4ca6a9baf2d0c7e7556163c9d98e095b8336/mobly/test_runner.py#L162)):*

Ok, in that case, could we:
* In `TestRunner` docstring, clarifiy the expectation that one `TestRunner`  is associated one specific test bed, and one `TestRunner.run` generates one single set of output files for all classes and tests.
* Each config object passed to `add_test_class` may have a different `log_path` specified. If `add_test_class` gets a config with a log path different from the TestRunner's, throw an error.
</blockquote></details>

Done.

---

*[mobly/test_runner.py, line 266 at r3](https://reviewable.io:443/reviews/google/mobly/205#-KkDOaf3p9OOFB5sKktZ:-KkDnivjO4GIxs6rCwPd:b-896fix) ([raw file](https://github.com/google/mobly/blob/9530cfb12978dd322978b2fbbd0a3f55ece9084f/mobly/test_runner.py#L266)):*

should we add a check here to throw an error if no class has been added?
</blockquote></details>

Done.

---

*[tests/mobly/test_runner_test.py, line 206 at r3](https://reviewable.io:443/reviews/google/mobly/205#-KkDNJT_aQnbXRAaMMBU:-KkE3jOZEarH3gWZS9QA:b-896fix) ([raw file](https://github.com/google/mobly/blob/9530cfb12978dd322978b2fbbd0a3f55ece9084f/tests/mobly/test_runner_test.py#L206)):*

can we have a test where the two configs are different and verify that each test is run with the corresponding config?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/test_runner.py, line 292 at r4](https://reviewable.io:443/reviews/google/mobly/205#-KkE57nxNQxXMn012yFO:-KkE57nxNQxXMn012yFP:b-x26iz9) ([raw file](https://github.com/google/mobly/blob/b93f3b6258547f9cdbbd72ac3264b43e2e7ded88/mobly/test_runner.py#L292)):*
> ```Python
>         """
>         if not self._test_run_infos:
>             raise Error('No tests to execute.')
> ```

Missing docstring for this

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

adorokhine:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/test_runner.py, line 292 at r4](https://reviewable.io:443/reviews/google/mobly/205#-KkE57nxNQxXMn012yFO:-KkE8tu-ai3UXwrckcrW:b-896fix) ([raw file](https://github.com/google/mobly/blob/b93f3b6258547f9cdbbd72ac3264b43e2e7ded88/mobly/test_runner.py#L292)):*

Missing docstring for this
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

xpconanfan:

---

*[mobly/test_runner.py, line 292 at r4](https://reviewable.io:443/reviews/google/mobly/205#-KkE57nxNQxXMn012yFO:-KkH69arrDgQnEeIHtmF:bdoth13) ([raw file](https://github.com/google/mobly/blob/b93f3b6258547f9cdbbd72ac3264b43e2e7ded88/mobly/test_runner.py#L292)):*

Done.
</blockquote></details>

Traditionally we've been indenting to after ": " in docstrings, could we follow the same convention here?
Here and elsewhere.

---

*[mobly/test_runner.py, line 211 at r5](https://reviewable.io:443/reviews/google/mobly/205#-KkH5vOyTu0Xzl7CV3TT:-KkH5vOzwplJP7CQMBbA:b-z81ty7) ([raw file](https://github.com/google/mobly/blob/e02cf7183eec350ad78489cefb98bca2ebbe1610/mobly/test_runner.py#L211)):*
> ```Python
>     """
> 
>     class TestRunInfo(object):
> ```

Does this have to be inside `TestRunner`? If so, should it's name start with "_", otherwise could we move it to module level?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

adorokhine:

---

*[mobly/test_runner.py, line 292 at r4](https://reviewable.io:443/reviews/google/mobly/205#-KkE57nxNQxXMn012yFO:-KkILgduoCPJnr6LlzaQ:b-896fix) ([raw file](https://github.com/google/mobly/blob/b93f3b6258547f9cdbbd72ac3264b43e2e7ded88/mobly/test_runner.py#L292)):*

Traditionally we've been indenting to after ": " in docstrings, could we follow the same convention here?
Here and elsewhere.
</blockquote></details>

Done.

---

*[mobly/test_runner.py, line 211 at r5](https://reviewable.io:443/reviews/google/mobly/205#-KkH5vOyTu0Xzl7CV3TT:-KkIKAMQuGpTsr9NxCP9:b-s1iuo7) ([raw file](https://github.com/google/mobly/blob/e02cf7183eec350ad78489cefb98bca2ebbe1610/mobly/test_runner.py#L211)):*

Does this have to be inside `TestRunner`? If so, should it's name start with "_", otherwise could we move it to module level?
</blockquote></details>

I've put it the class on the principle of most limited possible scope. Added an underscore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

xpconanfan:

---

*[mobly/suite_runner.py, line 106 at r6](https://reviewable.io:443/reviews/google/mobly/205#-KkJ-VHddgCo-4KdUOLF:-KkJ-VHddgCo-4KdUOLG:b-oi8i2r) ([raw file](https://github.com/google/mobly/blob/9397eacde5fc6aa26cbd15eedbe270dfcc4dfcdc/mobly/suite_runner.py#L106)):*
> ```Python
> 
> 
> def _compute_selected_testcases(test_classes, selected_test_cases):
> ```

There seem to be a mix use of terms in code, var names, and docstrings: "method", "test method", "testcase", and "test_case", as this function signature and the doc string next line demonstrate.

Could we use consistent terms (maybe "test_case" for code/"test case" for text) in all places? (Since the test case header in logs has been "[Test Case]" since the beginning)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

adorokhine:

---

*[mobly/suite_runner.py, line 106 at r6](https://reviewable.io:443/reviews/google/mobly/205#-KkJ-VHddgCo-4KdUOLF:-KkJwmwlcHgNy_TEaec9:b-896fix) ([raw file](https://github.com/google/mobly/blob/9397eacde5fc6aa26cbd15eedbe270dfcc4dfcdc/mobly/suite_runner.py#L106)):*

There seem to be a mix use of terms in code, var names, and docstrings: "method", "test method", "testcase", and "test_case", as this function signature and the doc string next line demonstrate.

Could we use consistent terms (maybe "test_case" for code/"test case" for text) in all places? (Since the test case header in logs has been "[Test Case]" since the beginning)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[mobly/suite_runner.py, line 66 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkK1lC5b_v5TIW5qtmy:-KkK1lC5b_v5TIW5qtmz:b-b47ei6) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/suite_runner.py#L66)):*
> ```Python
>     parser.add_argument(
>         '--test',
>         '--test_case',
> ```

shouldn't this name change too if we don't want to use "test case"?

---

*[mobly/suite_runner.py, line 107 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkJzvwoXJEHC1HWvMIm:-KkJzvwoXJEHC1HWvMIn:b-hdzl8v) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/suite_runner.py#L107)):*
> ```Python
> 
> def _compute_selected_tests(test_classes, selected_tests):
>     """Computes test methods to run for each class from selector strings.
> ```

Should be "Computes the tests to run ..."? since this is referring to test cases in a human sense and the output, as is indicated by the name of this function.

---

*[mobly/suite_runner.py, line 109 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkJztm5mK1t3L6ZJ7wB:-KkJztm5mK1t3L6ZJ7wC:b-5l9116) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/suite_runner.py#L109)):*
> ```Python
>     """Computes test methods to run for each class from selector strings.
> 
>     This method transforms a list of selector strings (such as FooTest or
> ```

Should be "This function ..." ?

---

*[mobly/suite_runner.py, line 140 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkK-ESLwUyyrmHPPy6E:-KkJzVUyVlp2CI919z3u:b-dfsoia) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/suite_runner.py#L140)):*
> ```Python
>     # Dict from test class name to list of test methods to execute (or None for
>     # all methods).
>     name_to_methods = collections.OrderedDict()
> ```

The output dict holds the names of the selected tests instead of the actual test methods right?
So the doc and var names should reflect this?

"test": a test case in the human sense.
"test method": the actual method that gets called for a test
"test class": a class that groups test methods

So for the string name of these we should use "test name", "test_method_name", and "test_class_name" accordingly?

---

*[mobly/test_runner.py, line 246 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkK07FgyTbgoo9gsdO4:-KkK07FgyTbgoo9gsdO5:b-czu6yq) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/test_runner.py#L246)):*
> ```Python
>                     test class with.
>             test_class: class, test class to execute.
>             test_methods: Optional list of test method names to execute within
> ```

according to the docstring, this arg should be called "test_method_names"?
also first part should be "list of strings"?

similarly in `_run_test_class`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

adorokhine:
Review status: 0 of 11 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[mobly/suite_runner.py, line 107 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkJzvwoXJEHC1HWvMIm:-KkSX59hBsnXbGQ8fUkv:b-896fix) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/suite_runner.py#L107)):*

Should be "Computes the tests to run ..."? since this is referring to test cases in a human sense and the output, as is indicated by the name of this function.
</blockquote></details>

Done.

---

*[mobly/suite_runner.py, line 109 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkJztm5mK1t3L6ZJ7wB:-KkSX7AHBsYnvpnrh_Wo:b-896fix) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/suite_runner.py#L109)):*

Should be "This function ..." ?
</blockquote></details>

Done.

---

*[mobly/suite_runner.py, line 140 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkK-ESLwUyyrmHPPy6E:-KkSX9ZAhzgXKIr-EHqg:b-896fix) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/suite_runner.py#L140)):*

The output dict holds the names of the selected tests instead of the actual test methods right?
So the doc and var names should reflect this?

"test": a test case in the human sense.
"test method": the actual method that gets called for a test
"test class": a class that groups test methods

So for the string name of these we should use "test name", "test_method_name", and "test_class_name" accordingly?
</blockquote></details>

Done.

---

*[mobly/test_runner.py, line 246 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkK07FgyTbgoo9gsdO4:-KkSXLAWRV79oBRz8Q-R:b-896fix) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/test_runner.py#L246)):*

according to the docstring, this arg should be called "test_method_names"?
also first part should be "list of strings"?

similarly in `_run_test_class`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

adorokhine:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[mobly/suite_runner.py, line 66 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkK1lC5b_v5TIW5qtmy:-KkK4_sRBphxDUuFR5ek:b-hn0ru1) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/suite_runner.py#L66)):*

shouldn't this name change too if we don't want to use "test case"?
</blockquote></details>

The flag has indeed changed. The old name is retained only for backwards compatibility.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

xpconanfan:

---

*[mobly/suite_runner.py, line 109 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkJztm5mK1t3L6ZJ7wB:-KkSXq1Mxv6xqTuOfU56:bi58ddr) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/suite_runner.py#L109)):*

Done.
</blockquote></details>

Hmm, I still see "this method" in r8.

---

*[mobly/test_runner.py, line 246 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkK07FgyTbgoo9gsdO4:-KkSYPVDQON0q1Q-n0T3:b32zphz) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/test_runner.py#L246)):*

Done.
</blockquote></details>

I don't see "list of strings" specifying the type in docstring in r8.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

adorokhine:

---

*[mobly/suite_runner.py, line 109 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkJztm5mK1t3L6ZJ7wB:-KkSZGU1k_EV6TdgqO49:b-ewdkzi) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/suite_runner.py#L109)):*

Hmm, I still see "this method" in r8.
</blockquote></details>

Uggh, forgot to commit... must be friday

---

*[mobly/test_runner.py, line 246 at r7](https://reviewable.io:443/reviews/google/mobly/205#-KkK07FgyTbgoo9gsdO4:-KkSZgoElZBxigH6Hxxx:b-896fix) ([raw file](https://github.com/google/mobly/blob/bfb33bc5e22a0db50331de84b81e90aa6c57dfdc/mobly/test_runner.py#L246)):*

I don't see "list of strings" specifying the type in docstring in r8.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205)*

xpconanfan:
:lgtm:

Thanks a lot!!!

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/205#-:-KkY92cmXtE1adMRhKnc:b-as90x7)*

PR #207, opened 2017-05-12
Add a tutorial example for new generated tests.

xpconanfan:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/207)

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -294,3 +294,68 @@ menu.
 To learn more about Mobly Snippet Lib, including features like Espresso support
 and asynchronous calls, see the [snippet lib examples]
 (https://github.com/google/mobly-snippet-lib/tree/master/examples).
+
+
+# Example 6: Generated Tests
+
+A common use case in writing tests is to execute the same test logic multiple
+times, each time with a different set of parameters. Instead of duplicating the
+same test case with minor tweaks, you could use the **Generated tests** in
+Mobly.
+
+Mobly could generate test cases for you based on a list of parameters and a
+function that contains the test logic. Each generated test case is equivalent
+to an actual test case written in the class in terms of execution, procedure
+functions (setup/teardown/on_fail), and result collection. You could also
+select generated test cases via the `--test_case` cli arg as well.
+
+Here's an example of generated tests in action. We will reuse the "Example 1:
+Hello World!". Instead of making one toast of "Hello World", we will generate
+several test cases and toast a different message in each one of them.
+
+You could reuse the config file from Example 1.
+
+The test class would look like:
+
+ 
+**many_greetings_test.py**
+ 
+```python
+from mobly import base_test
+from mobly import test_runner
+from mobly.controllers import android_device
+
+
+class ManyGreetingsTest(base_test.BaseTestClass):
+
+    def setup_generated_tests(self):
+        messages = [('Hello', 'World'), ('Aloha', 'Obama'),
+                    ('konichiwa', 'Satoshi')]
+        self.generate_tests(
+            test_logic=self.make_toast_logic,
+            name_func=self.make_toast_name_function,
+            arg_sets=messages)
+
+    def setup_class(self):
+        # Registering android_device controller module declares the test's
+        # dependency on Android device hardware. By default, we expect at least
+        # one object is created from this.
+        self.ads = self.register_controller(android_device)
+        self.dut = self.ads[0]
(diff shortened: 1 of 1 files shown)


adorokhine:

---

*[docs/tutorial.md, line 321 at r1](https://reviewable.io:443/reviews/google/mobly/207#-KjyljbFlEW86VT9rGY8:-KjyljbGkN8mIx-kscqR:bywi2vl) ([raw file](https://github.com/google/mobly/blob/02b739ffb4f7c6047fb101ee07fa03d9486252ca/docs/tutorial.md#L321)):*
> ```Markdown
> 
>  
> **many_greetings_test.py**
> ```

Can you add comments to this explaining what the test is doing and what people should look at? It's hard to understand a big code dump like this. The only explanatory comment seems to be about min_num which is not really relevant here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/207)*

xpconanfan:

---

*[docs/tutorial.md, line 321 at r1](https://reviewable.io:443/reviews/google/mobly/207#-KjyljbFlEW86VT9rGY8:-KjywlVVkYt9o1JiTUP0:b-gv0nww) ([raw file](https://github.com/google/mobly/blob/02b739ffb4f7c6047fb101ee07fa03d9486252ca/docs/tutorial.md#L321)):*

Can you add comments to this explaining what the test is doing and what people should look at? It's hard to understand a big code dump like this. The only explanatory comment seems to be about min_num which is not really relevant here.
</blockquote></details>

The paragraphs before this code explains what the test does.
What else would you like to see?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/207)*

adorokhine:

---

*[docs/tutorial.md, line 321 at r1](https://reviewable.io:443/reviews/google/mobly/207#-KjyljbFlEW86VT9rGY8:-KjzeyMo4rGYV4UH7kS6:bem3nbq) ([raw file](https://github.com/google/mobly/blob/02b739ffb4f7c6047fb101ee07fa03d9486252ca/docs/tutorial.md#L321)):*

The paragraphs before this code explains what the test does.
What else would you like to see?
</blockquote></details>

I would like to see comments in the code to the effect of:
- setup_generated_tests is a method invoked by the framework before setup_class; its purpose is to invoke generate_tests
- generate_tests generates testcases for each argument that was passed in. It may only be called from setup_generated_tests.
- You can load controllers before calling self.generate_tests here if needed, but most tests shouldn't need to access controllers.
- The name func will be called with the same arguments as your test logic func. Its return value will be the name of the testcase.
- You can select a generated test for execution by passing the --test_case arg: example
- This test makes a toast on the device with one of the greetings from the messages list. Each greeting gets its own testcase.

I want to add comments that make it a user-friendly explanatory tutorial rather than a code dump.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/207)*

xpconanfan:

---

*[docs/tutorial.md, line 321 at r1](https://reviewable.io:443/reviews/google/mobly/207#-KjyljbFlEW86VT9rGY8:-Kk-9TIuA1I4o_z_6hVt:b-g13b4j) ([raw file](https://github.com/google/mobly/blob/02b739ffb4f7c6047fb101ee07fa03d9486252ca/docs/tutorial.md#L321)):*

I would like to see comments in the code to the effect of:
- setup_generated_tests is a method invoked by the framework before setup_class; its purpose is to invoke generate_tests
- generate_tests generates testcases for each argument that was passed in. It may only be called from setup_generated_tests.
- You can load controllers before calling self.generate_tests here if needed, but most tests shouldn't need to access controllers.
- The name func will be called with the same arguments as your test logic func. Its return value will be the name of the testcase.
- You can select a generated test for execution by passing the --test_case arg: example
- This test makes a toast on the device with one of the greetings from the messages list. Each greeting gets its own testcase.

I want to add comments that make it a user-friendly explanatory tutorial rather than a code dump.


</blockquote></details>

While it's good to have these info, I'd rather not repeat the same info already covered by the doc strings of `setup_generated_tests` and `generate_tests`. Could we link to the doc string or is there a better way?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/207)*

adorokhine:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/207#-:-KkIkXXcClc0yffHgKPz:bnfp4nl)*

PR #218, opened 2017-05-25
Fix a bug where internal controller registry is exposed.

### LINKED ISSUE #217: `register_controller` exposes Mobly's internal object registry.
`register_controller` returns a ref to the actual list Mobly uses to keep track of controller objects.
So if a test manipulates this list, it'll mess up the controller's teardown.
E.g. if the test removes controller objs from the list returned by `register_controller`, the removed controller objs won't be torn down.

xpconanfan:
Fixes #217

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/218)

### DIFF
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -400,11 +400,13 @@ class TestRunner(object):
             raise signals.ControllerError(
                 'Expected to get at least %d controller objects, got %d.' %
                 (min_number, actual_number))
-        self._controller_registry[module_ref_name] = objects
+        # Save a shallow copy of the list for internal usage, so tests can't
+        # affect internal registry by manipulating the object list.
+        self._controller_registry[module_ref_name] = copy.copy(objects)
         # Collect controller information and write to test result.
         # Implementation of 'get_info' is optional for a controller module.
         if hasattr(module, 'get_info'):
-            controller_info = module.get_info(objects)
+            controller_info = module.get_info(copy.copy(objects))
             logging.debug('Controller %s: %s', module_config_name,
                           controller_info)
             self.results.add_controller_info(module_config_name,
@@ -417,9 +419,7 @@ class TestRunner(object):
                       len(objects), module_config_name)
         destroy_func = module.destroy
         self._controller_destructors[module_ref_name] = destroy_func
-        # Return a shallow copy of the list so tests can't affect teardown by
-        # manipulating the returned object list.
-        return copy.copy(objects)
+        return objects
 
     def _unregister_controllers(self):
         """Destroy controller objects and clear internal registry.
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -417,7 +417,9 @@ class TestRunner(object):
                       len(objects), module_config_name)
         destroy_func = module.destroy
         self._controller_destructors[module_ref_name] = destroy_func
-        return objects
+        # Return a shallow copy of the list so tests can't affect teardown by
+        # manipulating the returned object list.
+        return copy.copy(objects)
 
     def _unregister_controllers(self):
         """Destroy controller objects and clear internal registry.
--- a/tests/mobly/test_runner_test.py
+++ b/tests/mobly/test_runner_test.py
@@ -119,7 +119,8 @@ class TestRunnerTest(unittest.TestCase):
         tr = test_runner.TestRunner(self.log_dir, self.test_bed_name)
         magic_devices = tr._register_controller(mock_test_config,
                                                 mock_controller)
-        magic_devices.pop(0)
+        magic1 = magic_devices.pop(0)
+        self.assertIs(magic1, tr._controller_registry['mock_controller'][0])
         self.assertEqual(len(tr._controller_registry['mock_controller']), 2)
 
     def test_register_controller_less_than_min_number(self):
--- a/tests/mobly/test_runner_test.py
+++ b/tests/mobly/test_runner_test.py
@@ -110,6 +110,18 @@ class TestRunnerTest(unittest.TestCase):
         self.assertEqual(magic_devices[0].magic, 'magic1')
         self.assertEqual(magic_devices[1].magic, 'magic2')
 
+    def test_register_controller_change_return_value(self):
+        mock_test_config = self.base_mock_test_config.copy()
+        mock_ctrlr_config_name = mock_controller.MOBLY_CONTROLLER_CONFIG_NAME
+        mock_test_config.controller_configs = {
+            mock_ctrlr_config_name: ['magic1', 'magic2']
+        }
+        tr = test_runner.TestRunner(self.log_dir, self.test_bed_name)
+        magic_devices = tr._register_controller(mock_test_config,
+                                                mock_controller)
+        magic_devices.pop(0)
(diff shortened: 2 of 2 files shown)


dthkao:

---

*[tests/mobly/test_runner_test.py, line 113 at r1](https://reviewable.io:443/reviews/google/mobly/218#-Kl-HMOrqDRNGSwuZp9I:-Kl-HMOrqDRNGSwuZp9J:ba5tea4) ([raw file](https://github.com/google/mobly/blob/3b0f0b6eaaa0cf1229c2513820ef4d4fa5a625a5/tests/mobly/test_runner_test.py#L113)):*
> ```Python
>         self.assertEqual(magic_devices[1].magic, 'magic2')
> 
>     def test_register_controller_change_return_value(self):
> ```

Just so I'm clear on intent, this is to allow users to still modify the original individual controller objects *and* the list they get returned, without affecting the list Mobly tracks.

Can we add a unit test to make sure the individual controller is still modifiable (just so no one later changes this to deep copy)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/218)*

xpconanfan:

---

*[tests/mobly/test_runner_test.py, line 113 at r1](https://reviewable.io:443/reviews/google/mobly/218#-Kl-HMOrqDRNGSwuZp9I:-Kl-LDH87wFwXxrNval4:bogylse) ([raw file](https://github.com/google/mobly/blob/3b0f0b6eaaa0cf1229c2513820ef4d4fa5a625a5/tests/mobly/test_runner_test.py#L113)):*

Just so I'm clear on intent, this is to allow users to still modify the original individual controller objects *and* the list they get returned, without affecting the list Mobly tracks.

Can we add a unit test to make sure the individual controller is still modifiable (just so no one later changes this to deep copy)?
</blockquote></details>

Yes.
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/218)*

adorokhine:

---

*[mobly/test_runner.py, line 407 at r2](https://reviewable.io:443/reviews/google/mobly/218#-Kl0QtX4yKncCzbqPsFy:-Kl0QtX4yKncCzbqPsFz:b-emumpy) ([raw file](https://github.com/google/mobly/blob/289165976be976598b888e4e06394c21de6a68a6/mobly/test_runner.py#L407)):*
> ```Python
>         # Implementation of 'get_info' is optional for a controller module.
>         if hasattr(module, 'get_info'):
>             controller_info = module.get_info(objects)
> ```

There's other calls into user code, like get_info. If we're trying to be safe we should probably just copy the one that goes into self._controller_registry so we can clean up properly, and let the user code real with 'their' copy however they want.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/218)*

xpconanfan:

---

*[mobly/test_runner.py, line 407 at r2](https://reviewable.io:443/reviews/google/mobly/218#-Kl0QtX4yKncCzbqPsFy:-Kl0TlknABcasio0C-f6:b352r32) ([raw file](https://github.com/google/mobly/blob/289165976be976598b888e4e06394c21de6a68a6/mobly/test_runner.py#L407)):*

There's other calls into user code, like get_info. If we're trying to be safe we should probably just copy the one that goes into self._controller_registry so we can clean up properly, and let the user code real with 'their' copy however they want.
</blockquote></details>

Hmm, shouldn't we collect info regardless of what user does to the obj list?
`get_info` is also an internal mobly operation, so I don't think what user does to the list in their test should affect how mobly internal operation works.
I'd think it's better to return the copy?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/218)*

adorokhine:

---

*[mobly/test_runner.py, line 407 at r2](https://reviewable.io:443/reviews/google/mobly/218#-Kl0QtX4yKncCzbqPsFy:-Kl0UND7tfiB4tEdcatV:b-3k0ii0) ([raw file](https://github.com/google/mobly/blob/289165976be976598b888e4e06394c21de6a68a6/mobly/test_runner.py#L407)):*

Hmm, shouldn't we collect info regardless of what user does to the obj list?
`get_info` is also an internal mobly operation, so I don't think what user does to the list in their test should affect how mobly internal operation works.
I'd think it's better to return the copy?
</blockquote></details>

I thought `get_info` was implemented by the user? Not sure what you mean when you say it's an internal mobly operation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/218)*

xpconanfan:

---

*[mobly/test_runner.py, line 407 at r2](https://reviewable.io:443/reviews/google/mobly/218#-Kl0QtX4yKncCzbqPsFy:-Kl0Wdnh7pZ_9gz6k4m5:bymwqib) ([raw file](https://github.com/google/mobly/blob/289165976be976598b888e4e06394c21de6a68a6/mobly/test_runner.py#L407)):*

I thought `get_info` was implemented by the user? Not sure what you mean when you say it's an internal mobly operation.
</blockquote></details>

Calling `get_info` is an internal operation like calling `destroy`.

Which list we use decides which objs are used to call these functions.

So it would be inconsistent to call `destroy` with the copy list and call `get_info` with the original list.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/218)*

adorokhine:

---

*[mobly/test_runner.py, line 407 at r2](https://reviewable.io:443/reviews/google/mobly/218#-Kl0QtX4yKncCzbqPsFy:-Kl0Z-AaVqdWjqcBAXkP:bq0kvjk) ([raw file](https://github.com/google/mobly/blob/289165976be976598b888e4e06394c21de6a68a6/mobly/test_runner.py#L407)):*

Calling `get_info` is an internal operation like calling `destroy`.

Which list we use decides which objs are used to call these functions.

So it would be inconsistent to call `destroy` with the copy list and call `get_info` with the original list.
</blockquote></details>

Sure, that sounds good. All I'm saying is that, as of the latest patchset, the following implementation of `get_info` would still trigger the bug you're trying to fix:

```
def get_info(self, objects):
  objects.clear()
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/218)*

xpconanfan:

---

*[mobly/test_runner.py, line 407 at r2](https://reviewable.io:443/reviews/google/mobly/218#-Kl0QtX4yKncCzbqPsFy:-Kl0aV1XylDYfV2KCDXk:b-v802qq) ([raw file](https://github.com/google/mobly/blob/289165976be976598b888e4e06394c21de6a68a6/mobly/test_runner.py#L407)):*

Sure, that sounds good. All I'm saying is that, as of the latest patchset, the following implementation of `get_info` would still trigger the bug you're trying to fix:

```
def get_info(self, objects):
  objects.clear()
```
</blockquote></details>

Ah good point.
Fixed.

Perhaps we should move `get_info` to unregister controller phase too (later)...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/218)*

adorokhine:
:lgtm:

---

Reviewed 1 of 1 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/218#-:-Kl1Ac3eXFspVt7kIONb:bnfp4nl)*

PR #253, opened 2017-07-12
Drop support for deprecated v0 snippets.

adorokhine:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/253)

### DIFF
--- a/mobly/controllers/android_device_lib/snippet_client.py
+++ b/mobly/controllers/android_device_lib/snippet_client.py
@@ -23,8 +23,8 @@ from mobly.controllers.android_device_lib import jsonrpc_client_base
 _INSTRUMENTATION_RUNNER_PACKAGE = (
     'com.google.android.mobly.snippet.SnippetRunner')
 
-_LAUNCH_CMD = (
-    'am instrument -w -e action start %s/' + _INSTRUMENTATION_RUNNER_PACKAGE)
+_LAUNCH_CMD = ('%s am instrument -w -e action start %s/' +
+               _INSTRUMENTATION_RUNNER_PACKAGE)
 
 _STOP_CMD = (
     'am instrument -w -e action stop %s/' + _INSTRUMENTATION_RUNNER_PACKAGE)
--- a/mobly/controllers/android_device_lib/snippet_client.py
+++ b/mobly/controllers/android_device_lib/snippet_client.py
@@ -23,12 +23,8 @@ from mobly.controllers.android_device_lib import jsonrpc_client_base
 _INSTRUMENTATION_RUNNER_PACKAGE = (
     'com.google.android.mobly.snippet.SnippetRunner')
 
-# TODO(adorokhine): delete this in Mobly 1.6 when snippet v0 support is removed.
-_LAUNCH_CMD_V0 = ('%s am instrument -w -e action start -e port %s %s/' +
-                  _INSTRUMENTATION_RUNNER_PACKAGE)
-
-_LAUNCH_CMD_V1 = ('%s am instrument -w -e action start %s/' +
-                  _INSTRUMENTATION_RUNNER_PACKAGE)
+_LAUNCH_CMD = (
+    'am instrument -w -e action start %s/' + _INSTRUMENTATION_RUNNER_PACKAGE)
 
 _STOP_CMD = (
     'am instrument -w -e action stop %s/' + _INSTRUMENTATION_RUNNER_PACKAGE)
@@ -45,10 +41,6 @@ _SETSID_COMMAND = 'setsid'
 
 _NOHUP_COMMAND = 'nohup'
 
-# Maximum time to wait for a v0 snippet to start on the device (10 minutes).
-# TODO(adorokhine): delete this in Mobly 1.6 when snippet v0 support is removed.
-_APP_START_WAIT_TIME_V0 = 10 * 60
-
 
 class Error(Exception):
     pass
@@ -62,9 +54,6 @@ class SnippetClient(jsonrpc_client_base.JsonRpcClientBase):
     """A client for interacting with snippet APKs using Mobly Snippet Lib.
 
     See superclass documentation for a list of public attributes.
-
-    It currently supports both v0 and v1 snippet launch protocols, although
-    support for v0 will be removed in a future version.
 
     For a description of the launch protocols, see the documentation in
     mobly-snippet-lib, SnippetRunner.java.
@@ -83,60 +72,41 @@ class SnippetClient(jsonrpc_client_base.JsonRpcClientBase):
         self.package = package
         self._adb = adb_proxy
         self._proc = None
-        self._launch_version = 'v1'
 
     def start_app_and_connect(self):
         """Overrides superclass. Launches a snippet app and connects to it."""
         self._check_app_installed()
 
         persists_shell_cmd = self._get_persist_command()
-        # Try launching the app with the v1 protocol. If that fails, fall back
-        # to v0 for compatibility. Use info here so people know exactly what's
-        # happening here, which is helpful since they need to create their own
-        # instrumentations and manifest.
+        # Use info here so people can follow along with the snippet startup
+        # process. Starting snippets can be slow, especially if there are
+        # multiple, and this avoids the perception that the framework is hanging
+        # for a long time doing nothing.
         self.log.info('Launching snippet apk %s with protocol v1',
                       self.package)
-        cmd = _LAUNCH_CMD_V1 % (persists_shell_cmd, self.package)
+        cmd = _LAUNCH_CMD % (persists_shell_cmd, self.package)
         start_time = time.time()
         self._proc = self._do_start_app(cmd)
 
-        # "Instrumentation crashed" could be due to several reasons, eg
-        # exception thrown during startup or just a launch protocol 0 snippet
-        # dying because it needs the port flag. Sadly we have no way to tell so
-        # just warn and retry as v0.
-        # TODO(adorokhine): delete this in Mobly 1.6 when snippet v0 support is
-        # removed.
+        # Check protocol version and get the device port
         line = self._read_protocol_line()
+        match = re.match('^SNIPPET START, PROTOCOL ([0-9]+) ([0-9]+)$', line)
+        if not match or match.group(1) != '1':
+            raise ProtocolVersionError(line)
+
+        line = self._read_protocol_line()
+        match = re.match('^SNIPPET SERVING, PORT ([0-9]+)$', line)
+        if not match:
+            raise ProtocolVersionError(line)
+        self.device_port = int(match.group(1))
+
         # Forward the device port to a new host port, and connect to that port
         self.host_port = utils.get_available_host_port()
-        if line in ('INSTRUMENTATION_RESULT: shortMsg=Process crashed.',
-                    'INSTRUMENTATION_RESULT: shortMsg='
-                    'java.lang.IllegalArgumentException'):
-            self.log.warning('Snippet %s crashed on startup. This might be an '
-                             'actual error or a snippet using deprecated v0 '
-                             'start protocol. Retrying as a v0 snippet.',
-                             self.package)
-            # Reuse the host port as the device port in v0 snippet. This isn't
-            # safe in general, but the protocol is deprecated.
-            self.device_port = self.host_port
-            cmd = _LAUNCH_CMD_V0 % (persists_shell_cmd, self.device_port,
-                                    self.package)
-            self._proc = self._do_start_app(cmd)
-            self._connect_to_v0()
-            self._launch_version = 'v0'
-        else:
-            # Check protocol version and get the device port
-            match = re.match('^SNIPPET START, PROTOCOL ([0-9]+) ([0-9]+)$',
-                             line)
-            if not match or match.group(1) != '1':
-                raise ProtocolVersionError(line)
+        self._adb.forward(
+            ['tcp:%d' % self.host_port, 'tcp:%d' % self.device_port])
+        self.connect()
 
-            line = self._read_protocol_line()
-            match = re.match('^SNIPPET SERVING, PORT ([0-9]+)$', line)
-            if not match:
-                raise ProtocolVersionError(line)
-            self.device_port = int(match.group(1))
-            self._connect_to_v1()
+        # Yaaay! We're done!
         self.log.debug('Snippet %s started after %.1fs on host port %s',
                        self.package, time.time() - start_time, self.host_port)
 
@@ -157,11 +127,10 @@ class SnippetClient(jsonrpc_client_base.JsonRpcClientBase):
             AppRestoreConnectionError: When the app was not able to be started.
         """
         self.host_port = port or utils.get_available_host_port()
+        self._adb.forward(
+            ['tcp:%d' % self.host_port, 'tcp:%d' % self.device_port])
         try:
-            if self._launch_version == 'v0':
-                self._connect_to_v0()
-            else:
-                self._connect_to_v1()
+            self.connect()
         except:
             # Failed to connect to app, something went wrong.
             raise jsonrpc_client_base.AppRestoreConnectionError(
@@ -246,32 +215,6 @@ class SnippetClient(jsonrpc_client_base.JsonRpcClientBase):
             adb_cmd += ['-s', self._adb.serial]
         adb_cmd += ['shell', launch_cmd]
         return utils.start_standing_subprocess(adb_cmd, shell=False)
-
-    # TODO(adorokhine): delete this in Mobly 1.6 when snippet v0 support is
-    # removed.
-    def _connect_to_v0(self):
-        self._adb.forward(
-            ['tcp:%d' % self.host_port, 'tcp:%d' % self.device_port])
-        start_time = time.time()
-        expiration_time = start_time + _APP_START_WAIT_TIME_V0
-        while time.time() < expiration_time:
-            self.log.debug('Attempting to start %s.', self.package)
-            try:
-                self.connect()
-                return
-            except:
-                self.log.debug(
-                    'v0 snippet %s is not yet running, retrying',
-                    self.package,
-                    exc_info=True)
-            time.sleep(1)
-        raise jsonrpc_client_base.AppStartError(
-            '%s failed to start on %s.' % (self.package, self._adb.serial))
-
-    def _connect_to_v1(self):
-        self._adb.forward(
-            ['tcp:%d' % self.host_port, 'tcp:%d' % self.device_port])
-        self.connect()
 
     def _read_protocol_line(self):
         """Reads the next line of instrumentation output relevant to snippets.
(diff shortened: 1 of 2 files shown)


xpconanfan:

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 26 at r1](https://reviewable.io:443/reviews/google/mobly/253#-KooK3fAhyZ0nVykDNBe:-KooK3fBzMKxY5eEnoox:b-x65cjm) ([raw file](https://github.com/google/mobly/blob/c782f1b258c13e5b2becd0b49cb4afe3f62eaab8/mobly/controllers/android_device_lib/snippet_client.py#L26)):*
> ```Python
>     'com.google.android.mobly.snippet.SnippetRunner')
> 
> _LAUNCH_CMD_V1 = (
> ```

should we rename 'v1' from names as well?
it is a tiny bit confusing as there is only one version left now.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 68 at r1](https://reviewable.io:443/reviews/google/mobly/253#-KooKW4ff_7SNOxA20DR:-KooKW4ff_7SNOxA20DS:by46tt3) ([raw file](https://github.com/google/mobly/blob/c782f1b258c13e5b2becd0b49cb4afe3f62eaab8/mobly/controllers/android_device_lib/snippet_client.py#L68)):*
> ```Python
>         self._check_app_installed()
> 
>         self.log.info('Launching snippet apk %s with protocol v1',
> ```

the original comment explained why it's `info` here.
if that's being removed, does the justification not stand any more?
could we change this to `debug`?...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

adorokhine:

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 26 at r1](https://reviewable.io:443/reviews/google/mobly/253#-KooK3fAhyZ0nVykDNBe:-KooLic488clt3cPpZk3:b-896fix) ([raw file](https://github.com/google/mobly/blob/c782f1b258c13e5b2becd0b49cb4afe3f62eaab8/mobly/controllers/android_device_lib/snippet_client.py#L26)):*

should we rename 'v1' from names as well?
it is a tiny bit confusing as there is only one version left now.
</blockquote></details>

Done.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 68 at r1](https://reviewable.io:443/reviews/google/mobly/253#-KooKW4ff_7SNOxA20DR:-KooMGPowjBhdXdTw_E9:b-896fix) ([raw file](https://github.com/google/mobly/blob/c782f1b258c13e5b2becd0b49cb4afe3f62eaab8/mobly/controllers/android_device_lib/snippet_client.py#L68)):*

the original comment explained why it's `info` here.
if that's being removed, does the justification not stand any more?
could we change this to `debug`?...
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

xpconanfan:

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 72 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooN8fNtwiDCrCw5PGW:-KooN8fNtwiDCrCw5PGX:bskzn4h) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/mobly/controllers/android_device_lib/snippet_client.py#L72)):*
> ```Python
>         # multiple, and this avoids the perception that the framework is hanging
>         # for a long time doing nothing.
>         self.log.info('Launching snippet apk %s with protocol v1',
> ```

should 'v1' be removed here too?

---

*[tests/mobly/controllers/android_device_lib/snippet_client_test.py, line 220 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooMypnU5njIrAXiKnE:-KooMypoI_KYuAS-N1xw:bskzn4h) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/tests/mobly/controllers/android_device_lib/snippet_client_test.py#L220)):*
> ```Python
>     @mock.patch('mobly.controllers.android_device_lib.snippet_client.'
>                 'utils.get_available_host_port')
>     def test_snippet_start_app_and_connect_v1_header_junk(
> ```

should 'v1' be removed here too?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

adorokhine:

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 72 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooN8fNtwiDCrCw5PGW:-KooR2yamZpRaZYLP54R:b8haa09) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/mobly/controllers/android_device_lib/snippet_client.py#L72)):*

should 'v1' be removed here too?
</blockquote></details>

Since this particular log message is user-facing, I'd prefer to keep it. User actions can affect this, and it'll help us a lot going forward if we ever add a v2 protocol.

---

*[tests/mobly/controllers/android_device_lib/snippet_client_test.py, line 220 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooMypnU5njIrAXiKnE:-KooRJ344jh4L4knVVco:bozwzx2) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/tests/mobly/controllers/android_device_lib/snippet_client_test.py#L220)):*

should 'v1' be removed here too?
</blockquote></details>

This test is specifically seeding "PROTOCOL 1 0" into the mock and making sure junk before this line are ignored. Since the protocol explicitly appears here slight preference for keeping it, but I can remove it if you want.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

xpconanfan:

---

*[tests/mobly/controllers/android_device_lib/snippet_client_test.py, line 220 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooMypnU5njIrAXiKnE:-KooRc-IRi6mRe7dvsXi:b-z67lwq) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/tests/mobly/controllers/android_device_lib/snippet_client_test.py#L220)):*

This test is specifically seeding "PROTOCOL 1 0" into the mock and making sure junk before this line are ignored. Since the protocol explicitly appears here slight preference for keeping it, but I can remove it if you want.
</blockquote></details>

what is 'v1' relative to though...
it looks like the new mechanism has protocol version taken care of so we probably wouldn't need to hard code version name to function/var names going forward, unless there's a huge change somehow, which is unlikely?

Same for the other comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

adorokhine:

---

*[tests/mobly/controllers/android_device_lib/snippet_client_test.py, line 220 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooMypnU5njIrAXiKnE:-KooSWwP0YiUNuvX_dS9:b-896fix) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/tests/mobly/controllers/android_device_lib/snippet_client_test.py#L220)):*

what is 'v1' relative to though...
it looks like the new mechanism has protocol version taken care of so we probably wouldn't need to hard code version name to function/var names going forward, unless there's a huge change somehow, which is unlikely?

Same for the other comment.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

xpconanfan:

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 72 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooN8fNtwiDCrCw5PGW:-KooT0BLWUOzmoxycaIZ:bvhkfu5) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/mobly/controllers/android_device_lib/snippet_client.py#L72)):*

Since this particular log message is user-facing, I'd prefer to keep it. User actions can affect this, and it'll help us a lot going forward if we ever add a v2 protocol.
</blockquote></details>

I see, does this mean we should log this based on the number given back by the message 'SNIPPET START, PROTOCOL x x' ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

adorokhine:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 72 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooN8fNtwiDCrCw5PGW:-KoyWX7j97tjjmoFgUPC:b-2qitkz) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/mobly/controllers/android_device_lib/snippet_client.py#L72)):*

I see, does this mean we should log this based on the number given back by the message 'SNIPPET START, PROTOCOL x x' ?
</blockquote></details>

Actually that wouldn't be accurate. The protocol is a negotiation between the server and client, but ultimately the client decides the protocol to use to start the app. The log message being discussed here is logging the client's final decision, not what the server advertised.

For example suppose we add protocol 1.1 which adds a 'PING' message where the server replies 'PONG' and use this for a heartbeat. This is backwards compatible. The server would advertise 1.1, but until we modify the client code, we'd still be using 1.0 to bring it up. So it would be incorrect to say we're bringing up the snippet using protocol 1.1 until we actually adjust the client code to make use of the new protocol features. 1.0 would be the right thing to log here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

xpconanfan:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 72 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooN8fNtwiDCrCw5PGW:-KoyX2fXflGXzWbg68be:br797n8) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/mobly/controllers/android_device_lib/snippet_client.py#L72)):*

Actually that wouldn't be accurate. The protocol is a negotiation between the server and client, but ultimately the client decides the protocol to use to start the app. The log message being discussed here is logging the client's final decision, not what the server advertised.

For example suppose we add protocol 1.1 which adds a 'PING' message where the server replies 'PONG' and use this for a heartbeat. This is backwards compatible. The server would advertise 1.1, but until we modify the client code, we'd still be using 1.0 to bring it up. So it would be incorrect to say we're bringing up the snippet using protocol 1.1 until we actually adjust the client code to make use of the new protocol features. 1.0 would be the right thing to log here.
</blockquote></details>

ok, but could we make the format match the server side?
let's show something similar to "SNIPPET START, PROTOCOL 1 0"?
Like " Launching ... with client-side protocol 1.0"?

Also, if the client here starts with 1.0, do we expect the server side to respond with "SNIPPET START, PROTOCOL 1 0", or "SNIPPET START, PROTOCOL 1 1" ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

adorokhine:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 72 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooN8fNtwiDCrCw5PGW:-KoyYD0KRj2LMnfIQ2O1:bp9tnti) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/mobly/controllers/android_device_lib/snippet_client.py#L72)):*
> Like "Launching ... with client-side protocol 1.0"?

Sure, sounds good. Let me send a new patch.

> Also, if the client here starts with 1.0, do we expect the server side to respond with "SNIPPET START, PROTOCOL 1 0", or "SNIPPET START, PROTOCOL 1 1" ?

Actually you have the order backwards; the server prints its protocol version first thing, before it knows anything about the client. I haven't been able to come up with a way to get data about the snippet until it's actually launched. So for now the 'am instrument' call is step 1 before we even know what the snippet is capable of, and the protocol version is the first line printed in response to that. After this handshake, the client can start making decisions about how to proceed.

So in answer to your question, the client blindly does an 'am instrument' call, and reads back `SNIPPET START, PROTOCOL VERSION 1 0` or whatever else the snippet was compiled with, and then decides what to do.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

xpconanfan:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 72 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooN8fNtwiDCrCw5PGW:-KoyZWyljQT3hq5EQB1C:bvcz3w8) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/mobly/controllers/android_device_lib/snippet_client.py#L72)):*

> Like "Launching ... with client-side protocol 1.0"?

Sure, sounds good. Let me send a new patch.

> Also, if the client here starts with 1.0, do we expect the server side to respond with "SNIPPET START, PROTOCOL 1 0", or "SNIPPET START, PROTOCOL 1 1" ?

Actually you have the order backwards; the server prints its protocol version first thing, before it knows anything about the client. I haven't been able to come up with a way to get data about the snippet until it's actually launched. So for now the 'am instrument' call is step 1 before we even know what the snippet is capable of, and the protocol version is the first line printed in response to that. After this handshake, the client can start making decisions about how to proceed.

So in answer to your question, the client blindly does an 'am instrument' call, and reads back `SNIPPET START, PROTOCOL VERSION 1 0` or whatever else the snippet was compiled with, and then decides what to do.
</blockquote></details>

Hmm, i'm confused.

It sounds like the client decides what protocol to use based on what the server prints in "PROTOCOL VERSION x x"?
If so, shouldn't this log line always be consistent with what the server prints?
And what does the client do if the server prints a number unknown to the client?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

adorokhine:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 72 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooN8fNtwiDCrCw5PGW:-KoyZxHB5O54SuQT4NB9:bkc03ik) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/mobly/controllers/android_device_lib/snippet_client.py#L72)):*
> It sounds like the client decides what protocol to use based on what the server prints in "PROTOCOL VERSION x x"?

That's right

> If so, shouldn't this log line always be consistent with what the server prints?

Nope, because if the server prints 1.1 an old client might still choose to use 1.0, which will be fine due to backwards compatibility requirements.

> And what does the client do if the server prints a number unknown to the client?

Unknown major version number: crash with ProtocolVersionError.
Unknown minor version number: ignored because backwards compatibility is guaranteed between minor versions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

xpconanfan:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 72 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooN8fNtwiDCrCw5PGW:-Koy_wYd0jBgdHsBOEGB:b-x0aukc) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/mobly/controllers/android_device_lib/snippet_client.py#L72)):*

> It sounds like the client decides what protocol to use based on what the server prints in "PROTOCOL VERSION x x"?

That's right

> If so, shouldn't this log line always be consistent with what the server prints?

Nope, because if the server prints 1.1 an old client might still choose to use 1.0, which will be fine due to backwards compatibility requirements.

> And what does the client do if the server prints a number unknown to the client?

Unknown major version number: crash with ProtocolVersionError.
Unknown minor version number: ignored because backwards compatibility is guaranteed between minor versions.
</blockquote></details>

Oh, I wasn't aware of this requirement.
Where is this requirement documented?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

adorokhine:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 72 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooN8fNtwiDCrCw5PGW:-Koya_-wi07Zox81EAzp:b84x397) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/mobly/controllers/android_device_lib/snippet_client.py#L72)):*

Oh, I wasn't aware of this requirement.
Where is this requirement documented?
</blockquote></details>

In the javadoc of SnippetRunner, where the protocol versions are defined:
https://github.com/google/mobly-snippet-lib/blob/110f5c76b325b6ce71cc606966f6fa6a3583b140/third_party/sl4a/src/main/java/com/google/android/mobly/snippet/SnippetRunner.java#L72

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

xpconanfan:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 72 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooN8fNtwiDCrCw5PGW:-Koycx3jhSRdSlmTTEw3:bcsalol) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/mobly/controllers/android_device_lib/snippet_client.py#L72)):*

In the javadoc of SnippetRunner, where the protocol versions are defined:
https://github.com/google/mobly-snippet-lib/blob/110f5c76b325b6ce71cc606966f6fa6a3583b140/third_party/sl4a/src/main/java/com/google/android/mobly/snippet/SnippetRunner.java#L72
</blockquote></details>

Ah cool, should we mention that in this file too?
Easier for people who make changes in this file to find.
We change this more often than the server side.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

xpconanfan:
Ping

xpconanfan:
Ping +___+

adorokhine:

---

*[mobly/controllers/android_device_lib/snippet_client.py, line 72 at r2](https://reviewable.io:443/reviews/google/mobly/253#-KooN8fNtwiDCrCw5PGW:-KqjWCzd-lrA0Te_Ly7C:b-896fix) ([raw file](https://github.com/google/mobly/blob/364e275d5c50be9e999f72548fc63afcee1e026f/mobly/controllers/android_device_lib/snippet_client.py#L72)):*

Ah cool, should we mention that in this file too?
Easier for people who make changes in this file to find.
We change this more often than the server side.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253)*

xpconanfan:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/253#-:-KqjY0fVQ6o3ZMvZMV4Y:bnfp4nl)*

PR #269, opened 2017-07-24
Fix a bug where timestamps are set on tests that didn't even start.

### LINKED ISSUE #268: When failure occurs in setup class "begin_time" for skipped tests is null
End time is not null.

Can we either make them equal to each other (i.e. skipped test takes 0 time) or both null?

xpconanfan:
Also add unit test conditions for this case.
Fixes #268

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/269)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -503,9 +503,6 @@ class BaseTestClass(object):
             if not self.results.is_test_executed(test_name):
                 test_record = records.TestResultRecord(test_name, self.TAG)
                 test_record.test_skip(exception)
-                # Set the begin and end time to None as the test never started.
-                test_record.begin_time = None
-                test_record.end_time = None
                 self.results.add_record(test_record)
 
     def run(self, test_names=None):
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -503,6 +503,9 @@ class BaseTestClass(object):
             if not self.results.is_test_executed(test_name):
                 test_record = records.TestResultRecord(test_name, self.TAG)
                 test_record.test_skip(exception)
+                # Set the begin and end time to None as the test never started.
+                test_record.begin_time = None
+                test_record.end_time = None
                 self.results.add_record(test_record)
 
     def run(self, test_names=None):
--- a/mobly/records.py
+++ b/mobly/records.py
@@ -89,7 +89,8 @@ class TestResultRecord(object):
                 be any exception instance or of any subclass of
                 mobly.signals.TestSignal.
         """
-        self.end_time = utils.get_current_epoch_time()
+        if self.begin_time is not None:
+            self.end_time = utils.get_current_epoch_time()
         self.result = result
         if self.extra_errors:
             self.result = TestResultEnums.TEST_RESULT_ERROR
--- a/tests/mobly/base_test_test.py
+++ b/tests/mobly/base_test_test.py
@@ -871,6 +871,8 @@ class BaseTestTest(unittest.TestCase):
         bt_cls = MockBaseTest(self.mock_test_cls_configs)
         bt_cls.run(test_names=["test_func"])
         actual_record = bt_cls.results.skipped[0]
+        self.assertIsNotNone(actual_record.begin_time)
+        self.assertIsNotNone(actual_record.end_time)
         self.assertEqual(actual_record.test_name, "test_func")
         self.assertEqual(actual_record.details, MSG_EXPECTED_EXCEPTION)
         self.assertEqual(actual_record.extras, MOCK_EXTRA)
@@ -886,6 +888,8 @@ class BaseTestTest(unittest.TestCase):
         bt_cls = MockBaseTest(self.mock_test_cls_configs)
         bt_cls.run(test_names=["test_func"])
         actual_record = bt_cls.results.skipped[0]
+        self.assertIsNotNone(actual_record.begin_time)
+        self.assertIsNotNone(actual_record.end_time)
         self.assertEqual(actual_record.test_name, "test_func")
         self.assertEqual(actual_record.details, MSG_EXPECTED_EXCEPTION)
         self.assertEqual(actual_record.extras, MOCK_EXTRA)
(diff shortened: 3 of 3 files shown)


dthkao:

---

*[mobly/base_test.py, line 506 at r1](https://reviewable.io:443/reviews/google/mobly/269#-KpuoMgbdJ67hzZXQmjI:-KpuoMgbdJ67hzZXQmjJ:b-bse92z) ([raw file](https://github.com/google/mobly/blob/499531fe09c3b66b7b639f3a8d9ff5c1017c6083/mobly/base_test.py#L506)):*
> ```Python
>                 test_record = records.TestResultRecord(test_name, self.TAG)
>                 test_record.test_skip(exception)
>                 # Set the begin and end time to None as the test never started.
> ```

What happens to skips outside of an abort_class?

Why not fix in either TestResultRecord.test_skip() or TestResultRecord._end_test()?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/269)*

xpconanfan:

---

*[mobly/base_test.py, line 506 at r1](https://reviewable.io:443/reviews/google/mobly/269#-KpuoMgbdJ67hzZXQmjI:-Kpuq9cA9XCP9bS_hW7z:beu1ilq) ([raw file](https://github.com/google/mobly/blob/499531fe09c3b66b7b639f3a8d9ff5c1017c6083/mobly/base_test.py#L506)):*

What happens to skips outside of an abort_class?

Why not fix in either TestResultRecord.test_skip() or TestResultRecord._end_test()?


</blockquote></details>

bc they are different behaviors.
`test_skip` is used for in-test skipping, in which case both timestamps are not None

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/269)*

dthkao:

---

*[mobly/base_test.py, line 506 at r1](https://reviewable.io:443/reviews/google/mobly/269#-KpuoMgbdJ67hzZXQmjI:-KpvAxuvuSpjyIcjIO4b:bn6ug98) ([raw file](https://github.com/google/mobly/blob/499531fe09c3b66b7b639f3a8d9ff5c1017c6083/mobly/base_test.py#L506)):*

bc they are different behaviors.
`test_skip` is used for in-test skipping, in which case both timestamps are not None
</blockquote></details>

It seems odd to have the same type of result (SKIP) have different behavior due to the cause of the skip.

How about instead of setting them to NONE here, set them to the same value so that a computed duration will be zero?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/269)*

xpconanfan:

---

*[mobly/base_test.py, line 506 at r1](https://reviewable.io:443/reviews/google/mobly/269#-KpuoMgbdJ67hzZXQmjI:-KpvGc2ymMcCwT-TOH9S:bqcsjx) ([raw file](https://github.com/google/mobly/blob/499531fe09c3b66b7b639f3a8d9ff5c1017c6083/mobly/base_test.py#L506)):*

It seems odd to have the same type of result (SKIP) have different behavior due to the cause of the skip.

How about instead of setting them to NONE here, set them to the same value so that a computed duration will be zero?
</blockquote></details>

It makes sense because when setup skips, the test wouldn't even start, so start and end time are undefined.
In other cases, the skip happens after the test starts, so the test actually has a start and end time:

```
def test_a(self):
  ...
  asserts.skip
  ...
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/269)*

dthkao:

---

*[mobly/base_test.py, line 506 at r1](https://reviewable.io:443/reviews/google/mobly/269#-KpuoMgbdJ67hzZXQmjI:-KpvPCHDATRNW9cEsIY5:b-16javx) ([raw file](https://github.com/google/mobly/blob/499531fe09c3b66b7b639f3a8d9ff5c1017c6083/mobly/base_test.py#L506)):*

It makes sense because when setup skips, the test wouldn't even start, so start and end time are undefined.
In other cases, the skip happens after the test starts, so the test actually has a start and end time:

```
def test_a(self):
  ...
  asserts.skip
  ...
```
</blockquote></details>

But, it's inconsistent behavior and anyone looking at TestResultRecord.test_skip()  would fairly assume that there would at least be a non-null value for test end time (and it's not a stretch to assume that if there is a non-null end there is a non-null start).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/269)*

xpconanfan:

---

*[mobly/base_test.py, line 506 at r1](https://reviewable.io:443/reviews/google/mobly/269#-KpuoMgbdJ67hzZXQmjI:-KpvQ5bBqTsoktINZRoL:bt56qfy) ([raw file](https://github.com/google/mobly/blob/499531fe09c3b66b7b639f3a8d9ff5c1017c6083/mobly/base_test.py#L506)):*

But, it's inconsistent behavior and anyone looking at TestResultRecord.test_skip()  would fairly assume that there would at least be a non-null value for test end time (and it's not a stretch to assume that if there is a non-null end there is a non-null start).
</blockquote></details>

it's not supposed to be consistent because in one case the test execution started, in the other case the test execution never started.

How do you suggest we differentiate these two cases?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/269)*

xpconanfan:

---

*[mobly/base_test.py, line 506 at r1](https://reviewable.io:443/reviews/google/mobly/269#-KpuoMgbdJ67hzZXQmjI:-KqPJf0WowyCQSO84d9d:b-896fix) ([raw file](https://github.com/google/mobly/blob/499531fe09c3b66b7b639f3a8d9ff5c1017c6083/mobly/base_test.py#L506)):*

it's not supposed to be consistent because in one case the test execution started, in the other case the test execution never started.

How do you suggest we differentiate these two cases?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/269)*

dthkao:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/269#-:-Kr2bH0vQhRmGnZ2KF9C:bnfp4nl)*

PR #274, opened 2017-07-27
Fix timeout logic in AndroidDevice._wait_for_device

l-meng:
Change function name from AndroidDevice._wait_for_device to utils.wait_until()
Make it public as general utility to be used by users
Make _is_boot_completed and _is_adb_detectable to be public

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/274)

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -20,6 +20,7 @@ import contextlib
 import logging
 import os
 import time
+import timeout_decorator
 
 from mobly import logger as mobly_logger
 from mobly import signals
@@ -95,7 +96,7 @@ def create(configs):
     for ad in ads:
         if ad.serial not in valid_ad_identifiers:
             raise DeviceError(ad, 'Android device is specified in config but'
-                              ' is not attached.')
+                                  ' is not attached.')
     _start_services_on_ads(ads)
     return ads
 
@@ -250,8 +251,8 @@ def get_instances_with_configs(configs):
             serial = c.pop('serial')
         except KeyError:
             raise Error(
-                'Required value "serial" is missing in AndroidDevice config %s.'
-                % c)
+                    'Required value "serial" is missing in AndroidDevice config %s.'
+                    % c)
         is_required = c.get(KEY_DEVICE_REQUIRED, True)
         try:
             ad = AndroidDevice(serial)
@@ -331,8 +332,8 @@ def get_device(ads, **kwargs):
     filtered = filter_devices(ads, _get_device_filter)
     if not filtered:
         raise Error(
-            'Could not find a target device that matches condition: %s.' %
-            kwargs)
+                'Could not find a target device that matches condition: %s.' %
+                kwargs)
     elif len(filtered) == 1:
         return filtered[0]
     else:
@@ -670,9 +671,9 @@ class AndroidDevice(object):
         for k, v in config.items():
             if hasattr(self, k):
                 raise DeviceError(
-                    self,
-                    ('Attribute %s already exists with value %s, cannot set '
-                     'again.') % (k, getattr(self, k)))
+                        self,
+                        ('Attribute %s already exists with value %s, cannot set '
+                         'again.') % (k, getattr(self, k)))
             setattr(self, k, v)
 
     def root_adb(self):
@@ -705,25 +706,25 @@ class AndroidDevice(object):
         # Should not load snippet with the same attribute more than once.
         if name in self._snippet_clients:
             raise SnippetError(
-                self,
-                'Attribute "%s" is already registered with package "%s", it '
-                'cannot be used again.' %
-                (name, self._snippet_clients[name].package))
+                    self,
+                    'Attribute "%s" is already registered with package "%s", it '
+                    'cannot be used again.' %
+                    (name, self._snippet_clients[name].package))
         # Should not load snippet with an existing attribute.
         if hasattr(self, name):
             raise SnippetError(
-                self,
-                'Attribute "%s" already exists, please use a different name.' %
-                name)
+                    self,
+                    'Attribute "%s" already exists, please use a different name.' %
+                    name)
         # Should not load the same snippet package more than once.
         for client_name, client in self._snippet_clients.items():
             if package == client.package:
                 raise SnippetError(
-                    self,
-                    'Snippet package "%s" has already been loaded under name'
-                    ' "%s".' % (package, client_name))
+                        self,
+                        'Snippet package "%s" has already been loaded under name'
+                        ' "%s".' % (package, client_name))
         client = snippet_client.SnippetClient(
-            package=package, adb_proxy=self.adb, log=self.log)
+                package=package, adb_proxy=self.adb, log=self.log)
         client.start_app_and_connect()
         self._snippet_clients[name] = client
         setattr(self, name, client)
@@ -759,8 +760,8 @@ class AndroidDevice(object):
         """
         if not self.adb_logcat_file_path:
             raise DeviceError(
-                self,
-                'Attempting to cat adb log when none has been collected.')
+                    self,
+                    'Attempting to cat adb log when none has been collected.')
         end_time = mobly_logger.get_log_line_timestamp()
         self.log.debug('Extracting adb log from logcat.')
         adb_excerpt_path = os.path.join(self.log_path, 'AdbLogExcerpts')
@@ -808,8 +809,8 @@ class AndroidDevice(object):
         """
         if self._adb_logcat_process:
             raise DeviceError(
-                self,
-                'Logcat thread is already running, cannot start another one.')
+                    self,
+                    'Logcat thread is already running, cannot start another one.')
         if clear_log:
             self._clear_adb_log()
         # Disable adb log spam filter for rootable devices. Have to stop and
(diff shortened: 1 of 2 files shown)


xpconanfan:

---

*[mobly/utils.py, line 481 at r1](https://reviewable.io:443/reviews/google/mobly/274#-Kq0ZA7_ijdnQT9qLgWB:-Kq0ZA7_ijdnQT9qLgWC:biw3c9b) ([raw file](https://github.com/google/mobly/blob/c98227a729739163b018e763858fbe2becc248e4/mobly/utils.py#L481)):*
> ```Python
>                     return
>             except Exception:
>                 # adb shell calls may fail during certain period of booting
> ```

this function has nothing to do with adb, why do we want to refer to adb in comments here?

---

*[mobly/controllers/android_device.py, line 933 at r1](https://reviewable.io:443/reviews/google/mobly/274#-Kq0ZcLDpz74N676yflb:-Kq0ZcLDpz74N676yflc:bq6we33) ([raw file](https://github.com/google/mobly/blob/c98227a729739163b018e763858fbe2becc248e4/mobly/controllers/android_device.py#L933)):*
> ```Python
>         utils.wait_until(self.is_boot_completed, timeout)
> 
>     def wait_for_adb_detection(self, timeout=DEFAULT_TIMEOUT_USB_ON):
> ```

can't we simply use the `timeout_decorator` on this function with `DEFAULT_TIMEOUT_USB_ON`, and make this a busy wait? Something like:

```
@timeout_decorator
def wait_for_adb_detection(self):
    while True:
        # break if condition is met
        time.sleep(1)
```

---

*[mobly/controllers/android_device.py, line 965 at r1](https://reviewable.io:443/reviews/google/mobly/274#-Kq0ZKbHdPycHn_K73uW:-Kq0ZKbHdPycHn_K73uX:b-2vbyuu) ([raw file](https://github.com/google/mobly/blob/c98227a729739163b018e763858fbe2becc248e4/mobly/controllers/android_device.py#L965)):*
> ```Python
>         """Checks if device boot is completed by verifying system property.
> 
>         Note that adb.wait_for_device() could be blocking in some cases.
> ```

`adb.wait_for_device` is always blocking.
The point is that it might block for a long time.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/274)*

l-meng:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[mobly/utils.py, line 481 at r1](https://reviewable.io:443/reviews/google/mobly/274#-Kq0ZA7_ijdnQT9qLgWB:-Kq0_Pe0j1A7zg0p3VYW:b-896fix) ([raw file](https://github.com/google/mobly/blob/c98227a729739163b018e763858fbe2becc248e4/mobly/utils.py#L481)):*

this function has nothing to do with adb, why do we want to refer to adb in comments here?
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 933 at r1](https://reviewable.io:443/reviews/google/mobly/274#-Kq0ZcLDpz74N676yflb:-Kq0_tDuOsYZhU9p-Pn6:b-e22xzu) ([raw file](https://github.com/google/mobly/blob/c98227a729739163b018e763858fbe2becc248e4/mobly/controllers/android_device.py#L933)):*

can't we simply use the `timeout_decorator` on this function with `DEFAULT_TIMEOUT_USB_ON`, and make this a busy wait? Something like:

```
@timeout_decorator
def wait_for_adb_detection(self):
    while True:
        # break if condition is met
        time.sleep(1)
```
</blockquote></details>

In that case user wouldn't be able to overwrite the default timeout (unless modify package level DEFAULT_TIMEOUT_USB_ON)
Do you suggest user to write their own wait_for_adb_detection() if a different timeout value is needed?

Also, with device showing up on adb, should we just return or we also need to confirm self.adb.wait_for_device()

---

*[mobly/controllers/android_device.py, line 965 at r1](https://reviewable.io:443/reviews/google/mobly/274#-Kq0ZKbHdPycHn_K73uW:-Kq0_sO4h2cUQs-HoRGy:b-896fix) ([raw file](https://github.com/google/mobly/blob/c98227a729739163b018e763858fbe2becc248e4/mobly/controllers/android_device.py#L965)):*

`adb.wait_for_device` is always blocking.
The point is that it might block for a long time.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/274)*

xpconanfan:

---

*[mobly/controllers/android_device.py, line 933 at r1](https://reviewable.io:443/reviews/google/mobly/274#-Kq0ZcLDpz74N676yflb:-Kq0boc4iJYXN1C-q97v:b2f3zo6) ([raw file](https://github.com/google/mobly/blob/c98227a729739163b018e763858fbe2becc248e4/mobly/controllers/android_device.py#L933)):*
<details><summary><i>Previously, l-meng wrote…</i></summary><blockquote>

In that case user wouldn't be able to overwrite the default timeout (unless modify package level DEFAULT_TIMEOUT_USB_ON)
Do you suggest user to write their own wait_for_adb_detection() if a different timeout value is needed?

Also, with device showing up on adb, should we just return or we also need to confirm self.adb.wait_for_device()
</blockquote></details>

Do we have any use case where users wanted to use a different value?
If we don't have a use case and it makes the code simpler, I'd vote for simpler code...

Just return if the function terminates without timeout is fine.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/274)*

l-meng:

---

*[mobly/controllers/android_device.py, line 933 at r1](https://reviewable.io:443/reviews/google/mobly/274#-Kq0ZcLDpz74N676yflb:-Kq0dghafKqysPN_-OqQ:b-260axi) ([raw file](https://github.com/google/mobly/blob/c98227a729739163b018e763858fbe2becc248e4/mobly/controllers/android_device.py#L933)):*

Do we have any use case where users wanted to use a different value?
If we don't have a use case and it makes the code simpler, I'd vote for simpler code...

Just return if the function terminates without timeout is fine.
</blockquote></details>

From the battery test case, it depends what the test is doing. If it's something that takes time, for example, start an app, and people may want to do it for 1000 times and measure the over all battery usage. The device offline time would be over 5 minutes. test needs to wait for longer time for snippet to complete.

If people want to just test one time click, then 5min is ok

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/274)*

xpconanfan:

---

*[mobly/controllers/android_device.py, line 933 at r1](https://reviewable.io:443/reviews/google/mobly/274#-Kq0ZcLDpz74N676yflb:-Kq0eLtWjwoTo0r6VFTR:b-xxq7ee) ([raw file](https://github.com/google/mobly/blob/c98227a729739163b018e763858fbe2becc248e4/mobly/controllers/android_device.py#L933)):*
<details><summary><i>Previously, l-meng wrote…</i></summary><blockquote>

From the battery test case, it depends what the test is doing. If it's something that takes time, for example, start an app, and people may want to do it for 1000 times and measure the over all battery usage. The device offline time would be over 5 minutes. test needs to wait for longer time for snippet to complete.

If people want to just test one time click, then 5min is ok
</blockquote></details>

The test shouldn't be using this function to wait for the test to finish, should it?
This is used as a short term wait after the test intentionally finishes power measurement and reconnects the device's USB.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/274)*

l-meng:

---

*[mobly/controllers/android_device.py, line 933 at r1](https://reviewable.io:443/reviews/google/mobly/274#-Kq0ZcLDpz74N676yflb:-Kq0ewZzc7VHj2qwiCY5:b-y7gs7w) ([raw file](https://github.com/google/mobly/blob/c98227a729739163b018e763858fbe2becc248e4/mobly/controllers/android_device.py#L933)):*

The test shouldn't be using this function to wait for the test to finish, should it?
This is used as a short term wait after the test intentionally finishes power measurement and reconnects the device's USB.
</blockquote></details>

Because on python side, device is offline at that time, the test code is totally blind. And the offline duration could be arbitrary time depending on many factors, e.g., 
- user unexpected: if something fails on-device and snippet performs retry. e.g., totally takes 8 min
  - in this case, user wouldn't know how much it will take, and try to make an estimate time.sleep(X) before calling this method.
- user expected: user knows how much the test will take, e.g., it takes 8 min
  - in this case, to make this function simpler, means that, user would have to time.sleep(X) in their code first, and they call this function to make sure wait time is < 5min

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/274)*

l-meng:

---

*[mobly/controllers/android_device.py, line 933 at r1](https://reviewable.io:443/reviews/google/mobly/274#-Kq0ZcLDpz74N676yflb:-Kq0gnH8BdaB6sGE9BAF:boohw7s) ([raw file](https://github.com/google/mobly/blob/c98227a729739163b018e763858fbe2becc248e4/mobly/controllers/android_device.py#L933)):*
<details><summary><i>Previously, l-meng wrote…</i></summary><blockquote>

Because on python side, device is offline at that time, the test code is totally blind. And the offline duration could be arbitrary time depending on many factors, e.g., 
- user unexpected: if something fails on-device and snippet performs retry. e.g., totally takes 8 min
  - in this case, user wouldn't know how much it will take, and try to make an estimate time.sleep(X) before calling this method.
- user expected: user knows how much the test will take, e.g., it takes 8 min
  - in this case, to make this function simpler, means that, user would have to time.sleep(X) in their code first, and they call this function to make sure wait time is < 5min

</blockquote></details>

however, in both cases, user could just time.sleep(20*60) to be safe, if they don't care about test time

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/274)*

xpconanfan:
Shall we close this and upload a new one based on offline discussion?

Issue #264, opened 2017-07-17
Consistent behavior from errors in setup and teardown


dthkao:
Currently we have the following

Exception in `setup_class`: Extra result entry, mark test cases as skipped
Exception in `setup_test`: Mark test as error using details
Exception in `teardown_test`: Mark test as error using extra_errors
Exception in `teardown_class`: No record

I propose the following for the sake of consistency (both with pyunit and with self):

Exception in `setup_class`:
    No extra result entry, mark test cases as skipped, add to extra_errors for each result
Exception in `setup_test`: Either
    1)  mark test case result as skipped and use extra_errors for error message
    2)  mark test case as error and use details
Exception in `teardown_test`: Either (while being consistent with `setup_test`)
    3)  Do not change the test case result (leave it as is) and use `extra_errors` for error message
    4)  mark test case as error and use `details`
Exception in `teardown_class`:
    Do not change the test case results, add to `extra_errors` for all non-skipped results

xpconanfan:
1. There is an additional entry for setup_class in this case and it is marked error. Maybe the entries for test cases shouldn't have error msg at all as they are just repeating the same thing from setup_test's entry, which can be a lot of repetition with no added value and make the summary hard to read.
2. no change needed?
3. This is for consistency with pyunit and other internal test frameworks.
4. This is a bug. teardown_class's reporting behavior should also be consistent with that of pyunit's `tearDownClass`.

dthkao:
1. This leads to an inconsistent number of records.

2. & 3. This behavior should be coupled and self consistent. If current behavior for teardown_test is considered "consistent" with pyunit, then option 1 would be consistent with pyunit. Additionally, the assertiong that current behavior for 3 is consistent with pyunit is not quite correct: in pyunit it's possible to pass a test but have an error in teardown, with the test still recorded as pass.

dthkao:
Also re: point 1, the following test has multiple tests but a setUpClass error results in

$ python pytesttest.py 
E
======================================================================
ERROR: setUpClass (__main__.TestErrorReporting)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "pytesttest.py", line 7, in setUpClass
    raise Exception('ouch class')
Exception: ouch class

----------------------------------------------------------------------
Ran 0 tests in 0.002s

FAILED (errors=1)

xpconanfan:
Discussed offline.
Things that need fixing:
[...]

Issue #281, opened 2017-07-31
Provide more information in conditional functions like `on_fail`


xpconanfan:
The current conditional functions have signatures like:
```
def on_fail(self, test_name, begin_time)
```

This has very limited information. For example, if a test fails with multiple errors, which is very common in large e2e tests, the `on_fail` would have no way to know exactly what failed.

We should provide more information to the test in these functions.
We could do this by passing in a copy of the test record for a particular test.

xpconanfan:
Fixed by #301

PR #284, opened 2017-08-01
Properly record extra errors.

### LINKED ISSUE #261: Include stack traces of extra errors
Right now only the stack trace for the main error is recorded in `TeatResultRecord`.
We should store the stack traces for extra errors as well.

xpconanfan:
* Bump the first extra error to the main error of the record, if the test implicitly passed.
* Adjust unit tests to check the new behavior.
* Fix docstring for `TestResultRecord`.

This is the first PR for #264, more to come.
Fixes #261 

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/284)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -327,6 +327,7 @@ class BaseTestClass(object):
         tr_record = records.TestResultRecord(test_name, self.TAG)
         tr_record.test_begin()
         logging.info('%s %s', TEST_CASE_TOKEN, test_name)
+        teardown_test_failed = False
         try:
             try:
                 self._setup_test(test_name)
@@ -348,6 +349,7 @@ class BaseTestClass(object):
                 except Exception as e:
                     logging.exception(e)
                     tr_record.add_error('teardown_test', e)
+                    teardown_test_failed = True
         except (signals.TestFailure, AssertionError) as e:
             tr_record.test_fail(e)
         except signals.TestSkip as e:
@@ -364,7 +366,8 @@ class BaseTestClass(object):
             # Exception happened during test.
             tr_record.test_error(e)
         else:
-            tr_record.test_pass()
+            if not teardown_test_failed:
+                tr_record.test_pass()
         finally:
             if tr_record.result in (records.TestResultEnums.TEST_RESULT_ERROR,
                                     records.TestResultEnums.TEST_RESULT_FAIL):
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -301,9 +301,9 @@ class BaseTestClass(object):
                 executed.
         """
         try:
-          # Pass a copy of the record instead of the actual object so that it
-          # will not be modified.
-          func(copy.deepcopy(tr_record))
+            # Pass a copy of the record instead of the actual object so that it
+            # will not be modified.
+            func(copy.deepcopy(tr_record))
         except signals.TestAbortAll:
             raise
         except Exception as e:
@@ -369,6 +369,7 @@ class BaseTestClass(object):
             if not teardown_test_failed:
                 tr_record.test_pass()
         finally:
+            tr_record.update_record()
             if tr_record.result in (records.TestResultEnums.TEST_RESULT_ERROR,
                                     records.TestResultEnums.TEST_RESULT_FAIL):
                 self._exec_procedure_func(self._on_fail, tr_record)
--- a/mobly/records.py
+++ b/mobly/records.py
@@ -218,10 +218,21 @@ class TestResultRecord(object):
         self.termination_signal = None
         self.extra_errors = collections.OrderedDict()
         self.result = None
-        # The fields below are based on termination_signal's content.
-        self.extras = None
-        self.details = None
-        self.stacktrace = None
+
+    @property
+    def details(self):
+        if self.termination_signal:
+            return self.termination_signal.details
+
+    @property
+    def stacktrace(self):
+        if self.termination_signal:
+            return self.termination_signal.stacktrace
+
+    @property
+    def extras(self):
+        if self.termination_signal:
+            return self.termination_signal.extras
 
     def test_begin(self):
         """Call this when the test begins execution.
@@ -255,17 +266,12 @@ class TestResultRecord(object):
         the appropirate fields.
         """
         if self.extra_errors:
-            if self.result in (TestResultEnums.TEST_RESULT_PASS,
-                               TestResultEnums.TEST_RESULT_SKIP):
+            if self.result != TestResultEnums.TEST_RESULT_FAIL:
                 self.result = TestResultEnums.TEST_RESULT_ERROR
         # If no termination signal is provided, use the first exception
         # occurred as the termination signal.
         if not self.termination_signal and self.extra_errors:
             _, self.termination_signal = self.extra_errors.popitem(last=False)
-        if self.termination_signal:
-            self.details = self.termination_signal.details
-            self.stacktrace = self.termination_signal.stacktrace
-            self.extras = self.termination_signal.extras
 
     def test_pass(self, e=None):
         """To mark the test as passed in this record.
@@ -320,9 +326,11 @@ class TestResultRecord(object):
             position: string, where this error occurred, e.g. 'teardown_test'.
             e: An exception object.
         """
-        if self.result in (None, TestResultEnums.TEST_RESULT_PASS,
-                           TestResultEnums.TEST_RESULT_SKIP):
+        if self.result != TestResultEnums.TEST_RESULT_FAIL:
             self.result = TestResultEnums.TEST_RESULT_ERROR
+        if position in self.extra_errors:
+            raise Error('An exception is already recorded with position "%s",'
+                        ' cannot reuse.' % position)
         self.extra_errors[position] = ExceptionRecord(e, position=position)
 
     def __str__(self):
--- a/mobly/records.py
+++ b/mobly/records.py
@@ -196,17 +196,10 @@ class TestResultRecord(object):
         begin_time: Epoch timestamp of when the test started.
         end_time: Epoch timestamp of when the test ended.
         uid: Unique identifier of a test.
+        termination_signal: ExceptionRecord, the main exception of the test.
+        extra_errors: OrderedDict, all exceptions occurred during the entire
+            test lifecycle. The order of occurrence is preserved.
         result: TestResultEnum.TEAT_RESULT_*, PASS/FAIL/SKIP.
-        extras: User defined extra information of the test result, must be
-                serializable.
-        details: string, description of the cause of the test's termination.
-                 Note a passed test can have this as well due to the explicit
-                 pass signal. If the test passed implicitly, this field would
-                 be None.
-        stacktrace: string, the stacktrace for the exception that terminated
-                    the test.
-        extra_errors: OrderedDict, all exceptions occurred during the entire test
-                    lifecycle. The order of occurrence is preserved.
     """
 
     def __init__(self, t_name, t_class=None):
@@ -221,16 +214,27 @@ class TestResultRecord(object):
 
     @property
     def details(self):
+        """String description of the cause of the test's termination.
+
+        Note a passed test can have this as well due to the explicit pass
+        signal. If the test passed implicitly, this field would be None.
+        """
         if self.termination_signal:
             return self.termination_signal.details
 
     @property
     def stacktrace(self):
+        """The stacktrace string for the exception that terminated the test.
+        """
         if self.termination_signal:
             return self.termination_signal.stacktrace
 
     @property
     def extras(self):
+        """User defined extra information of the test result.
+
+        Must be serializable.
+        """
         if self.termination_signal:
             return self.termination_signal.extras
 
@@ -247,8 +251,8 @@ class TestResultRecord(object):
         Args:
             result: One of the TEST_RESULT enums in TestResultEnums.
             e: A test termination signal (usually an exception object). It can
-               be any exception instance or of any subclass of
-               mobly.signals.TestSignal.
+                be any exception instance or of any subclass of
+                mobly.signals.TestSignal.
         """
         if self.begin_time is not None:
             self.end_time = utils.get_current_epoch_time()
--- a/mobly/records.py
+++ b/mobly/records.py
@@ -127,6 +127,44 @@ class TestResultEnums(object):
     TEST_RESULT_ERROR = 'ERROR'
 
 
+class ExceptionRecord(object):
+    """A wrapper class for representing exception objects in TestResultRecord.
+    """
+
+    def __init__(self, e, position=None):
+        self.exception = e
+        self.stacktrace = None
+        self.extras = None
+        self.position = position
+        self.is_test_signal = isinstance(e, signals.TestSignal)
+        # Record stacktrace of the exception.
+        # This check cannot be based on try...except, which messes up
+        # `exc_info`.
+        if hasattr(e, '__traceback__'):
+            exc_traceback = e.__traceback__
+        else:
+            # In py2, exception objects don't have built-in traceback, so we
+            # have to immediately retrieve stacktrace from `sys.exc_info`.
+            _, _, exc_traceback = sys.exc_info()
+        if exc_traceback:
+            self.stacktrace = ''.join(
+                traceback.format_exception(e.__class__, e, exc_traceback))
+        # Populate fields based on the type of the termination signal.
+        if self.is_test_signal:
+            self.details = str(e.details)
+            self.extras = e.extras
+        else:
+            self.details = str(e)
+
+    def to_dict(self):
+        result = {}
+        result[TestResultEnums.RECORD_DETAILS] = self.details
+        result[TestResultEnums.RECORD_POSITION] = self.position
+        result[TestResultEnums.RECORD_STACKTRACE] = self.stacktrace
+        result[TestResultEnums.RECORD_EXTRAS] = copy.deepcopy(self.extras)
+        return result
+
+
 class TestResultRecord(object):
     """A record that holds the information of a single test.
 
@@ -169,12 +207,13 @@ class TestResultRecord(object):
         self.begin_time = None
         self.end_time = None
         self.uid = None
+        self.termination_signal = None
+        self.extra_errors = collections.OrderedDict()
         self.result = None
+        # The fields below are based on termination_signal's content.
         self.extras = None
         self.details = None
         self.stacktrace = None
-        self.termination_signal = None
-        self.extra_errors = collections.OrderedDict()
 
     def test_begin(self):
         """Call this when the test begins execution.
@@ -184,10 +223,7 @@ class TestResultRecord(object):
         self.begin_time = utils.get_current_epoch_time()
 
     def _test_end(self, result, e):
-        """Class internal function to signal the end of a test execution.
-
-        This method finalizes the content of a record, so no modification
-        should be made after this method is called.
+        """Marks the end of the test logic.
 
         Args:
             result: One of the TEST_RESULT enums in TestResultEnums.
@@ -198,39 +234,22 @@ class TestResultRecord(object):
         if self.begin_time is not None:
             self.end_time = utils.get_current_epoch_time()
         self.result = result
-        if self.extra_errors:
-            self.result = TestResultEnums.TEST_RESULT_ERROR
-        failure_location = ''
-        # If a termination signal is provided, record it.
         if e:
-            self.termination_signal = e
-        # Otherwise, use the first exception occurred to populate fields like
-        # details and extras.
-        elif e is None and self.extra_errors:
-            failure_location, e = self.extra_errors.popitem(last=False)
-        if failure_location:
-            failure_location += ': '
-        # Populate record fields based on the type of the termination signal.
-        if isinstance(e, signals.TestSignal):
-            self.details = failure_location + str(e.details)
-            self.extras = e.extras
-        elif isinstance(e, Exception):
-            self.details = failure_location + str(e)
-        # Record stacktrace of the exception.
-        # This check cannot be based on try...except, which messes up
-        # `exc_info`.
-        if e and hasattr(e, '__traceback__'):
-            exc_traceback = e.__traceback__
-        else:
-            # In py2, exception objects don't have built-in traceback, so we
-            # have to immediately retrieve stacktrace from `sys.exc_info`.
-            _, _, exc_traceback = sys.exc_info()
-        if exc_traceback:
-            stacktrace_str = ''.join(
-                traceback.format_exception(e.__class__, e, exc_traceback))
-            self.stacktrace = stacktrace_str
-            if e and not hasattr(e, 'stacktrace_str'):
-                e.stacktrace_str = stacktrace_str
+            self.termination_signal = ExceptionRecord(e)
+
+    def finalize_record(self):
+        if self.extra_errors:
+            if self.result in (TestResultEnums.TEST_RESULT_PASS,
+                               TestResultEnums.TEST_RESULT_SKIP):
+                self.result = TestResultEnums.TEST_RESULT_ERROR
+        # If no termination signal is provided, use the first exception
+        # occurred as the termination signal.
+        if not self.termination_signal and self.extra_errors:
+            _, self.termination_signal = self.extra_errors.popitem(last=False)
+        if self.termination_signal:
+            self.details = self.termination_signal.details
+            self.stacktrace = self.termination_signal.stacktrace
+            self.extras = self.termination_signal.extras
 
     def test_pass(self, e=None):
         """To mark the test as passed in this record.
@@ -269,7 +288,9 @@ class TestResultRecord(object):
         self._test_end(TestResultEnums.TEST_RESULT_ERROR, e)
 
     def add_error(self, position, e):
-        """Add extra error happened during a test mark the test result as
+        """Add extra error happened during a test.
+
+        If the test has passed or skipped, this will mark the test result as
         ERROR.
 
         If an error is added the test record, the record's result is equivalent
@@ -283,21 +304,10 @@ class TestResultRecord(object):
             position: string, where this error occurred, e.g. 'teardown_test'.
             e: An exception object.
         """
-        self.result = TestResultEnums.TEST_RESULT_ERROR
-        self.extra_errors[position] = e
-        # This check cannot be based on try...except, which messes up
-        # `exc_info`.
-        if e and hasattr(e, '__traceback__'):
-            exc_traceback = e.__traceback__
-        else:
-            # In py2, exception objects don't have built-in traceback, so we
-            # have to immediately retrieve stacktrace from `sys.exc_info`.
-            _, _, exc_traceback = sys.exc_info()
-        if exc_traceback:
-            stacktrace_str = ''.join(
-                traceback.format_exception(e.__class__, e, exc_traceback))
-            if not hasattr(e, 'stacktrace_str'):
-                e.stacktrace_str = stacktrace_str
+        if self.result in (None, TestResultEnums.TEST_RESULT_PASS,
+                           TestResultEnums.TEST_RESULT_SKIP):
+            self.result = TestResultEnums.TEST_RESULT_ERROR
+        self.extra_errors[position] = ExceptionRecord(e, position=position)
 
     def __str__(self):
         d = self.to_dict()
@@ -309,30 +319,6 @@ class TestResultRecord(object):
         """This returns a short string representation of the test record."""
         t = utils.epoch_to_human_time(self.begin_time)
         return '%s %s %s' % (t, self.test_name, self.result)
-
-    def _get_extra_errors_list(self):
-        """Gets a list of dicts representing the extra errors in order.
-
-        Converts the exception objects for extra errors to dicts, so they can
-        be properly serialized.
-
-        Returns:
-            A list of dicts, each dict represent an exception. The order of the
-            list matches the order of the errors occurred in a test.
-        """
-        results = []
-        for position, e in self.extra_errors.items():
-            exception_dict = {}
-            exception_dict[TestResultEnums.RECORD_DETAILS] = str(e)
-            exception_dict[TestResultEnums.RECORD_POSITION] = position
-            exception_dict[
-                TestResultEnums.RECORD_STACKTRACE] = e.stacktrace_str
-            try:
-                exception_dict[TestResultEnums.RECORD_EXTRAS] = e.extras
-            except AttributeError:
-                pass
-            results.append(exception_dict)
-        return results
 
     def to_dict(self):
         """Gets a dictionary representating the content of this class.
@@ -349,7 +335,9 @@ class TestResultRecord(object):
         d[TestResultEnums.RECORD_UID] = self.uid
         d[TestResultEnums.RECORD_EXTRAS] = self.extras
         d[TestResultEnums.RECORD_DETAILS] = self.details
-        d[TestResultEnums.RECORD_EXTRA_ERRORS] = self._get_extra_errors_list()
+        d[TestResultEnums.RECORD_EXTRA_ERRORS] = [
+            e.to_dict() for e in self.extra_errors.values()
+        ]
         d[TestResultEnums.RECORD_STACKTRACE] = self.stacktrace
         return d
 
@@ -433,6 +421,7 @@ class TestResult(object):
         Args:
             record: A test record object to add.
         """
+        record.finalize_record()
         if record.result == TestResultEnums.TEST_RESULT_SKIP:
             self.skipped.append(record)
             return
@@ -465,6 +454,7 @@ class TestResult(object):
         Args:
             test_record: A TestResultRecord object for the test class.
         """
+        test_record.finalize_record()
         self.error.append(test_record)
 
     def is_test_executed(self, test_name):
--- a/mobly/records.py
+++ b/mobly/records.py
@@ -14,6 +14,7 @@
 """This module has classes for test result collection, and test result output.
 """
 
+import collections
 import itertools
 import copy
 import enum
@@ -126,16 +127,24 @@ class TestResultEnums(object):
 
 
 class TestResultRecord(object):
-    """A record that holds the information of a test execution.
+    """A record that holds the information of a single test case.
 
     Attributes:
-        test_name: A string representing the name of the test method.
+        test_name: string, the name of the test.
         begin_time: Epoch timestamp of when the test started.
         end_time: Epoch timestamp of when the test ended.
-        self.uid: Unique identifier of a test.
-        self.result: Test result, PASS/FAIL/SKIP.
-        self.extras: User defined extra information of the test result.
-        self.details: A string explaining the details of the test.
+        uid: Unique identifier of a test.
+        result: TestResultEnum.TEAT_RESULT_*, PASS/FAIL/SKIP.
+        extras: User defined extra information of the test result, must be
+                serializable.
+        details: string, description of the cause of the test's termination.
+                 Note a passed test can have this as well due to the explicit
+                 pass signal. If the test passed implicitly, this field would
+                 be None.
+        stacktrace: string, the stacktrace for the exception that terminated
+                    the test.
+        extra_errors: OrderedDict, any extra errors occurred during the entire
+                      test lifecycle. The order of occurrence is preserved.
     """
 
     def __init__(self, t_name, t_class=None):
@@ -148,7 +157,7 @@ class TestResultRecord(object):
         self.extras = None
         self.details = None
         self.stacktrace = None
-        self.extra_errors = {}
+        self.extra_errors = collections.OrderedDict()
 
     def test_begin(self):
         """Call this when the test begins execution.
@@ -160,28 +169,35 @@ class TestResultRecord(object):
     def _test_end(self, result, e):
         """Class internal function to signal the end of a test execution.
 
+        This method finalizes the content of a record, so no modification
+        should be made after this method is called.
+
         Args:
             result: One of the TEST_RESULT enums in TestResultEnums.
             e: A test termination signal (usually an exception object). It can
-                be any exception instance or of any subclass of
-                mobly.signals.TestSignal.
+               be any exception instance or of any subclass of
+               mobly.signals.TestSignal.
         """
         if self.begin_time is not None:
             self.end_time = utils.get_current_epoch_time()
         self.result = result
         if self.extra_errors:
             self.result = TestResultEnums.TEST_RESULT_ERROR
+        # When no explicit termination signal is provided, use the first extra
+        # error as the signal if there is any.
+        failure_location = ''
+        if e is None and self.extra_errors:
+            failure_location, e = self.extra_errors.popitem(last=False)
+        if failure_location:
+            failure_location += ': '
         if isinstance(e, signals.TestSignal):
-            self.details = e.details
-            _, _, exc_traceback = sys.exc_info()
-            if exc_traceback:
-                self.stacktrace = ''.join(traceback.format_tb(exc_traceback))
+            self.details = failure_location + str(e.details)
             self.extras = e.extras
         elif isinstance(e, Exception):
-            self.details = str(e)
-            _, _, exc_traceback = sys.exc_info()
-            if exc_traceback:
-                self.stacktrace = ''.join(traceback.format_tb(exc_traceback))
+            self.details = failure_location + str(e)
+        _, _, exc_traceback = sys.exc_info()
+        if exc_traceback:
+            self.stacktrace = ''.join(traceback.format_tb(exc_traceback))
 
     def test_pass(self, e=None):
         """To mark the test as passed in this record.
@@ -226,12 +242,16 @@ class TestResultRecord(object):
         If an error is added the test record, the record's result is equivalent
         to the case where an uncaught exception happened.
 
+        If the test record has not recorded any error, the newly added error
+        would be the main error of the test record. Otherwise the newly added
+        error is added to the record's extra errors.
+
         Args:
             tag: A string describing where this error came from, e.g. 'on_pass'.
             e: An exception object.
         """
         self.result = TestResultEnums.TEST_RESULT_ERROR
-        self.extra_errors[tag] = str(e)
+        self.extra_errors[tag] = e
 
     def __str__(self):
         d = self.to_dict()
@@ -259,7 +279,7 @@ class TestResultRecord(object):
         d[TestResultEnums.RECORD_UID] = self.uid
         d[TestResultEnums.RECORD_EXTRAS] = self.extras
         d[TestResultEnums.RECORD_DETAILS] = self.details
-        d[TestResultEnums.RECORD_EXTRA_ERRORS] = self.extra_errors
+        d[TestResultEnums.RECORD_EXTRA_ERRORS] = dict(self.extra_errors)
         d[TestResultEnums.RECORD_STACKTRACE] = self.stacktrace
         return d
 
--- a/mobly/records.py
+++ b/mobly/records.py
@@ -226,7 +226,8 @@ class TestResultRecord(object):
             # have to immediately retrieve stacktrace from `sys.exc_info`.
             _, _, exc_traceback = sys.exc_info()
         if exc_traceback:
-            stacktrace_str = ''.join(traceback.format_tb(exc_traceback))
+            stacktrace_str = ''.join(
+                traceback.format_exception(e.__class__, e, exc_traceback))
             self.stacktrace = stacktrace_str
             if e and not hasattr(e, 'stacktrace_str'):
                 e.stacktrace_str = stacktrace_str
@@ -293,7 +294,8 @@ class TestResultRecord(object):
             # have to immediately retrieve stacktrace from `sys.exc_info`.
             _, _, exc_traceback = sys.exc_info()
         if exc_traceback:
-            stacktrace_str = ''.join(traceback.format_tb(exc_traceback))
+            stacktrace_str = ''.join(
+                traceback.format_exception(e.__class__, e, exc_traceback))
             if not hasattr(e, 'stacktrace_str'):
                 e.stacktrace_str = stacktrace_str
 
--- a/mobly/records.py
+++ b/mobly/records.py
@@ -127,7 +127,7 @@ class TestResultEnums(object):
 
 
 class TestResultRecord(object):
-    """A record that holds the information of a single test case.
+    """A record that holds the information of a single test.
 
     Attributes:
         test_name: string, the name of the test.
--- a/mobly/records.py
+++ b/mobly/records.py
@@ -245,11 +245,14 @@ class TestResultRecord(object):
         if e:
             self.termination_signal = ExceptionRecord(e)
 
-    def finalize_record(self):
-        """Finalizes the content of a record.
+    def update_record(self):
+        """Updates the content of a record.
 
-        No write operation should be performed on the record after this is
-        called.
+        Several display fields like "details" and "stacktrace" need to be
+        updated based on the content of the record object.
+
+        As the content of the record change, call this method to update all
+        the appropirate fields.
         """
         if self.extra_errors:
             if self.result in (TestResultEnums.TEST_RESULT_PASS,
@@ -437,7 +440,7 @@ class TestResult(object):
         Args:
             record: A test record object to add.
         """
-        record.finalize_record()
+        record.update_record()
         if record.result == TestResultEnums.TEST_RESULT_SKIP:
             self.skipped.append(record)
             return
@@ -470,7 +473,7 @@ class TestResult(object):
         Args:
             test_record: A TestResultRecord object for the test class.
         """
-        test_record.finalize_record()
+        test_record.update_record()
         self.error.append(test_record)
 
     def is_test_executed(self, test_name):
--- a/mobly/records.py
+++ b/mobly/records.py
@@ -216,6 +216,9 @@ class TestResultRecord(object):
             self.extras = e.extras
         elif isinstance(e, Exception):
             self.details = failure_location + str(e)
+        # Record stacktrace of the exception.
+        # In py2, exception objects don't have built-in traceback, so we have
+        # to immediately retrieve stacktrace from `sys.exc_info`.
         _, _, exc_traceback = sys.exc_info()
         if exc_traceback:
             stacktrace_str = ''.join(traceback.format_tb(exc_traceback))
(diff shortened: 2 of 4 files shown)


dthkao:

---

*[mobly/records.py, line 122 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTgIHnBFx-nHQQcOAq:-KqTgIHnBFx-nHQQcOAr:b-acx593) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L122)):*
> ```Python
> 
> class TestResultRecord(object):
>     """A record that holds the information of a single test case.
> ```

"test case" nomenclature?

---

*[mobly/records.py, line 132 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTgS0N6lQ8TLdFtNJ5:-KqTgS0OviyvCih_C-0R:bnc8sd2) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L132)):*
> ```Python
>         extras: User defined extra information of the test result, must be
>                 serializable.
>         details: string, description of the cause of the test's termination.
> ```

nit: is yapf formatting this way? the different alignments for each arg are kind of distracting.

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-KqTiCM4ZWPPVnwvXdjG:boxqafn) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*
> ```Python
>         # error as the signal if there is any.
>         failure_location = ''
>         if e is None and self.extra_errors:
> ```

Just for my understanding, is there never a scenario where we have extra errors but the result is something other than ERROR or FAIL?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:

---

*[mobly/records.py, line 132 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTgS0N6lQ8TLdFtNJ5:-KqTlZLblhxBIk6Sik2R:b-sa7406) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L132)):*

nit: is yapf formatting this way? the different alignments for each arg are kind of distracting.
</blockquote></details>

Our convention is to indent to after ": ".

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-KqTlkFZb7qqCuCXoPuR:b-j9rm3x) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*

Just for my understanding, is there never a scenario where we have extra errors but the result is something other than ERROR or FAIL?
</blockquote></details>

If extra error exists, the test should be ERROR, otherwise it's a bug.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:

---

*[mobly/records.py, line 122 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTgIHnBFx-nHQQcOAq:-KqTm-oNT1t1p-LYmAAk:b-896fix) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L122)):*

"test case" nomenclature?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

dthkao:

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-KqTor4h_KJbD4nSbht7:by8gldi) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*

If extra error exists, the test should be ERROR, otherwise it's a bug.
</blockquote></details>

That does not seem right. If a test fails and then an error occurs in teardown the result should be fail.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-KqTp4dZkO_WrQ2rWfOE:b-wh33nw) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*

That does not seem right. If a test fails and then an error occurs in teardown the result should be fail.
</blockquote></details>

No, the result should be error in that case, as it is today and in all other frameworks (junit, pyunit etc).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

dthkao:

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-KqTpbZVQM6Vi-tcjDXF:b1f97c1) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*

No, the result should be error in that case, as it is today and in all other frameworks (junit, pyunit etc).
</blockquote></details>

1) Pyunit does not have a notion of an extra error,
2) It still records the failure
3) Your code does not seem to support that behavior: if there is a failure marked as failure that is not changed with the extra error message moved to details

$ cat pytesttest.py
davidkao@davidkao:~$ cat pytesttest.py 
import unittest

class TestErrorReporting(unittest.TestCase):

    def tearDown(self):
        raise Exception('ow')

    def test_fail(self):
        self.assertEqual('bar', 'FOO')

if __name__ == '__main__':
  unittest.main()

$ python pytesttest.py 
FE
======================================================================
ERROR: test_fail (__main__.TestErrorReporting)
----------------------------------------------------------------------
  File "pytesttest.py", line 16, in tearDown
    raise Exception('ow')
Exception: ow

======================================================================
FAIL: test_fail (__main__.TestErrorReporting)
----------------------------------------------------------------------
  File "pytesttest.py", line 23, in test_fail
    self.assertEqual('bar', 'FOO')
AssertionError: 'bar' != 'FOO'

----------------------------------------------------------------------
Ran 1 test in 0.001s

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-KqTrz_Hc9lHr0GvZNIT:b-pl5y07) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*

1) Pyunit does not have a notion of an extra error,
2) It still records the failure
3) Your code does not seem to support that behavior: if there is a failure marked as failure that is not changed with the extra error message moved to details

$ cat pytesttest.py
davidkao@davidkao:~$ cat pytesttest.py 
import unittest

class TestErrorReporting(unittest.TestCase):

    def tearDown(self):
        raise Exception('ow')

    def test_fail(self):
        self.assertEqual('bar', 'FOO')

if __name__ == '__main__':
  unittest.main()

$ python pytesttest.py 
FE
======================================================================
ERROR: test_fail (__main__.TestErrorReporting)
----------------------------------------------------------------------
  File "pytesttest.py", line 16, in tearDown
    raise Exception('ow')
Exception: ow

======================================================================
FAIL: test_fail (__main__.TestErrorReporting)
----------------------------------------------------------------------
  File "pytesttest.py", line 23, in test_fail
    self.assertEqual('bar', 'FOO')
AssertionError: 'bar' != 'FOO'

----------------------------------------------------------------------
Ran 1 test in 0.001s

</blockquote></details>

Pyunit always says "FAILED" in the end as long as it's not pass.
But if you see the detailed output, it singles out errors.

For our purposes, if something threw in `teardown_test`, then something went unexpectedly wrong, which translates to ERROR. Otherwise, a failed test with or without teardown error would be reported the same, and somebody would have to click the summary to read the extra errors to discover that something went unexpectedly wrong.

That behavior would make teardown errors hard to discover, which is not good. Hence we intentionally change the test result to error.

I don't really know what 3 means. if the test body explicitly Failed, then we won't move extra error to main.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

dthkao:

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-KqTumLKX8fA_esUqlpn:b-kc55ub) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*

> Pyunit always says "FAILED" in the end as long as it's not pass.
> But if you see the detailed output, it singles out errors.

It records them yes, but dow not single them out. I encourage you to run the simple testcase above. Your suggestion here would be to allow a teardown error to hide a test failure?

> For our purposes, if something threw in teardown_test, then something went unexpectedly wrong, which translates to ERROR. Otherwise, a failed test with or without teardown error would be reported the same, and somebody would have to click the summary to read the extra errors to discover that something went unexpectedly wrong.

"Otherwise, a failed test with or without teardown error would be reported the same" <-- this is exactly what should happen. The failure occurred first and them potentially caused extra errors.

> That behavior would make teardown errors hard to discover, which is not good. Hence we intentionally change the test result to error. 

We should only change the result to error if it was Pass before teardown. Please see comment in https://github.com/google/mobly/issues/264

> I don't really know what 3 means. if the test body explicitly Failed, then we won't move extra error to main.
What happens if somehow we call `self._test_end(TestResultEnums.TEST_RESULT_FAIL, None)`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-KqUJFOd69F66aABin7p:b-xxjkh1) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*


> Pyunit always says "FAILED" in the end as long as it's not pass.
> But if you see the detailed output, it singles out errors.

It records them yes, but dow not single them out. I encourage you to run the simple testcase above. Your suggestion here would be to allow a teardown error to hide a test failure?

> For our purposes, if something threw in teardown_test, then something went unexpectedly wrong, which translates to ERROR. Otherwise, a failed test with or without teardown error would be reported the same, and somebody would have to click the summary to read the extra errors to discover that something went unexpectedly wrong.

"Otherwise, a failed test with or without teardown error would be reported the same" <-- this is exactly what should happen. The failure occurred first and them potentially caused extra errors.

> That behavior would make teardown errors hard to discover, which is not good. Hence we intentionally change the test result to error. 

We should only change the result to error if it was Pass before teardown. Please see comment in https://github.com/google/mobly/issues/264

> I don't really know what 3 means. if the test body explicitly Failed, then we won't move extra error to main.
What happens if somehow we call `self._test_end(TestResultEnums.TEST_RESULT_FAIL, None)`?
</blockquote></details>

That comment only affects where the error description and stacktrace are recorded, it doesn't mention the final result of the test entry.
If error occurs in teardown, the test is ERROR; this is expected behavior.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
Made the change to properly record exception objects.
* Record stacktrace for each exception object.
* Record the actual exception object for the main termination signal.

This paves the road for better debug info in `on_fail` as required by #281

dthkao:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/records.py, line 132 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTgS0N6lQ8TLdFtNJ5:-Kr2ef8oe7Ird8oaa9l_:b7tfnuk) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L132)):*

Our convention is to indent to after ": ".
</blockquote></details>

Discussed offline, filed https://github.com/google/mobly/issues/300. Acknowledged for now

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-Kr2fHm7qvD8Mu8EFaL-:b-9exoj3) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*

That comment only affects where the error description and stacktrace are recorded, it doesn't mention the final result of the test entry.
If error occurs in teardown, the test is ERROR; this is expected behavior.
</blockquote></details>

It is not expected behavior. If a test fails in the test case an error in teardown should not cause that failure to vanish. This was one of the #1 reasons to reexamine test result recording. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-Kr2yMMfCWPH9jvSHs6B:bbpfs5s) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*

It is not expected behavior. If a test fails in the test case an error in teardown should not cause that failure to vanish. This was one of the #1 reasons to reexamine test result recording. 
</blockquote></details>

This is intentional.

1. test fail, teardown pass -> FAIL
2. teat fail, teardown error -> ERROR

The "failure" is not "lost" in case 2. The termination signal is still the original failure, the details and stacktrace fields still use the original failure's info.
But an error in teardown test *should* change the status of the test result because something unexpected happened in teardown.

A test is tallied as FAIL only if an explicit test failure signal is thrown in test body and no other errors occurred in the test.
If anything unexpected happen, the status is ERROR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
There's a bug where errors in functions like `on_fail` are not tallied properly.
Need another patch.

dthkao:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-Kr49v8Qlmz11BJJodd9:byltc7v) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*

This is intentional.

1. test fail, teardown pass -> FAIL
2. teat fail, teardown error -> ERROR

The "failure" is not "lost" in case 2. The termination signal is still the original failure, the details and stacktrace fields still use the original failure's info.
But an error in teardown test *should* change the status of the test result because something unexpected happened in teardown.

A test is tallied as FAIL only if an explicit test failure signal is thrown in test body and no other errors occurred in the test.
If anything unexpected happen, the status is ERROR.
</blockquote></details>

1) I disagree entirely. See the pytest example above. When a test fails and an error is thrown in teardown a failure and error are BOTH counted: there is a failure result and an EXTRA error.  If the details and stacktrace reflect the failure then the result should also. Otherwise this will only lead to confusion as to why a result with information coming from a mode of failure is marked as error. 

2) The logic implemented here itself is inconsistent with your claim. It is possible to call test_fail with no error message:
`def test_fail(self, e=None)` and in this scenario the code explicitly subverts the result marked by the user. to replace it with an 'extra_error'.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-Kr7vy8js6-dUfo3Pu30:bsgoujm) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*

1) I disagree entirely. See the pytest example above. When a test fails and an error is thrown in teardown a failure and error are BOTH counted: there is a failure result and an EXTRA error.  If the details and stacktrace reflect the failure then the result should also. Otherwise this will only lead to confusion as to why a result with information coming from a mode of failure is marked as error. 

2) The logic implemented here itself is inconsistent with your claim. It is possible to call test_fail with no error message:
`def test_fail(self, e=None)` and in this scenario the code explicitly subverts the result marked by the user. to replace it with an 'extra_error'.
</blockquote></details>

Regardless of what you agree, can we get this PR in?
This PR addresses the issue of exception objects not being stored properly in records.
As the initial commit message says, this is the first of multiple PRs to come.

We can address other concerns in later PRs.
We are keeping that tracking issue open until all the PRs are in.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

k2fong:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-Kr8PO87zIe_2us-T9V7:b-8lvomq) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*

Regardless of what you agree, can we get this PR in?
This PR addresses the issue of exception objects not being stored properly in records.
As the initial commit message says, this is the first of multiple PRs to come.

We can address other concerns in later PRs.
We are keeping that tracking issue open until all the PRs are in.
</blockquote></details>

Just giving my 2 cents here, my recommendation is to follow how pytest handles since that'll probably be the user's expectation of mobly.  From a little test of mine, looks like a test failure (with teardown error) results in a reporting of  "FAILED (failures=1, errors=1)".  Can we do similar?  Also, it provided stack trace for the failure in test and the error in teardown.  From an offline chat, sounds like to do so might be a big task, what you guys think about the path forward to just continue however mobly is reporting currently today so as to minimize user impacted in the meantime.  Unless that was partly the reason for this PR and I'm assuming users are either okay with how Mobly is reporting or has created some workaround.

---

*[mobly/records.py, line 197 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3inUC5cHYesCg3hIN:-Kr3inUC5cHYesCg3hIO:b-bvf5k5) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L197)):*
>         if e:
>             self.termination_signal = e
no need for this 'if e' clause.  Just assign?

---

*[mobly/records.py, line 200 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3jiRas15KiEg1ApFP:-Kr3jiRas15KiEg1ApFQ:bd3d4ok) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L200)):*
> e is None
e should be None at this point already

---

*[mobly/records.py, line 203 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3kD9eWG2D5rUDEOU-:-Kr3kD9eWG2D5rUDEOU0:b4eguto) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L203)):*
>         if failure_location:
>             failure_location += ': '
simplify one less logic by adding this inside the failure_location assignment above.  Unless its possible that the popitem() above might return None or ''

---

*[mobly/records.py, line 205 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3lPLRAOw3SbbSD8V7:-Kr3lPLRAOw3SbbSD8V8:bh6yzzm) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L205)):*
>         if isinstance(e, signals.TestSignal):
>             self.details = failure_location + str(e.details)
>  
>  
>  
>             self.extras = e.extras
>         elif isinstance(e, Exception):
>             self.details = failure_location + str(e)
Seems like if the logic enters into either of these, failure_location will be empty string.  Is that intended?

---

*[mobly/records.py, line 223 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3nm3ORrhOCtOJM6Km:-Kr3nm3ORrhOCtOJM6Kn:b1nwfy1) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L223)):*
>             if e and not hasattr(e, 'stacktrace_str'):
>                 e.stacktrace_str = stacktrace_str
should test end have side effect on e passed in?

---

*[mobly/records.py, line 284 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3oVzGwyFl-d3k9t2Q:-Kr3oVzGwyFl-d3k9t2R:bxfio11) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L284)):*
<details><summary><i alt="summary">Quoted 12 lines of code…</i></summary>
>         # This check cannot be based on try...except, which messes up
>         # `exc_info`.
>         if e and hasattr(e, '__traceback__'):
>             exc_traceback = e.__traceback__
>         else:
>             # In py2, exception objects don't have built-in traceback, so we
>             # have to immediately retrieve stacktrace from `sys.exc_info`.
>             _, _, exc_traceback = sys.exc_info()
>         if exc_traceback:
>             stacktrace_str = ''.join(traceback.format_tb(exc_traceback))
>             if not hasattr(e, 'stacktrace_str'):
>                 e.stacktrace_str = stacktrace_str
</details>
This look like a candidate for refactor.  Exact code above

---

*[mobly/records.py, line 322 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3ptASE2tC2CxyjWH7:-Kr3ptASE2tC2CxyjWH8:b-4oha6q) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L322)):*
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>
>             try:
>                 exception_dict[TestResultEnums.RECORD_EXTRAS] = e.extras
>             except AttributeError:
>                 pass
</details>
can hasattr() be used here instead of relying on exception being raised?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-Kr8VMvx7t5gc8BpMUR2:b-xqgz0k) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*

Just giving my 2 cents here, my recommendation is to follow how pytest handles since that'll probably be the user's expectation of mobly.  From a little test of mine, looks like a test failure (with teardown error) results in a reporting of  "FAILED (failures=1, errors=1)".  Can we do similar?  Also, it provided stack trace for the failure in test and the error in teardown.  From an offline chat, sounds like to do so might be a big task, what you guys think about the path forward to just continue however mobly is reporting currently today so as to minimize user impacted in the meantime.  Unless that was partly the reason for this PR and I'm assuming users are either okay with how Mobly is reporting or has created some workaround.
</blockquote></details>

The line "FAILED (failures=1, errors=1)" is the test run summary line of all tests in a run, we have a similar summary line at the end, and we are not really discussing this line here...

We are talking about the status of a single test record.

We are not a unit test framework, so we can't copy pyunit as it is. This is one of the differ points between unit tests and E2E tests.

If that's what you prefer, I'm ok with keeping the current behavior because there have been no complaints from users who run tests everyday.

---

*[mobly/records.py, line 200 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3jiRas15KiEg1ApFP:-Kr8WTTciz-tcCwsDzco:brza0n9) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L200)):*

> e is None
e should be None at this point already
</blockquote></details>

why? e is passed in, it is not always None.

---

*[mobly/records.py, line 203 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3kD9eWG2D5rUDEOU-:-Kr8XCyLsylMTQ0wha9y:bjxq2oh) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L203)):*
It actually can be...
```
>>> x = OrderedDict()
>>> x[None] = 'a'
>>> x
OrderedDict([(None, 'a')])
```

---

*[mobly/records.py, line 205 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3lPLRAOw3SbbSD8V7:-Kr8XHqZm-qbRsSkiSX0:baioqxy) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L205)):*

>         if isinstance(e, signals.TestSignal):
>             self.details = failure_location + str(e.details)
>  
>  
>  
>             self.extras = e.extras
>         elif isinstance(e, Exception):
>             self.details = failure_location + str(e)
Seems like if the logic enters into either of these, failure_location will be empty string.  Is that intended?
</blockquote></details>

Yes. E.g. there's no `failure_location` for the self.termination_signal.

---

*[mobly/records.py, line 223 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3nm3ORrhOCtOJM6Km:-Kr8XkXmbMJv_MgYYAZ_:b-h3ul1p) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L223)):*

>             if e and not hasattr(e, 'stacktrace_str'):
>                 e.stacktrace_str = stacktrace_str
should test end have side effect on e passed in?
</blockquote></details>

this e is used internally for recording test execution info, so it's kinda ok...
the thing is, if we don't save the stacktrace right here, we lose it.
it's sad but it's python...

---

*[mobly/records.py, line 284 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3oVzGwyFl-d3k9t2Q:-Kr8YNUnQM53SkOIRPXC:bs0rvn9) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L284)):*

>         # This check cannot be based on try...except, which messes up
>         # `exc_info`.
>         if e and hasattr(e, '__traceback__'):
>             exc_traceback = e.__traceback__
>         else:
>             # In py2, exception objects don't have built-in traceback, so we
>             # have to immediately retrieve stacktrace from `sys.exc_info`.
>             _, _, exc_traceback = sys.exc_info()
>         if exc_traceback:
>             stacktrace_str = ''.join(traceback.format_tb(exc_traceback))
>             if not hasattr(e, 'stacktrace_str'):
>                 e.stacktrace_str = stacktrace_str
This look like a candidate for refactor.  Exact code above
</blockquote></details>

not sure what you mean. can you clarify.
also it's really hard to read this post due to formatting...

---

*[mobly/records.py, line 322 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3ptASE2tC2CxyjWH7:-Kr8Y4l2nK2TlcupyFHX:b1undhl) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L322)):*

>             try:
>                 exception_dict[TestResultEnums.RECORD_EXTRAS] = e.extras
>             except AttributeError:
>                 pass
can hasattr() be used here instead of relying on exception being raised?
</blockquote></details>

Don't have a strong preference.

i thought you liked pythonic style?
"it's better to ask for forgiveness than permission" :P

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

k2fong:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[mobly/records.py, line 200 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3jiRas15KiEg1ApFP:-Kr9NlBZLx3BU5371B2y:b1r6lk5) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L200)):*

why? e is passed in, it is not always None.
</blockquote></details>

because of the 'if e' on above

---

*[mobly/records.py, line 203 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3kD9eWG2D5rUDEOU-:-Kr9To43i_0oHdx3dDI3:bg7zopp) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L203)):*

It actually can be...
```
>>> x = OrderedDict()
>>> x[None] = 'a'
>>> x
OrderedDict([(None, 'a')])
```
</blockquote></details>

can it be like below?  just trying to reduce the number of conditional logics here.

elif e is None and self.extra_errors:
  failure_location, e = self.extra_errors.popitem(last=False)
  failure_location += ': '

---

*[mobly/records.py, line 205 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3lPLRAOw3SbbSD8V7:-Kr9Osv8BYHFtQlRGpHY:buvaev8) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L205)):*

Yes. E.g. there's no `failure_location` for the self.termination_signal.
</blockquote></details>

if that's the case then I'll suggest not using failure_location here at all

---

*[mobly/records.py, line 223 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3nm3ORrhOCtOJM6Km:-Kr9PNuYPYR77a1qfcyD:b-tg73hw) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L223)):*

this e is used internally for recording test execution info, so it's kinda ok...
the thing is, if we don't save the stacktrace right here, we lose it.
it's sad but it's python...
</blockquote></details>

if so, might be worthwhile to indicate so in the header

---

*[mobly/records.py, line 284 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3oVzGwyFl-d3k9t2Q:-Kr9PbW-PVzlWsD5ShSQ:bsgxs7k) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L284)):*

not sure what you mean. can you clarify.
also it's really hard to read this post due to formatting...
</blockquote></details>

sorry about that, I was expecting it'll highlight code similar to critique :p my comment is that line 219-233 and 288-300 are pretty much identical beside 2 lines that I'm not sure if they should be different either.  can it be refactored to be combine into some method?

---

*[mobly/records.py, line 322 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3ptASE2tC2CxyjWH7:-Kr9S1FKj2qHXgu0nPcN:b-qna566) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L322)):*

Don't have a strong preference.

i thought you liked pythonic style?
"it's better to ask for forgiveness than permission" :P
</blockquote></details>

hmm not me.  but is this style really the pythonic style?  I prefer hasattr() especially since its use else where in this file as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[mobly/records.py, line 200 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3jiRas15KiEg1ApFP:-KrCGa4Aefw6mqIKkqST:bjgacef) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L200)):*

because of the 'if e' on above
</blockquote></details>

Ok, but I fail to see the point here...
If e is None, we want to assign it a not None value from extra errors.

---

*[mobly/records.py, line 203 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3kD9eWG2D5rUDEOU-:-KrCGo05U8oXQTCWbjPt:bnoc1ps) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L203)):*

can it be like below?  just trying to reduce the number of conditional logics here.

elif e is None and self.extra_errors:
  failure_location, e = self.extra_errors.popitem(last=False)
  failure_location += ': '
</blockquote></details>

No, because failure_location could have be overwritten to None by the pop and the `+=` would fail with type error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[mobly/records.py, line 205 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3lPLRAOw3SbbSD8V7:-KrCGyyoTtcsammxe0LQ:bxu03q6) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L205)):*

if that's the case then I'll suggest not using failure_location here at all
</blockquote></details>

But we do need to prepend it here if failure_location is not empty string.
This is just a shortcut to reduce `if` statements.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[mobly/records.py, line 284 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3oVzGwyFl-d3k9t2Q:-KrCHAAK5pYCavnPI-bj:bc4us8f) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L284)):*

sorry about that, I was expecting it'll highlight code similar to critique :p my comment is that line 219-233 and 288-300 are pretty much identical beside 2 lines that I'm not sure if they should be different either.  can it be refactored to be combine into some method?
</blockquote></details>

No, because we are taking stacktraces here so another func stack would create noise in the stacktrace

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[mobly/records.py, line 322 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3ptASE2tC2CxyjWH7:-KrCHYXuXXUVNDbXHZd6:b-5caqun) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L322)):*

hmm not me.  but is this style really the pythonic style?  I prefer hasattr() especially since its use else where in this file as well.
</blockquote></details>

Yes. It is quite common and somewhat recommended to use try...except in place of `if` in Python.
https://stackoverflow.com/questions/7604636/better-to-try-something-and-catch-the-exception-or-test-if-its-possible-first

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[mobly/records.py, line 197 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3inUC5cHYesCg3hIN:-KrCqIigbI--dMXyrhd3:b3vi6kr) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L197)):*

>         if e:
>             self.termination_signal = e
no need for this 'if e' clause.  Just assign?
</blockquote></details>

Will refactor

---

*[mobly/records.py, line 200 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3jiRas15KiEg1ApFP:-KrCq90ebf4aJ56LV1cE:b9ty8wm) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L200)):*

Ok, but I fail to see the point here...
If e is None, we want to assign it a not None value from extra errors.
</blockquote></details>

Oh, i see.
`if e` doesn't gurantee `e is not None`
e can be '' or other values that evaluate to false.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
Done with the fixes.
Here's an example of a grand failure where test body, `teardown_test`, and `on_fail` all failed.

The entry looks like:
```
--- 
"Begin Time": 1502406537536
Details: "fail in body."
"End Time": 1502406537537
"Extra Errors": 
  - 
    Details: "teardown test error."
    Extras: null
    Position: teardown_test
    Stacktrace: |
          File "/Users/angli/Developer/mobly/mobly/base_test.py", line 341, in exec_one_test
            self._teardown_test(test_name)
          File "/Users/angli/Developer/mobly/mobly/base_test.py", line 205, in _teardown_test
            self.teardown_test()
          File "tests/lib/integration_test.py", line 37, in teardown_test
            raise Exception('teardown test error.')
        Exception: teardown test error.
  - 
    Details: "on_fail error."
    Extras: null
    Position: _on_fail
    Stacktrace: |
          File "/Users/angli/Developer/mobly/mobly/base_test.py", line 301, in _exec_procedure_func
            func(tr_record)
          File "/Users/angli/Developer/mobly/mobly/base_test.py", line 226, in _on_fail
            self.on_fail(test_name, begin_time)
          File "tests/lib/integration_test.py", line 40, in on_fail
            raise Exception('on_fail error.')
        Exception: on_fail error.
Extras: null
Result: FAIL
Stacktrace: |
      File "/Users/angli/Developer/mobly/mobly/base_test.py", line 332, in exec_one_test
        test_method()
      File "tests/lib/integration_test.py", line 33, in test_hello_world
        asserts.fail('fail in body.')
      File "/Users/angli/Developer/mobly/mobly/asserts.py", line 241, in fail
        raise signals.TestFailure(msg, extras)
    TestFailure: Details=fail in body., Extras=None
"Test Class": IntegrationTest
"Test Name": test_hello_world
Type: Record
UID: null
```

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 9 unresolved discussions.

---

*[mobly/records.py, line 180 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTiCM4ZWPPVnwvXdjF:-KrDR-RS0xcanNO7EY7o:b-896fix) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L180)):*

The line "FAILED (failures=1, errors=1)" is the test run summary line of all tests in a run, we have a similar summary line at the end, and we are not really discussing this line here...

We are talking about the status of a single test record.

We are not a unit test framework, so we can't copy pyunit as it is. This is one of the differ points between unit tests and E2E tests.

If that's what you prefer, I'm ok with keeping the current behavior because there have been no complaints from users who run tests everyday.
</blockquote></details>

Done.

---

*[mobly/records.py, line 197 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3inUC5cHYesCg3hIN:-KrDR-lwuqUrfUCxzh88:b-896fix) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L197)):*

Will refactor
</blockquote></details>

Done.

---

*[mobly/records.py, line 200 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3jiRas15KiEg1ApFP:-KrDR0LNJcYqMPQj2_AW:b-896fix) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L200)):*

Oh, i see.
`if e` doesn't gurantee `e is not None`
e can be '' or other values that evaluate to false.
</blockquote></details>

Done.

---

*[mobly/records.py, line 203 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3kD9eWG2D5rUDEOU-:-KrDR2JLvD3aQz1wBbuv:b-896fix) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L203)):*

No, because failure_location could have be overwritten to None by the pop and the `+=` would fail with type error.
</blockquote></details>

Done.

---

*[mobly/records.py, line 205 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3lPLRAOw3SbbSD8V7:-KrDR2xwa7grak5QWfX1:b-ruqkw2) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L205)):*

But we do need to prepend it here if failure_location is not empty string.
This is just a shortcut to reduce `if` statements.
</blockquote></details>

Obsolete

---

*[mobly/records.py, line 223 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3nm3ORrhOCtOJM6Km:-KrDR4rUDgRX4uXbHfhK:b-ruqkw2) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L223)):*

if so, might be worthwhile to indicate so in the header
</blockquote></details>

Obsolete

---

*[mobly/records.py, line 284 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3oVzGwyFl-d3k9t2Q:-KrDR5PkwfBzTeyiP04y:b-ruqkw2) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L284)):*

No, because we are taking stacktraces here so another func stack would create noise in the stacktrace
</blockquote></details>

Obsolete

---

*[mobly/records.py, line 322 at r3](https://reviewable.io:443/reviews/google/mobly/284#-Kr3ptASE2tC2CxyjWH7:-KrDR7V21Wjo1AyvM3sm:b-896fix) ([raw file](https://github.com/google/mobly/blob/d3fd8b09c1d6875e09c2aba36aea69c2b15aa67d/mobly/records.py#L322)):*

Yes. It is quite common and somewhat recommended to use try...except in place of `if` in Python.
https://stackoverflow.com/questions/7604636/better-to-try-something-and-catch-the-exception-or-test-if-its-possible-first
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
Rebased

k2fong:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[mobly/records.py, line 260 at r8](https://reviewable.io:443/reviews/google/mobly/284#-KrDywQj97Vemgr_3o27:-KrDywQj97Vemgr_3o28:b-ce0f4g) ([raw file](https://github.com/google/mobly/blob/be96f35805b0015c8d1ae1456be4c4c133b7e167/mobly/records.py#L260)):*
> ```Python
>             if self.result in (TestResultEnums.TEST_RESULT_PASS,
>                                TestResultEnums.TEST_RESULT_SKIP):
>                 self.result = TestResultEnums.TEST_RESULT_ERROR
> ```

Nit. if pass, skip, error, and fail are all the possible result value, will it read better if its if self.result != TestResultEnums.TEST_RESULT_FAIL?  I think there might be one more similar down below.

---

*[mobly/records.py, line 268 at r8](https://reviewable.io:443/reviews/google/mobly/284#-KrE-3-Q6Zo1E-oKKmX6:-KrE-3-Q6Zo1E-oKKmX7:b-r67v17) ([raw file](https://github.com/google/mobly/blob/be96f35805b0015c8d1ae1456be4c4c133b7e167/mobly/records.py#L268)):*
> ```Python
>             self.details = self.termination_signal.details
>             self.stacktrace = self.termination_signal.stacktrace
>             self.extras = self.termination_signal.extras
> ```

another nit.  instead of keeping more states within this class, how about defining properties for details, stacktrace, and extras and just return the results from termination_signal?

---

*[mobly/records.py, line 326 at r8](https://reviewable.io:443/reviews/google/mobly/284#-KrEUJVSrNSl4QdD03V0:-KrEUJVSrNSl4QdD03V1:bt2g9lo) ([raw file](https://github.com/google/mobly/blob/be96f35805b0015c8d1ae1456be4c4c133b7e167/mobly/records.py#L326)):*
> ```Python
>                            TestResultEnums.TEST_RESULT_SKIP):
>             self.result = TestResultEnums.TEST_RESULT_ERROR
>         self.extra_errors[position] = ExceptionRecord(e, position=position)
> ```

I don't believe its used that way but if in case the same position is passed in, the extra error will be overridden for that position.  is that intended?  if that is the intend, how about call this set_error()?  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[mobly/records.py, line 260 at r8](https://reviewable.io:443/reviews/google/mobly/284#-KrDywQj97Vemgr_3o27:-KrEs1w5TtPCylkxPY6e:b-896fix) ([raw file](https://github.com/google/mobly/blob/be96f35805b0015c8d1ae1456be4c4c133b7e167/mobly/records.py#L260)):*

Nit. if pass, skip, error, and fail are all the possible result value, will it read better if its if self.result != TestResultEnums.TEST_RESULT_FAIL?  I think there might be one more similar down below.
</blockquote></details>

Done.

---

*[mobly/records.py, line 268 at r8](https://reviewable.io:443/reviews/google/mobly/284#-KrE-3-Q6Zo1E-oKKmX6:-KrEtbz8tzWeG6-LZ139:b-896fix) ([raw file](https://github.com/google/mobly/blob/be96f35805b0015c8d1ae1456be4c4c133b7e167/mobly/records.py#L268)):*

another nit.  instead of keeping more states within this class, how about defining properties for details, stacktrace, and extras and just return the results from termination_signal?
</blockquote></details>

Done.

---

*[mobly/records.py, line 326 at r8](https://reviewable.io:443/reviews/google/mobly/284#-KrEUJVSrNSl4QdD03V0:-KrEuE6uo4SkpLsupxmH:boji8w4) ([raw file](https://github.com/google/mobly/blob/be96f35805b0015c8d1ae1456be4c4c133b7e167/mobly/records.py#L326)):*

I don't believe its used that way but if in case the same position is passed in, the extra error will be overridden for that position.  is that intended?  if that is the intend, how about call this set_error()?  
</blockquote></details>

Added a check.
This is not a huge concern as only the framework itself sets the position.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

dthkao:
Review status: 0 of 4 files reviewed at latest revision, 6 unresolved discussions.

---

*[mobly/records.py, line 132 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTgS0N6lQ8TLdFtNJ5:-KrFCajBpfuFPCe6Dw20:b46bouh) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L132)):*

Discussed offline, filed https://github.com/google/mobly/issues/300. Acknowledged for now
</blockquote></details>

indentation pr is in so bringing this back up.

---

*[mobly/records.py, line 250 at r9](https://reviewable.io:443/reviews/google/mobly/284#-KrFDkoRafxNWkrjewJw:-KrFDkoRafxNWkrjewJx:b-3csec2) ([raw file](https://github.com/google/mobly/blob/cda0a0c386c5a562c6a7c513748620304480760a/mobly/records.py#L250)):*
> ```Python
>             result: One of the TEST_RESULT enums in TestResultEnums.
>             e: A test termination signal (usually an exception object). It can
>                be any exception instance or of any subclass of
> ```

indent?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[mobly/records.py, line 132 at r1](https://reviewable.io:443/reviews/google/mobly/284#-KqTgS0N6lQ8TLdFtNJ5:-KrFFqqLFyTZgRBmidyB:b-896fix) ([raw file](https://github.com/google/mobly/blob/b8b32f7a4ec1519a61a320a6b3b2c5a1811fb48d/mobly/records.py#L132)):*

indentation pr is in so bringing this back up.
</blockquote></details>

Done.

---

*[mobly/records.py, line 250 at r9](https://reviewable.io:443/reviews/google/mobly/284#-KrFDkoRafxNWkrjewJw:-KrFENweh8wyOPOsTBtJ:b-896fix) ([raw file](https://github.com/google/mobly/blob/cda0a0c386c5a562c6a7c513748620304480760a/mobly/records.py#L250)):*

indent?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

k2fong:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284#-:-KrHxDuGb-hjwujWNkXS:bnfp4nl)*

xpconanfan:
Reviewed 1 of 3 files at r8, 1 of 1 files at r10.
Review status: 2 of 4 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/284)*

PR #290, opened 2017-08-04
Refactor standing subprocess utils.

### LINKED ISSUE #288: `utils._assert_subprocess_running` throws `ValueError`
If the process died, we try to call `p.communicate` to get info from the process.
Then `communicate` would sometimes fail with ValueError saying it's trying to do I/O operation on a closed file.

xpconanfan:
* Remove _assert_subprocess_running checks from util functions.
  - Users can use psutil.Process(pid) to do this check if needed.
* Better error msg for `stop_standing_subprocess`.

Fixes #288

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/290)

### DIFF
--- a/mobly/utils.py
+++ b/mobly/utils.py
@@ -339,6 +339,9 @@ def stop_standing_subprocess(proc, kill_signal=signal.SIGTERM):
         try:
             child.kill()
             child.wait(timeout=10)
+        except psutil.NoSuchProcess:
+            # Ignore if the child process has already terminated.
+            pass
         except:
             failed.append(child.pid)
             logging.exception('Failed to kill standing subprocess %d',
@@ -346,8 +349,11 @@ def stop_standing_subprocess(proc, kill_signal=signal.SIGTERM):
     try:
         process.kill()
         process.wait(timeout=10)
+    except psutil.NoSuchProcess:
+        # Ignore if the process has already terminated.
+        pass
     except:
-        failed.append(process.pid)
+        failed.append(pid)
         logging.exception('Failed to kill standing subprocess %d', pid)
     if failed:
         raise Error('Failed to kill standing subprocesses: %s' % failed)
--- a/mobly/utils.py
+++ b/mobly/utils.py
@@ -273,29 +273,13 @@ def concurrent_exec(func, param_list):
             try:
                 return_vals.append(future.result())
             except Exception as exc:
-                print("{} generated an exception: {}".format(
+                logging.exception("{} generated an exception: {}".format(
                     params, traceback.format_exc()))
                 return_vals.append(exc)
         return return_vals
 
 
-def _assert_subprocess_running(proc):
-    """Checks if a subprocess has terminated on its own.
-
-    Args:
-        proc: A subprocess returned by subprocess.Popen.
-
-    Raises:
-        Error is raised if the subprocess has stopped.
-    """
-    ret = proc.poll()
-    if ret is not None:
-        out, err = proc.communicate()
-        raise Error("Process %d has terminated. ret: %d, stderr: %s,"
-                    " stdout: %s" % (proc.pid, ret, err, out))
-
-
-def start_standing_subprocess(cmd, check_health_delay=0, shell=False):
+def start_standing_subprocess(cmd, shell=False):
     """Starts a long-running subprocess.
 
     This is not a blocking call and the subprocess started by it should be
@@ -304,36 +288,27 @@ def start_standing_subprocess(cmd, check_health_delay=0, shell=False):
     For short-running commands, you should use subprocess.check_call, which
     blocks.
 
-    You can specify a health check after the subprocess is started to make sure
-    it did not stop prematurely.
-
     Args:
         cmd: string, the command to start the subprocess with.
-        check_health_delay: float, the number of seconds to wait after the
-                            subprocess starts to check its health. Default is 0,
-                            which means no check.
         shell: bool, True to run this command through the system shell,
             False to invoke it directly. See subprocess.Proc() docs.
 
     Returns:
         The subprocess that was started.
     """
-    logging.debug('Start standing subprocess with cmd: %s', cmd)
+    logging.debug('Starting standing subprocess with: %s', cmd)
     proc = subprocess.Popen(
         cmd,
         stdin=subprocess.PIPE,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
         shell=shell)
-    logging.debug('Start standing subprocess with cmd: %s', cmd)
     # Leaving stdin open causes problems for input, e.g. breaking the
     # code.inspect() shell (http://stackoverflow.com/a/25512460/1612937), so
     # explicitly close it assuming it is not needed for standing subprocesses.
     proc.stdin.close()
     proc.stdin = None
-    if check_health_delay > 0:
-        time.sleep(check_health_delay)
-        _assert_subprocess_running(proc)
+    logging.debug('Started standing subprocess %d', proc.pid)
     return proc
 
 
@@ -352,10 +327,9 @@ def stop_standing_subprocess(proc, kill_signal=signal.SIGTERM):
         Error: if the subprocess could not be stopped.
     """
     pid = proc.pid
-    logging.debug('Stop standing subprocess %d', pid)
-    _assert_subprocess_running(proc)
+    logging.debug('Stopping standing subprocess %d', pid)
     process = psutil.Process(pid)
-    success = True
+    failed = []
     try:
         children = process.children(recursive=True)
     except AttributeError:
@@ -366,17 +340,18 @@ def stop_standing_subprocess(proc, kill_signal=signal.SIGTERM):
             child.kill()
             child.wait(timeout=10)
         except:
-            success = False
+            failed.append(child.pid)
             logging.exception('Failed to kill standing subprocess %d',
                               child.pid)
     try:
         process.kill()
         process.wait(timeout=10)
     except:
-        success = False
+        failed.append(process.pid)
         logging.exception('Failed to kill standing subprocess %d', pid)
-    if not success:
-        raise Error('Some standing subprocess failed to die')
+    if failed:
+        raise Error('Failed to kill standing subprocesses: %s' % failed)
+    logging.debug('Stopped standing subprocess %d', pid)
 
 
 def wait_for_standing_subprocess(proc, timeout=None):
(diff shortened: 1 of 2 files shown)


k2fong:

---

*[mobly/utils.py, line 340 at r1](https://reviewable.io:443/reviews/google/mobly/290#-KqjcavYl_voCDSbFGsO:-KqjcavYl_voCDSbFGsP:b-xsm2av) ([raw file](https://github.com/google/mobly/blob/1314e49cb229441d918a9ba28b928f110ea3eb32/mobly/utils.py#L340)):*
> child.kill()
child process can potentially be gone/completed at this point which might result in a NoSuchProcess exception that will cause incorrect outcome of this method.

---

*[mobly/utils.py, line 350 at r1](https://reviewable.io:443/reviews/google/mobly/290#-KqjbFidrCN29lUHQB7h:-KqjbFidrCN29lUHQB7i:b-8zix3z) ([raw file](https://github.com/google/mobly/blob/1314e49cb229441d918a9ba28b928f110ea3eb32/mobly/utils.py#L350)):*
> process.pid
just pid to keep consistency?

---

*[tests/mobly/utils_test.py, line 35 at r1](https://reviewable.io:443/reviews/google/mobly/290#-KqjZyqJj6RgKfcRn7hR:-KqjZyqJj6RgKfcRn7hS:b-gvpjpn) ([raw file](https://github.com/google/mobly/blob/1314e49cb229441d918a9ba28b928f110ea3eb32/tests/mobly/utils_test.py#L35)):*
> ```Python
>         p = utils.start_standing_subprocess(['sleep', '1'])
>         p1 = psutil.Process(p.pid)
>         self.assertTrue(p1.is_running())
> ```

Since this is an unit test, does it make sense to verify that popen() was called accordingly rather than checking that the process indeed got created.  similar below for stopping.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/290)*

xpconanfan:

---

*[mobly/utils.py, line 340 at r1](https://reviewable.io:443/reviews/google/mobly/290#-KqjcavYl_voCDSbFGsO:-KqjizBgkNcoTQdWiRlY:b-896fix) ([raw file](https://github.com/google/mobly/blob/1314e49cb229441d918a9ba28b928f110ea3eb32/mobly/utils.py#L340)):*

> child.kill()
child process can potentially be gone/completed at this point which might result in a NoSuchProcess exception that will cause incorrect outcome of this method.
</blockquote></details>

Done.

---

*[mobly/utils.py, line 350 at r1](https://reviewable.io:443/reviews/google/mobly/290#-KqjbFidrCN29lUHQB7h:-Kqjj1WqPrsDZVh2_BiI:b-896fix) ([raw file](https://github.com/google/mobly/blob/1314e49cb229441d918a9ba28b928f110ea3eb32/mobly/utils.py#L350)):*

> process.pid
just pid to keep consistency?
</blockquote></details>

Done.

---

*[tests/mobly/utils_test.py, line 35 at r1](https://reviewable.io:443/reviews/google/mobly/290#-KqjZyqJj6RgKfcRn7hR:-Kqjj4Xd76jJveQ_kEZe:b-4hxxe) ([raw file](https://github.com/google/mobly/blob/1314e49cb229441d918a9ba28b928f110ea3eb32/tests/mobly/utils_test.py#L35)):*

Since this is an unit test, does it make sense to verify that popen() was called accordingly rather than checking that the process indeed got created.  similar below for stopping.
</blockquote></details>

Hmm, that seems to run too close to the implementation details.
I think the check level these tests touch today is good.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/290)*

k2fong:

---

*[tests/mobly/utils_test.py, line 35 at r1](https://reviewable.io:443/reviews/google/mobly/290#-KqjZyqJj6RgKfcRn7hR:-KqjrTcy3q93wHV3aozb:b-so59jr) ([raw file](https://github.com/google/mobly/blob/1314e49cb229441d918a9ba28b928f110ea3eb32/tests/mobly/utils_test.py#L35)):*

Hmm, that seems to run too close to the implementation details.
I think the check level these tests touch today is good.
</blockquote></details>

Depending on the underlying function popen() to operate properly might risk potential flakiness for this test.  Yes, it is testing to the implementation details here but I believe it make sense here.  In addition, mocking popen make sense also since then cmd arguments & shell can be verified as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/290)*

xpconanfan:

---

*[tests/mobly/utils_test.py, line 35 at r1](https://reviewable.io:443/reviews/google/mobly/290#-KqjZyqJj6RgKfcRn7hR:-KqjyBTx--j3dP4ECxw3:bcfsirr) ([raw file](https://github.com/google/mobly/blob/1314e49cb229441d918a9ba28b928f110ea3eb32/tests/mobly/utils_test.py#L35)):*

Depending on the underlying function popen() to operate properly might risk potential flakiness for this test.  Yes, it is testing to the implementation details here but I believe it make sense here.  In addition, mocking popen make sense also since then cmd arguments & shell can be verified as well.
</blockquote></details>

Not sure what you mean by flakiness though.
We've been using this test for a while and never had issue with this flakiness, could you elaborate?

From a pure software point of view, yes mocking Popen would be good.
However, from experience, practically, we want to actually call Popen to make sure it behaves the way we use it.
This has proven useful for ensuring the code works on different platforms and different python versions.
So in that sense these are not actually pure unit tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/290)*

k2fong:

---

*[tests/mobly/utils_test.py, line 35 at r1](https://reviewable.io:443/reviews/google/mobly/290#-KqjZyqJj6RgKfcRn7hR:-KqurwV_tnWKNONEs_fF:b-h4b181) ([raw file](https://github.com/google/mobly/blob/1314e49cb229441d918a9ba28b928f110ea3eb32/tests/mobly/utils_test.py#L35)):*

Not sure what you mean by flakiness though.
We've been using this test for a while and never had issue with this flakiness, could you elaborate?

From a pure software point of view, yes mocking Popen would be good.
However, from experience, practically, we want to actually call Popen to make sure it behaves the way we use it.
This has proven useful for ensuring the code works on different platforms and different python versions.
So in that sense these are not actually pure unit tests.
</blockquote></details>

What i mean be flaky is that this test depends on other functions operating properly for this test to pass (in this case Popen(), fork(), Sleep).  Flaky might have been the wrong word, maybe fragile?? but not really neither since most likely popen will work fine unless the system is in a pretty bad state...

Of another note, I don't know if Mobly is meant to covers non Linux/Unix like system or not.  I don't believe platform like Windows will have the sleep command.    This is small, but no reason for this to make Mobly platform dependent.

Yes, if this is an integration test, I'm probably more okay with it :)  I can be good with this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/290)*

dthkao:

---

*[mobly/utils.py, line 311 at r2](https://reviewable.io:443/reviews/google/mobly/290#-Kqw7pA-EwfCHdGzwTHe:-Kqw7pA00hpuiPjVsumP:b-sxywhp) ([raw file](https://github.com/google/mobly/blob/f49ec617b18e41f4ea6f1f55fba2d30445a6dfda/mobly/utils.py#L311)):*
> ```Python
>     proc.stdin.close()
>     proc.stdin = None
>     logging.debug('Started standing subprocess %d', proc.pid)
> ```

After this change, any standing subprocess that fails at start would do so silently. That's a pretty significant change in behavior.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/290)*

dthkao:

---

*[mobly/utils.py, line 311 at r2](https://reviewable.io:443/reviews/google/mobly/290#-Kqw7pA-EwfCHdGzwTHe:-Kqw9bY4zTHBOwiXZMGP:b-srw74s) ([raw file](https://github.com/google/mobly/blob/f49ec617b18e41f4ea6f1f55fba2d30445a6dfda/mobly/utils.py#L311)):*

After this change, any standing subprocess that fails at start would do so silently. That's a pretty significant change in behavior.
</blockquote></details>

There are at least a few users of check_health_delay.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/290)*

k2fong:

---

*[tests/mobly/utils_test.py, line 35 at r1](https://reviewable.io:443/reviews/google/mobly/290#-KqjZyqJj6RgKfcRn7hR:-KqxrmtykfsHc6e04ARW:bdkfwdr) ([raw file](https://github.com/google/mobly/blob/1314e49cb229441d918a9ba28b928f110ea3eb32/tests/mobly/utils_test.py#L35)):*

What i mean be flaky is that this test depends on other functions operating properly for this test to pass (in this case Popen(), fork(), Sleep).  Flaky might have been the wrong word, maybe fragile?? but not really neither since most likely popen will work fine unless the system is in a pretty bad state...

Of another note, I don't know if Mobly is meant to covers non Linux/Unix like system or not.  I don't believe platform like Windows will have the sleep command.    This is small, but no reason for this to make Mobly platform dependent.

Yes, if this is an integration test, I'm probably more okay with it :)  I can be good with this.

</blockquote></details>

Per offline chat, if this is an issue, a separate bug will be created

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/290)*

xpconanfan:

---

*[mobly/utils.py, line 311 at r2](https://reviewable.io:443/reviews/google/mobly/290#-Kqw7pA-EwfCHdGzwTHe:-Kqxvm8RDd4HuYopyxVh:b-r0prop) ([raw file](https://github.com/google/mobly/blob/f49ec617b18e41f4ea6f1f55fba2d30445a6dfda/mobly/utils.py#L311)):*

There are at least a few users of check_health_delay.
</blockquote></details>

Yes, but this does not block anybody as users could still check the liveliness of the process if they choose to, via psutil or whatever method appropriate to the context.
This function's feature was bloated and not reliable.
* The check did not make sense in the first place. If the caller specifies `check_health_delay=1`, and the process dies unexpectedly at 1.5s, then what? It was added a long time ago as a hack.
* Also as peter brought up, `sleep` doesn't even exist on Win, so the unit test shouldn't have worked.

So I'm trimming this function to do exactly what it says - start a standing subprocess.
If the process dies after 10s, it's the test's problem.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/290)*

PR #292, opened 2017-08-04
Modified parse_args and added unit test

maddychan:
Updated pull request with changing parse_args to parse_known args and added unit test

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/292)

### DIFF
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -1,4 +1,4 @@
-# Copyright 2017 Google Inc. 
+# Copyright 2017 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -1,4 +1,4 @@
-# Copyright 2017 Google Inc.
+# Copyright 2017 Google Inc. 
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -70,7 +70,7 @@ def run_suite(test_classes, argv=None):
         help='A list of test classes and optional tests to execute.')
     if not argv:
         argv = sys.argv[1:]
-    args = parser.parse_known_args(argv)
+    args = parser.parse_args(argv)
     # Load test config file.
     test_configs = config_parser.load_test_config_file(args.config[0])
 
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -70,7 +70,7 @@ def run_suite(test_classes, argv=None):
         help='A list of test classes and optional tests to execute.')
     if not argv:
         argv = sys.argv[1:]
-    args = parser.parse_args(argv)
+    args = parser.parse_known_args(argv)
     # Load test config file.
     test_configs = config_parser.load_test_config_file(args.config[0])
 
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -80,7 +80,7 @@ def main(argv=None):
         help='Specify which test beds to run tests on.')
     if not argv:
         argv = sys.argv[1:]
-    args = parser.parse_args(argv)
+    args = parser.parse_known_args(argv)
     # Load test config file.
     test_configs = config_parser.load_test_config_file(args.config[0],
                                                        args.test_bed)
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -80,7 +80,7 @@ def main(argv=None):
         help='Specify which test beds to run tests on.')
     if not argv:
         argv = sys.argv[1:]
-    args = parser.parse_args(argv)
+    args = parser.parse_known_args(argv)[0]
     # Load test config file.
     test_configs = config_parser.load_test_config_file(args.config[0],
                                                        args.test_bed)
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -80,7 +80,7 @@ def main(argv=None):
         help='Specify which test beds to run tests on.')
     if not argv:
         argv = sys.argv[1:]
-    args = parser.parse_known_args(argv)
+    args = parser.parse_args(argv)
     # Load test config file.
     test_configs = config_parser.load_test_config_file(args.config[0],
                                                        args.test_bed)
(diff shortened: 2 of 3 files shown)


xpconanfan:

---

*[tests/mobly/test_runner_test.py, line 338 at r2](https://reviewable.io:443/reviews/google/mobly/292#-Kqj9dsXbIK1eteY3NHT:-Kqj9dsXbIK1eteY3NHU:brdw88s) ([raw file](https://github.com/google/mobly/blob/49d8dae88f8f32b872c20c0f7e66e6985de2cbd2/tests/mobly/test_runner_test.py#L338)):*
> ```Python
>         'mobly.test_runner.TestRunner',
>         return_value=mock.MagicMock())
>     def test_main(self, mock_test_runner, mock_config, mock_find_test):
> ```

need a better name here.
you're testing that arbitrary args can be passed in, so the name needs to reflect that.

---

*[tests/mobly/test_runner_test.py, line 340 at r2](https://reviewable.io:443/reviews/google/mobly/292#-Kqj9kkPVLxtNOSWhDDn:-Kqj9kkPVLxtNOSWhDDo:bje455h) ([raw file](https://github.com/google/mobly/blob/49d8dae88f8f32b872c20c0f7e66e6985de2cbd2/tests/mobly/test_runner_test.py#L340)):*
> ```Python
>     def test_main(self, mock_test_runner, mock_config, mock_find_test):
>         try:
>             test_runner.main(['-c', "test_config", '-b', 'hello'])
> ```

you're not verifying that the value of -c and -b are actually correct.

A test needs to have explicit conditions that it checks, and this test doesn't seem to check anything.

---

*[tests/mobly/test_runner_test.py, line 341 at r2](https://reviewable.io:443/reviews/google/mobly/292#-Kqj9rqeKq8AE10o4sfV:-Kqj9rqeKq8AE10o4sfW:b7lrbd5) ([raw file](https://github.com/google/mobly/blob/49d8dae88f8f32b872c20c0f7e66e6985de2cbd2/tests/mobly/test_runner_test.py#L341)):*
> ```Python
>         try:
>             test_runner.main(['-c', "test_config", '-b', 'hello'])
>         except:
> ```

this is a very broad except that doesn't provide much value to the test.
In a test, if you don't know what you're catching, don't catch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/292)*

maddychan:
Thanks for the feedback! I'll get started on it, but I need more clarification on the second comment. Here I'm testing whether the method will ignore -b "hello" because it's not an argument it recognizes, and I don't think there's a 'correct' input for -c? All the functions that would interact with its contents have been mocked out, and the correctness of that specific parameter isn't what I'm testing. Should I check the normal case (just -c 'test_config") as well as the abnormal case (-b 'hello') instead to make it more clear?

maddychan:
Changing the parse_known_args function in test_runner.main back to parse_args would fail the test that I wrote, and causes a SystemExit to happen. I can't specify that in the try/except, so should I just not wrap it with the try/except? I write it in to provide more information about the failure.

xpconanfan:
1. Shouldn't the test make sure the parsed value of `-c` is the same as the input value?
2. If you expect no raise, then just let the error through and not catch anything.

xpconanfan:

---

*[tests/mobly/test_runner_test.py, line 339 at r3](https://reviewable.io:443/reviews/google/mobly/292#-KqjgWXoKGDPBS8u9GL1:-KqjgWXoKGDPBS8u9GL2:b-n3qom9) ([raw file](https://github.com/google/mobly/blob/2cb2d7c50ce37678282596f892160b8724f0d973/tests/mobly/test_runner_test.py#L339)):*
> ```Python
>         return_value=mock.MagicMock())
>     def test_main_parse_args(self, mock_test_runner, mock_config, mock_find_test):
>         test_runner.main(['-c', "test_config", '-b', 'hello'])
> ```

1. use single quotes to be consistent.
2. 'test_config' is a bit confusing for the value of `-c`, let's use something more realistic like 'some/path/foo.yaml'
3. Can this test also check that the arg not consumed by mobly is properly passed in?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/292)*

maddychan:
So I tweaked the test_runner code such that it'll ignore the args not consumed by mobly; I can check that the ignored args are properly passed in but that would require changing the code I changed in test_runner. Is that what you want?

xpconanfan:
How do we use those extra args if they are not used by Mobly?

---

---

*[tests/mobly/test_runner_test.py, line 341 at r4](https://reviewable.io:443/reviews/google/mobly/292#-KqxvCkyyn0AurDuv0_U:-KqxvCkyyn0AurDuv0_V:bgz1hxl) ([raw file](https://github.com/google/mobly/blob/e4da6fe896257b041579bf7f8725b5071fa31094/tests/mobly/test_runner_test.py#L341)):*
> ```Python
>         test_runner.main(['-c', 'some/path/foo.yaml', '-b', 'hello'])
>         mock_config.assert_called_with('some/path/foo.yaml', None)
> ```

An extra empty line here.
Could you run linter `yapf`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/292#-:-KqxvKm644Ul67zF_DeD:b-wywi5h)*

maddychan:
parse_args returns a namespace of the args it recognizes, and parse_known_args returns a tuple with that namespace and a list of the unrecognized args. I just stored the namespace so that it would be returning the same thing as if it were parse_args.

xpconanfan:

---

*[tests/mobly/test_runner_test.py, line 343 at r5](https://reviewable.io:443/reviews/google/mobly/292#-Kqy34dK39Hv_eOJHCsK:-Kqy34dK39Hv_eOJHCsL:bthugf9) ([raw file](https://github.com/google/mobly/blob/13535c75737c83663123d1c53b87041f07ee3317/tests/mobly/test_runner_test.py#L343)):*
> ```Python
> 
> if __name__ == "__main__":
>     unittest.main()
> ```

missing extra empty line at the end.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/292)*

xpconanfan:

---

*[tests/mobly/test_runner_test.py, line 344 at r6](https://reviewable.io:443/reviews/google/mobly/292#-Kqy3lHpf-DDvZig8Gxj:-Kqy3lHpf-DDvZig8Gxk:b-k10qpq) ([raw file](https://github.com/google/mobly/blob/7922dc5ff6ce4dec484c2680b067f50b5800fed2/tests/mobly/test_runner_test.py#L344)):*
> ```Python
> if __name__ == "__main__":
>     unittest.main()
>     
> ```

extra spaces

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/292)*

xpconanfan:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/292#-:-KqyEGH25IkUeznmxhlO:bnfp4nl)*

PR #305, opened 2017-08-11
Fix bad string template in adb proxy.

dthkao:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/305)

### DIFF
--- a/mobly/controllers/android_device_lib/adb.py
+++ b/mobly/controllers/android_device_lib/adb.py
@@ -147,7 +147,7 @@ class AdbProxy(object):
             if self.serial:
                 adb_cmd = '"%s" -s "%s" %s %s' % (ADB, self.serial, name, args)
             else:
-                adb_cmd = '"adb" %s %s' % (ADB, name, args)
+                adb_cmd = '"%s" %s %s' % (ADB, name, args)
         else:
             adb_cmd = [ADB]
             if self.serial:
--- a/tests/mobly/controllers/android_device_lib/adb_test.py
+++ b/tests/mobly/controllers/android_device_lib/adb_test.py
@@ -91,18 +91,24 @@ class AdbTest(unittest.TestCase):
     def test_exec_adb_cmd(self):
         with mock.patch.object(adb.AdbProxy, '_exec_cmd') as mock_exec_cmd:
             adb.AdbProxy().shell(['arg1', 'arg2'])
-            mock_exec_cmd.assert_called_once_with(['adb', 'shell', 'arg1', 'arg2'], shell=False, timeout=None)
+            mock_exec_cmd.assert_called_once_with(
+                ['adb', 'shell', 'arg1', 'arg2'], shell=False, timeout=None)
         with mock.patch.object(adb.AdbProxy, '_exec_cmd') as mock_exec_cmd:
             adb.AdbProxy('12345').shell(['arg1', 'arg2'])
-            mock_exec_cmd.assert_called_once_with(['adb', '-s', '12345', 'shell', 'arg1', 'arg2'], shell=False, timeout=None)
+            mock_exec_cmd.assert_called_once_with(
+                ['adb', '-s', '12345', 'shell', 'arg1', 'arg2'],
+                shell=False,
+                timeout=None)
 
     def test_exec_adb_cmd_with_shell_true(self):
         with mock.patch.object(adb.AdbProxy, '_exec_cmd') as mock_exec_cmd:
             adb.AdbProxy().shell('arg1 arg2', shell=True)
-            mock_exec_cmd.assert_called_once_with('"adb" shell arg1 arg2', shell=True, timeout=None)
+            mock_exec_cmd.assert_called_once_with(
+                '"adb" shell arg1 arg2', shell=True, timeout=None)
         with mock.patch.object(adb.AdbProxy, '_exec_cmd') as mock_exec_cmd:
             adb.AdbProxy('12345').shell('arg1 arg2', shell=True)
-            mock_exec_cmd.assert_called_once_with('"adb" -s "12345" shell arg1 arg2', shell=True, timeout=None)
+            mock_exec_cmd.assert_called_once_with(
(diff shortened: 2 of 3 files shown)


xpconanfan:
:lgtm:

---

---

*[mobly/controllers/android_device_lib/adb.py, line 125 at r1](https://reviewable.io:443/reviews/google/mobly/305#-KrFDosFbUreVJC4fsIR:-KrFDosFbUreVJC4fsIS:b-gzn1my) ([raw file](https://github.com/google/mobly/blob/99cd5d2530802bbe78925ee5b00f63915635ed30/mobly/controllers/android_device_lib/adb.py#L125)):*
> ```Python
>                 adb_cmd = '"%s" -s "%s" %s %s' % (ADB, self.serial, name, args)
>             else:
>                 adb_cmd = '"%s" %s %s' % (ADB, name, args)
> ```

Mind adding a unit test for this path?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/305#-:-KrFDsr-BEtjI-dqoW_G:bnfp4nl)*

xpconanfan:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/305)*

xpconanfan:
Sorry, not sure what I did but reviewable posted another commit...

k2fong:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[mobly/controllers/android_device_lib/adb.py, line 122 at r1](https://reviewable.io:443/reviews/google/mobly/305#-KrbccHOcHCasmCwJvRx:-KrbccHOcHCasmCwJvRy:bkaq1a) ([raw file](https://github.com/google/mobly/blob/1a7da6a24d4e92f1476cf292715ec7b9faed92af/mobly/controllers/android_device_lib/adb.py#L122)):*
> ```Python
>             # Add quotes around "adb" in case the ADB path contains spaces. This
>             # is pretty common on Windows (e.g. Program Files).
>             if self.serial:
> ```

nit, seems like the path for shell and not shell builds the adb command quite differently, will suggest keeping it consistent.  Also, whats the difference between two?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/305)*

dthkao:

---

*[mobly/controllers/android_device_lib/adb.py, line 122 at r1](https://reviewable.io:443/reviews/google/mobly/305#-KrbccHOcHCasmCwJvRx:-Krbns4SdtOELcotztgL:b-tctzkq) ([raw file](https://github.com/google/mobly/blob/1a7da6a24d4e92f1476cf292715ec7b9faed92af/mobly/controllers/android_device_lib/adb.py#L122)):*

nit, seems like the path for shell and not shell builds the adb command quite differently, will suggest keeping it consistent.  Also, whats the difference between two?
</blockquote></details>

This is a pretty good explanation: https://stackoverflow.com/questions/3172470/actual-meaning-of-shell-true-in-subprocess

Basically, shell=True implies using the system shell (whatever that may be) and all args are given as a single string.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/305)*

k2fong:

---

*[mobly/controllers/android_device_lib/adb.py, line 122 at r1](https://reviewable.io:443/reviews/google/mobly/305#-KrbccHOcHCasmCwJvRx:-Krcu_5OWKN-4Nuf_b4-:b9hzmxr) ([raw file](https://github.com/google/mobly/blob/1a7da6a24d4e92f1476cf292715ec7b9faed92af/mobly/controllers/android_device_lib/adb.py#L122)):*

This is a pretty good explanation: https://stackoverflow.com/questions/3172470/actual-meaning-of-shell-true-in-subprocess

Basically, shell=True implies using the system shell (whatever that may be) and all args are given as a single string.
</blockquote></details>

ah, thanks.  Then how about doing something like ' '.join(adb_cmd) so that we can reduce to duplication if in case shell is True?  

Also, as caller, it has to know to pass in string vs list of string depending on shell.   This seems to be where this dual pipeline of constructing adb_cmd stem from.  how about choosing either string or list of string for _exec_adb_cmd() and have _exec_adb_cmd() change it accordingly depending on the value of shell?  Simplify caller from having to know this popen() detail.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/305)*

dthkao:

---

*[mobly/controllers/android_device_lib/adb.py, line 122 at r1](https://reviewable.io:443/reviews/google/mobly/305#-KrbccHOcHCasmCwJvRx:-KrdWbBOPSCKnT-CmIVk:bntfxy0) ([raw file](https://github.com/google/mobly/blob/ae3719efd3806bc3a7cbfb28cc505794d46f3349/mobly/controllers/android_device_lib/adb.py#L122)):*
>Then how about doing something like ' '.join(adb_cmd) so that we can reduce to duplication if in case shell is True?

Then we'd have some confusing logic to rewrite the first (and possibly third) element of the array (the adb path) to add quotes in the string version. It's doable, but IMO is not as clear what is going on.

>Also, as caller, it has to know to pass in string vs list of string depending on shell.

That is common behavior of all subprocess calls. We're trying to keep consistency with that because we do want to expose the shell=True selector that behaves in a similar way.

---

*[mobly/controllers/android_device_lib/adb.py, line 125 at r1](https://reviewable.io:443/reviews/google/mobly/305#-KrFDosFbUreVJC4fsIR:-Kre4XOrfdjvAqHH5ryX:b-896fix) ([raw file](https://github.com/google/mobly/blob/ae3719efd3806bc3a7cbfb28cc505794d46f3349/mobly/controllers/android_device_lib/adb.py#L125)):*

Mind adding a unit test for this path?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/305)*

k2fong:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/305#-:-KrfrQZXBLG3z3KRQGg-:bnfp4nl)*

PR #306, opened 2017-08-11
Properly report errors in `teardown_class`.

xpconanfan:
Errors in `teardown_class` have not been properly recorded until now.

This is the second PR for #264

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/306)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -175,6 +175,19 @@ class BaseTestClass(object):
         Implementation is optional.
         """
 
+    def _teardown_class(self):
+        """Proxy function to guarantee the base implementation of
+        teardown_class is called.
+        """
+        record = records.TestResultRecord('teardown_class', self.TAG)
+        record.test_begin()
+        try:
+            self.teardown_class()
+        except Exception as e:
+            record.test_error(e)
+            record.update_record()
+            self.results.add_class_error(record)
+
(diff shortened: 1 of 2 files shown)


k2fong:
:lgtm:

I wonder if context manager can be used to ensure the setup_() & teardown_() function pair calling can be use to simplify.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/306#-:-Krd2KHKUCYeJhtc_x6D:b5f8xa5)*

xpconanfan:
We don't have problem with one of them not getting called, which is what context manager is useful for.
We are dealing with handling errors from these functions, so I'm not sure how would context manager help here.
Can you elaborate a bit?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/306#-:-KrdCsdda15CfFZeACG9:butvj0f)*

k2fong:
currently, we are ensuring teardown__() type function gets call appropriately within exceptions and finally.  This seems to me like one could easily be missed (especially if for example a new exception is introduced for some reason).  My thoughts on context manager was so it can be used to ensure the setup... and teardown... function gets call regardless if exception or proper flow.  something like the pseudo code below

with class_context():
  for each testcase:
    with testcase_context():
      ..

PR #311, opened 2017-08-16
Fix tallying for FAIL from `setup_test`

### LINKED ISSUE #229: Exceptions in `setup_test` should leave the test in `ERROR` status
Regardless of the type of the exception, `setup_test` error should cause `ERROR` status.

This is different from a test method.
In a test method, an exception based on signals.TestFailure should cause the test to exit with `FAILED` status.

This is to be consistent with pyunit's behavior.

xpconanfan:
If `setup_test` fails with TestFailure, it should be tallied with ERROR status.

Fixes #229 
Related to #264

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/311)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -17,6 +17,7 @@ import copy
 import functools
 import inspect
 import logging
+import sys
 
 from mobly import logger
 from mobly import records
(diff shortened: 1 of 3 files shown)


dthkao:
:lgtm:

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/311#-:-KrfNAQqQFc2zHoQZqRQ:bnfp4nl)*

PR #316, opened 2017-08-17
Fix deepcopy of `ExceptionRecord` objects.

### LINKED ISSUE #315: Improper usage of `deepcopy` in `record`
`copy.deepcopy` does not always work with user-defined types, as described [here](https://stackoverflow.com/questions/10618956/copy-deepcopy-raises-typeerror-on-objects-with-self-defined-new-method).

So if a user-defined object is put into `record`, attempts to `deepcopy` the object would crash.

xpconanfan:
Fixes #315

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/316)

### DIFF
--- a/mobly/records.py
+++ b/mobly/records.py
@@ -186,6 +186,9 @@ class ExceptionRecord(object):
             exception = self.exception
         result = ExceptionRecord(exception, self.position)
         result.stacktrace = self.stacktrace
+        result.details = self.details
+        result.extras = self.extras
+        result.position = self.position
         return result
 
 
--- a/mobly/records.py
+++ b/mobly/records.py
@@ -172,6 +172,22 @@ class ExceptionRecord(object):
         result[TestResultEnums.RECORD_EXTRAS] = copy.deepcopy(self.extras)
         return result
 
+    def __deepcopy__(self, memo):
+        """Overrides deepcopy for the class.
+
+        If the exception object has a constructor that takes extra args, deep
+        copy won't work. So we need to have a custom logic for deepcopy.
+        """
+        try:
+            exception = copy.deepcopy(self.exception)
+        except TypeError:
+            # If the exception object cannot be copied, use the original
+            # exception object.
+            exception = self.exception
+        result = ExceptionRecord(exception, self.position)
+        result.stacktrace = self.stacktrace
+        return result
+
 
 class TestResultRecord(object):
     """A record that holds the information of a single test.
--- a/tests/mobly/records_test.py
+++ b/tests/mobly/records_test.py
@@ -340,7 +340,9 @@ class RecordsTest(unittest.TestCase):
             raise RecordTestError('Oh ha!')
         except RecordTestError as e:
             er = records.ExceptionRecord(e)
-        copy.deepcopy(er)
+        new_er = copy.deepcopy(er)
+        self.assertIsNot(er, new_er)
+        self.assertDictEqual(er.to_dict(), new_er.to_dict())
 
 
 if __name__ == "__main__":
--- a/tests/mobly/records_test.py
+++ b/tests/mobly/records_test.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import copy
 import os
 import shutil
 import tempfile
@@ -23,6 +24,16 @@ from mobly import records
 from mobly import signals
 
 from tests.lib import utils
+
+
+class RecordTestError(Exception):
+    """Error class with constructors that take extra args.
+
+    Used for ExceptionRecord tests.
+    """
+
+    def __init__(self, something):
+        self._something = something
 
 
 class RecordsTest(unittest.TestCase):
@@ -322,6 +333,14 @@ class RecordsTest(unittest.TestCase):
             content = yaml.load(f)
             self.assertEqual(content['Type'],
                              records.TestSummaryEntryType.RECORD.value)
+
+    def test_exception_record_deepcopy(self):
+        """Makes sure ExceptionRecord wrapper handles deep copy properly."""
+        try:
+            raise RecordTestError('Oh ha!')
+        except RecordTestError as e:
+            er = records.ExceptionRecord(e)
+        copy.deepcopy(er)
 
 
 if __name__ == "__main__":


k2fong:

---

*[mobly/records.py, line 188 at r1](https://reviewable.io:443/reviews/google/mobly/316#-KrmDOHDo07e2qIt24Ez:-KrmDOHDo07e2qIt24F-:bfp4fxl) ([raw file](https://github.com/google/mobly/blob/4a26135f5e14cef8329068770572a7b51bb54344/mobly/records.py#L188)):*
> ```Python
>             exception = self.exception
>         result = ExceptionRecord(exception, self.position)
>         result.stacktrace = self.stacktrace
> ```

does the other fields need to be copy also?

---

*[tests/mobly/records_test.py, line 343 at r1](https://reviewable.io:443/reviews/google/mobly/316#-Krm8pVbiFfkBKumwrWE:-Krm8pVcmwa2S8tGrN8i:byn67rj) ([raw file](https://github.com/google/mobly/blob/4a26135f5e14cef8329068770572a7b51bb54344/tests/mobly/records_test.py#L343)):*
> ```Python
>         except RecordTestError as e:
>             er = records.ExceptionRecord(e)
>         copy.deepcopy(er)
> ```

missing assertions?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/316)*

xpconanfan:

---

*[mobly/records.py, line 188 at r1](https://reviewable.io:443/reviews/google/mobly/316#-KrmDOHDo07e2qIt24Ez:-KrmLe0WK6pLvZSpkIY9:btau1av) ([raw file](https://github.com/google/mobly/blob/4a26135f5e14cef8329068770572a7b51bb54344/mobly/records.py#L188)):*

does the other fields need to be copy also?
</blockquote></details>

No, other fields are primitive types.

---

*[tests/mobly/records_test.py, line 343 at r1](https://reviewable.io:443/reviews/google/mobly/316#-Krm8pVbiFfkBKumwrWE:-KrmLkW0p-DysGSvtlSd:b-zeohax) ([raw file](https://github.com/google/mobly/blob/4a26135f5e14cef8329068770572a7b51bb54344/tests/mobly/records_test.py#L343)):*

missing assertions?
</blockquote></details>

the assertion is that this doesn't throw anything

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/316)*

k2fong:

---

*[mobly/records.py, line 185 at r1](https://reviewable.io:443/reviews/google/mobly/316#-KrmU5oFpy8vlAD40Olw:-KrmU5oGP3UdeJx0LQID:bhcve76) ([raw file](https://github.com/google/mobly/blob/4a26135f5e14cef8329068770572a7b51bb54344/mobly/records.py#L185)):*
>             # If the exception object cannot be copied, use the original
>             # exception object.
>             exception = self.exception
Wouldn't this give user the wrong(not deepcopy)/inconsistent(sometime deep, sometime shallow) behavior?

---

*[mobly/records.py, line 188 at r1](https://reviewable.io:443/reviews/google/mobly/316#-KrmDOHDo07e2qIt24Ez:-KrmS5iVVcpP2puIcMEP:bbdi17x) ([raw file](https://github.com/google/mobly/blob/4a26135f5e14cef8329068770572a7b51bb54344/mobly/records.py#L188)):*

No, other fields are primitive types.
</blockquote></details>

but there isn't a shallow copy of them into the result that is return

---

*[tests/mobly/records_test.py, line 343 at r1](https://reviewable.io:443/reviews/google/mobly/316#-Krm8pVbiFfkBKumwrWE:-KrmSTYaA2d4on06xpf-:b-tmunz) ([raw file](https://github.com/google/mobly/blob/4a26135f5e14cef8329068770572a7b51bb54344/tests/mobly/records_test.py#L343)):*

the assertion is that this doesn't throw anything
</blockquote></details>

I guess I was looking for the part that it indeed performed deep copy

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/316)*

xpconanfan:

---

*[mobly/records.py, line 185 at r1](https://reviewable.io:443/reviews/google/mobly/316#-KrmU5oFpy8vlAD40Olw:-KrmWCFp37g3Cf9XbDiZ:b-jswluk) ([raw file](https://github.com/google/mobly/blob/4a26135f5e14cef8329068770572a7b51bb54344/mobly/records.py#L185)):*

>             # If the exception object cannot be copied, use the original
>             # exception object.
>             exception = self.exception
Wouldn't this give user the wrong(not deepcopy)/inconsistent(sometime deep, sometime shallow) behavior?
</blockquote></details>

Only the exception object, which is impossible to deepcopy in such cases.
Otherwise we'd have to not give a copy of exception at all.
Something is better than nothing here.

Also, the intention is to prevent the test from modifying the internal test record.
Even if the test somehow modifies the exception object, there is no impact as the actual exception object is not serialized to summary anyway. So it actually makes little difference.

---

*[mobly/records.py, line 188 at r1](https://reviewable.io:443/reviews/google/mobly/316#-KrmDOHDo07e2qIt24Ez:-KrmWLV064-Kx57XSob2:bgdriw3) ([raw file](https://github.com/google/mobly/blob/4a26135f5e14cef8329068770572a7b51bb54344/mobly/records.py#L188)):*

but there isn't a shallow copy of them into the result that is return
</blockquote></details>

The constructor takes care of that as they are all primitive types.
We only need to preserve stacktrace from the original as the new one would have been inaccurate.

---

*[tests/mobly/records_test.py, line 343 at r1](https://reviewable.io:443/reviews/google/mobly/316#-Krm8pVbiFfkBKumwrWE:-KrmX0NknZKX0lvtecZg:b-896fix) ([raw file](https://github.com/google/mobly/blob/4a26135f5e14cef8329068770572a7b51bb54344/tests/mobly/records_test.py#L343)):*

I guess I was looking for the part that it indeed performed deep copy
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/316)*

k2fong:

---

*[mobly/records.py, line 188 at r1](https://reviewable.io:443/reviews/google/mobly/316#-KrmDOHDo07e2qIt24Ez:-KrmXpZj7jXZZwLf9D5X:b-4do7x5) ([raw file](https://github.com/google/mobly/blob/4a26135f5e14cef8329068770572a7b51bb54344/mobly/records.py#L188)):*

The constructor takes care of that as they are all primitive types.
We only need to preserve stacktrace from the original as the new one would have been inaccurate.
</blockquote></details>

ah I see, the __init__ derive them from exception.  But then it assumes those properties has not been changed since the copy from ExceptionRecord's __init__().

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/316)*

xpconanfan:

---

*[mobly/records.py, line 188 at r1](https://reviewable.io:443/reviews/google/mobly/316#-KrmDOHDo07e2qIt24Ez:-KrmeK-nEhgRd0NJwSZs:bnuz5go) ([raw file](https://github.com/google/mobly/blob/4a26135f5e14cef8329068770572a7b51bb54344/mobly/records.py#L188)):*

ah I see, the __init__ derive them from exception.  But then it assumes those properties has not been changed since the copy from ExceptionRecord's __init__().
</blockquote></details>

Yes, it's like a copy constructor with a twist here ;)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/316)*

k2fong:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/316#-:-Krn3uCXTYFVfPi6i4Wz:bnfp4nl)*

PR #332, opened 2017-09-27
Fix runtime warnings in py3.

### LINKED ISSUE #254: `utils_test` has an unclosed file warning
This warning is shown when running unit tests in py3.

```
tests/mobly/utils_test.py::UtilsTest::test_stop_standing_subproc
  /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py:601: ResourceWarning: unclosed file <_io.BufferedReader name=14>
    testMethod()
  /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py:601: ResourceWarning: unclosed file <_io.BufferedReader name=16>
    testMethod()
```

xpconanfan:
Applying two fixes in both `stop_standing_subprocess` and the subprocess unit tests.

* Call `wait` on the actual Python object. Because Python throws a warning if `wait` is not called on the original object regardless of the actual state of the subprocess.
* Close pipes for subprocesses.

Fixes #254

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/332)

### DIFF
--- a/mobly/utils.py
+++ b/mobly/utils.py
@@ -357,6 +357,11 @@ def stop_standing_subprocess(proc, kill_signal=signal.SIGTERM):
         logging.exception('Failed to kill standing subprocess %d', pid)
     if failed:
         raise Error('Failed to kill standing subprocesses: %s' % failed)
+    # Call wait and close pipes on the original Python object so we don't get
+    # runtime warnings.
+    proc.stdout.close()
+    proc.stderr.close()
+    proc.wait()
     logging.debug('Stopped standing subprocess %d', pid)
 
 
--- a/tests/mobly/utils_test.py
+++ b/tests/mobly/utils_test.py
@@ -35,15 +35,23 @@ class UtilsTest(unittest.TestCase):
         self.sleep_cmd = 'timeout' if system == 'Windows' else 'sleep'
 
     def test_start_standing_subproc(self):
-        p = utils.start_standing_subprocess([self.sleep_cmd, '1'])
-        p1 = psutil.Process(p.pid)
-        self.assertTrue(p1.is_running())
+        try:
+            p = utils.start_standing_subprocess([self.sleep_cmd, '0.1'])
+            p1 = psutil.Process(p.pid)
+            self.assertTrue(p1.is_running())
+        finally:
+            p.stdout.close()
+            p.stderr.close()
+            p.wait()
 
     def test_stop_standing_subproc(self):
         p = utils.start_standing_subprocess([self.sleep_cmd, '4'])
         p1 = psutil.Process(p.pid)
+        start_time = time.time()
         utils.stop_standing_subprocess(p)
         self.assertFalse(p1.is_running())
+        stop_time = time.time()
+        self.assertTrue(stop_time - start_time < 0.1)
 
     @mock.patch(
         'mobly.controllers.android_device_lib.adb.list_occupied_adb_ports')


dthkao:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-Kv191JRbRWGQosw194R:b-25pw7d) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*
> ```Python
>     # Call wait and close pipes on the original Python object so we don't get
>     # runtime warnings.
>     proc.stdout.close()
> ```

The std*.close() commands I assume prevent the possible deadlock?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

kdart:
You should always call wait on a subprocess or else you get zombie processes. 

---

Reviewed 1 of 2 files at r1, 1 of 1 files at r2.

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvEPBo6V6Wya2ZXjCKy:b-xrcww3) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*
You should start the original process with stdio redirected to devnull. Many modules and libc functions expect those files to be open and when you close them bad things happen. 

example:

> DEVNULL = open(os.devnull, 'wb')
> process = Popen(["mycmd", "myarg"], stdout=DEVNULL, stderr=DEVNULL)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332#-:-KvETWQqJmC2O-1ZLAwe:b7j1ctq)*

xpconanfan:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvEUpUMniWDqplrnGdx:b-3o998y) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

You should start the original process with stdio redirected to devnull. Many modules and libc functions expect those files to be open and when you close them bad things happen. 

example:

> DEVNULL = open(os.devnull, 'wb')
> process = Popen(["mycmd", "myarg"], stdout=DEVNULL, stderr=DEVNULL)

</blockquote></details>

what does devnull do? would the output just be lost?
We do need to read the content of them.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

kdart:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvEY0fhiR82XONRup86:b-wbn9ka) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*

what does devnull do? would the output just be lost?
We do need to read the content of them.
</blockquote></details>

Actaully, it's stdin that is closed in start_standing_subprocess. That could cause problems. If you don't want to send it data, then redirect only stdin from /dev/null. 

If you do want to read the output, then yes you'll have to leave stdout and stderr open to a pipe. But you'll need to read from that pipe frequently or else the subprocess may block. 

This is why I use async event loops for subprocess management. :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

xpconanfan:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvEcOIoDZ8MV0sHvHeY:b-1cpz8o) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

Actaully, it's stdin that is closed in start_standing_subprocess. That could cause problems. If you don't want to send it data, then redirect only stdin from /dev/null. 

If you do want to read the output, then yes you'll have to leave stdout and stderr open to a pipe. But you'll need to read from that pipe frequently or else the subprocess may block. 

This is why I use async event loops for subprocess management. :-)

</blockquote></details>

ok, I don't think anything need to change here then?
stdin is not giving us warnings, and it seems reasonable to close the process's stdout/stderr pipes upon killing the process.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

kdart:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvF09iEy2QTHM_TPXrz:b-u5thc6) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*

ok, I don't think anything need to change here then?
stdin is not giving us warnings, and it seems reasonable to close the process's stdout/stderr pipes upon killing the process.
</blockquote></details>

The process was already waited on by the psutil code. You could remove that, and close the file before that. 

The closing of stdin may cause the subprocess to fail silently (EPIPE), but no warning in your main process. You should fix that in another PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

xpconanfan:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvF2WQXbj0uxdrLh6AN:b-20k4ms) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

The process was already waited on by the psutil code. You could remove that, and close the file before that. 

The closing of stdin may cause the subprocess to fail silently (EPIPE), but no warning in your main process. You should fix that in another PR.
</blockquote></details>

Python creates these warnings by checking if the original subprocess's `close` were called.
`psutil` operates with PID, so it won't touch the original subprocess object.

So we have to explicitly call close here as the comment mentions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

kdart:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvF3ARiEg59mqavhf2D:b6kvg84) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*

Python creates these warnings by checking if the original subprocess's `close` were called.
`psutil` operates with PID, so it won't touch the original subprocess object.

So we have to explicitly call close here as the comment mentions.
</blockquote></details>

Can't you get rid of the psutil call to wait then? You only need one. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

xpconanfan:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvF3fiYokOprcqTfpOB:b-y5cacv) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

Can't you get rid of the psutil call to wait then? You only need one. 
</blockquote></details>

No we need both, psutil is used here to get rid of all children processes.
the extra close call is to make python happy

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

kdart:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvF43zcsR--M9ZMKr5f:bv8sgto) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*

No we need both, psutil is used here to get rid of all children processes.
the extra close call is to make python happy
</blockquote></details>

I mean the call to the main subprocess wait. You can keep the child waits (but they really should be in their own process group).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

xpconanfan:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvFJ9Y51DzQ9Q06593I:bne35u) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

I mean the call to the main subprocess wait. You can keep the child waits (but they really should be in their own process group).
</blockquote></details>

No we have to do that wait because `psutil` supports timeout and regular py2 doesn't.
Otherwise the process main hang forever.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

kdart:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvFLp6GMDi6W-jW7CIX:bif3k6o) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*

No we have to do that wait because `psutil` supports timeout and regular py2 doesn't.
Otherwise the process main hang forever.
</blockquote></details>

Ah, yes. Py3 does support a timeout there, but Py2 doesn't.  The subprocess module in Py2 is pretty poor overall, but much better in Py3. 

But you should really never call wait unless you want to block. That's why it's called "wait".  Use poll method if you don't want to block. If the main thread is blocking forever then something isn't right with the process management... You might have to read from it first to drain the pipe that's it's blocked on.

You're also not getting the exit status, so the main thread won't know if the subprocess terminated normally or not.

But given that this change is just working around another issue I'll just say go with it. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

kdart:
:lgtm: Good enough for a workaround. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332#-:-KvFQQ9aYNjn94h2ZfJF:bopz58i)*

xpconanfan:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvFR-WWME8ikymjgqiL:b-hs1zha) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

Ah, yes. Py3 does support a timeout there, but Py2 doesn't.  The subprocess module in Py2 is pretty poor overall, but much better in Py3. 

But you should really never call wait unless you want to block. That's why it's called "wait".  Use poll method if you don't want to block. If the main thread is blocking forever then something isn't right with the process management... You might have to read from it first to drain the pipe that's it's blocked on.

You're also not getting the exit status, so the main thread won't know if the subprocess terminated normally or not.

But given that this change is just working around another issue I'll just say go with it. 

</blockquote></details>

Thx. wanna file an issue to track any improvements needed for this function?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

kdart:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvFVRQJET4b0WVbP0gp:b4ilmiq) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*

Thx. wanna file an issue to track any improvements needed for this function?
</blockquote></details>

OK. Maybe even a PR. :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

kdart:

---

*[mobly/utils.py, line 362 at r1](https://reviewable.io:443/reviews/google/mobly/332#-Kv191JRbRWGQosw194Q:-KvFbnl6kQHf07nTN4_o:bb6ywmx) ([raw file](https://github.com/google/mobly/blob/d3b74e054873e3bbdd7b65101d095760a90ddf64/mobly/utils.py#L362)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

OK. Maybe even a PR. :-)
</blockquote></details>

LGTM

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/332)*

PR #344, opened 2017-10-04
Delayed termination support (Expect)

xpconanfan:
* Add new module `mobly.expects` with expectation APIs.
* Modify base_test to utilize the new module.
* Unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/344)

### DIFF
--- a/mobly/asserts.py
+++ b/mobly/asserts.py
@@ -31,13 +31,14 @@ _pyunit_proxy = _ProxyTest()
 
 
 def assert_equal(first, second, msg=None, extras=None):
-    """Assert an expression evaluates to true, otherwise fail the test.
+    """Assert the equality of objects, otherwise fail the test.
 
     Error message is "first != second" by default. Additional explanation can
     be supplied in the message.
 
     Args:
-        expr: The expression that is evaluated.
+        first: The first object to compare.
+        second: The second object to compare.
         msg: A string that adds additional info about the failure.
         extras: An optional field for extra information to be included in
             test result.
@@ -98,9 +99,8 @@ def assert_raises_regex(expected_exception,
         extras: An optional field for extra information to be included in
             test result.
     """
-    context = _AssertRaisesContext(expected_exception,
-                                   expected_regex,
-                                   extras=extras)
+    context = _AssertRaisesContext(
+        expected_exception, expected_regex, extras=extras)
     return context
 
 
@@ -242,9 +242,12 @@ def fail(msg, extras=None):
 def explicit_pass(msg, extras=None):
     """Explicitly pass a test.
 
-    A test with not uncaught exception will pass implicitly so the usage of
-    this is optional. It is intended for reporting extra information when a
-    test passes.
+    This will pass the test explicitly regardless of any other error happened
+    in the test body. E.g. even if errors have been recorded with `expects`,
+    the test will still be marked pass if this is called.
+
+    A test without uncaught exception will pass implicitly so this should be
+    used scarcely.
 
     Args:
         msg: A string explaining the details of the passed test.
@@ -275,8 +278,8 @@ class _AssertRaisesContext(object):
                 exc_name = self.expected.__name__
             except AttributeError:
                 exc_name = str(self.expected)
-            raise signals.TestFailure('%s not raised' % exc_name,
-                                      extras=self.extras)
+            raise signals.TestFailure(
+                '%s not raised' % exc_name, extras=self.extras)
         if not issubclass(exc_type, self.expected):
             # let unexpected exceptions pass through
             return False
@@ -289,7 +292,7 @@ class _AssertRaisesContext(object):
             expected_regexp = re.compile(expected_regexp)
         if not expected_regexp.search(str(exc_value)):
             raise signals.TestFailure(
-                '"%s" does not match "%s"' %
-                (expected_regexp.pattern, str(exc_value)),
+                '"%s" does not match "%s"' % (expected_regexp.pattern,
+                                              str(exc_value)),
                 extras=self.extras)
         return True
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -384,10 +384,6 @@ class BaseTestClass(object):
                     # Check if anything failed by `expects`.
                     if before_count < expects.recorder.error_count:
                         teardown_test_failed = True
-            # No exception is thrown from test and teardown, if `expects` has
-            # error, the test should fail with the first error in `expects`.
-            if expects.recorder.has_error and not teardown_test_failed:
-                fail_by_expect = True
         except (signals.TestFailure, AssertionError) as e:
             tr_record.test_fail(e)
         except signals.TestSkip as e:
@@ -404,8 +400,11 @@ class BaseTestClass(object):
             # Exception happened during test.
             tr_record.test_error(e)
         else:
-            if fail_by_expect:
+            # No exception is thrown from test and teardown, if `expects` has
+            # error, the test should fail with the first error in `expects`.
+            if expects.recorder.has_error and not teardown_test_failed:
                 tr_record.test_fail()
+            # Otherwise the test passed.
             elif not teardown_test_failed:
                 tr_record.test_pass()
         finally:
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -19,6 +19,7 @@ import inspect
 import logging
 import sys
 
+from mobly import expects
 from mobly import records
 from mobly import signals
 from mobly import runtime_test_info
@@ -345,8 +346,12 @@ class BaseTestClass(object):
         tr_record.test_begin()
         self.current_test_info = runtime_test_info.RuntimeTestInfo(
             test_name, self.log_path, tr_record)
+        expects.recorder.reset_with_record(tr_record)
         logging.info('%s %s', TEST_CASE_TOKEN, test_name)
+        # Did teardown_test throw an error.
         teardown_test_failed = False
+        # Should the test fail solely because of errors recorded by `expects`.
+        fail_by_expect = False
         try:
             try:
                 try:
@@ -366,6 +371,7 @@ class BaseTestClass(object):
                                   self.current_test_name)
                 raise
             finally:
+                before_count = expects.recorder.error_count
                 try:
                     self._teardown_test(test_name)
                 except signals.TestAbortSignal:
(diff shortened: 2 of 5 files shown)


kdart:
So the point of this is to have "expectations" that are like asserts but don't throw exceptions?

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/344#-:-KwCaV3lWZ8efS-XaF3U:bivcchi)*

k2fong:
Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[mobly/base_test.py, line 390 at r1](https://reviewable.io:443/reviews/google/mobly/344#-KwffUdWxDf6rGL99RSU:-KwffUdWxDf6rGL99RSV:b-3cytoa) ([raw file](https://github.com/google/mobly/blob/e153a9a924073e236e181f65425df0c769e8505e/mobly/base_test.py#L390)):*
>             if expects.recorder.has_error and not teardown_test_failed:
>                 fail_by_expect = True
how about the case in which there is expect failure (from prior teardown) and also teardown failures?   Current logic seems to not consider it as expect failure

---

*[mobly/expects.py, line 32 at r1](https://reviewable.io:443/reviews/google/mobly/344#-Kwfhe9IXDVCt1nX5pMU:-Kwfhe9IXDVCt1nX5pMV:b-4lc5wf) ([raw file](https://github.com/google/mobly/blob/e153a9a924073e236e181f65425df0c769e8505e/mobly/expects.py#L32)):*
> ```Python
> 
>     def __init__(self):
>         self._reset_internal_states()
> ```

self._record = None?  _record is never set unless reset_with_record() is called

---

*[mobly/expects.py, line 36 at r1](https://reviewable.io:443/reviews/google/mobly/344#-KwfgMnU4zl0dBd3-xKp:-KwfgMnU4zl0dBd3-xKq:b-orr4pl) ([raw file](https://github.com/google/mobly/blob/e153a9a924073e236e181f65425df0c769e8505e/mobly/expects.py#L36)):*
> ```Python
>     def _reset_internal_states(self):
>         """Resets the internal state of the recorder."""
>         self._has_error = False
> ```

suggest removing _has_error and replace it with on _count > 0

---

*[mobly/expects.py, line 83 at r1](https://reviewable.io:443/reviews/google/mobly/344#-KwfiS3ogQOo3N8Im2SX:-KwfiS3ogQOo3N8Im2SY:b3kbvye) ([raw file](https://github.com/google/mobly/blob/e153a9a924073e236e181f65425df0c769e8505e/mobly/expects.py#L83)):*
> ```Python
>     try:
>         asserts.assert_true(condition, msg, extras)
>     except signals.TestSignal as e:
> ```

What you think about adding a log message to indicate an expect failure so that it will be in context to when it occurs to help with debugging logs?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/344)*

xpconanfan:
Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[mobly/base_test.py, line 390 at r1](https://reviewable.io:443/reviews/google/mobly/344#-KwffUdWxDf6rGL99RSU:-Kwh0tGJ_EQpBuDcgfTT:b-rh1da7) ([raw file](https://github.com/google/mobly/blob/e153a9a924073e236e181f65425df0c769e8505e/mobly/base_test.py#L390)):*

>             if expects.recorder.has_error and not teardown_test_failed:
>                 fail_by_expect = True
how about the case in which there is expect failure (from prior teardown) and also teardown failures?   Current logic seems to not consider it as expect failure
</blockquote></details>

Sorry I'm not clear on exactly what case you are referring to.
Can you clarify with an example?

---

*[mobly/expects.py, line 32 at r1](https://reviewable.io:443/reviews/google/mobly/344#-Kwfhe9IXDVCt1nX5pMU:-Kwh1Q06WrZCivktyEtU:b-896fix) ([raw file](https://github.com/google/mobly/blob/e153a9a924073e236e181f65425df0c769e8505e/mobly/expects.py#L32)):*

self._record = None?  _record is never set unless reset_with_record() is called
</blockquote></details>

Done.

---

*[mobly/expects.py, line 36 at r1](https://reviewable.io:443/reviews/google/mobly/344#-KwfgMnU4zl0dBd3-xKp:-Kwh1ZJNmY9HwjLFKtVk:b-896fix) ([raw file](https://github.com/google/mobly/blob/e153a9a924073e236e181f65425df0c769e8505e/mobly/expects.py#L36)):*

suggest removing _has_error and replace it with on _count > 0
</blockquote></details>

Done.

---

*[mobly/expects.py, line 83 at r1](https://reviewable.io:443/reviews/google/mobly/344#-KwfiS3ogQOo3N8Im2SX:-Kwh325VBzM3vRqQbe1S:bj5qfja) ([raw file](https://github.com/google/mobly/blob/e153a9a924073e236e181f65425df0c769e8505e/mobly/expects.py#L83)):*

What you think about adding a log message to indicate an expect failure so that it will be in context to when it occurs to help with debugging logs?
</blockquote></details>

Makes sense. Since the `asserts` would cause similar log lines.
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/344)*

k2fong:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/base_test.py, line 390 at r1](https://reviewable.io:443/reviews/google/mobly/344#-KwffUdWxDf6rGL99RSU:-KwhdxQnnfrdRlExynaw:b-60xxyk) ([raw file](https://github.com/google/mobly/blob/e153a9a924073e236e181f65425df0c769e8505e/mobly/base_test.py#L390)):*

Sorry I'm not clear on exactly what case you are referring to.
Can you clarify with an example?
</blockquote></details>

ah, I see, I got confused with the variable name 'failed_by_expect' and that its meaning is only pertinent in the context of teardown. How about instead of introducing the variable 'failed_by_expect', restructure the below else clause to something like so? seems clearer and more straight forward to me but could be just me, your call.

```
if not teardown_test_failed:
    tr_record.test_fail() if expects.recorder.has_error else tr_record.test_pass()
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/344)*

xpconanfan:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/base_test.py, line 390 at r1](https://reviewable.io:443/reviews/google/mobly/344#-KwffUdWxDf6rGL99RSU:-Kwn1cayPuYhyazyXbyg:b-896fix) ([raw file](https://github.com/google/mobly/blob/e153a9a924073e236e181f65425df0c769e8505e/mobly/base_test.py#L390)):*

ah, I see, I got confused with the variable name 'failed_by_expect' and that its meaning is only pertinent in the context of teardown. How about instead of introducing the variable 'failed_by_expect', restructure the below else clause to something like so? seems clearer and more straight forward to me but could be just me, your call.

```
if not teardown_test_failed:
    tr_record.test_fail() if expects.recorder.has_error else tr_record.test_pass()
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/344)*

k2fong:
:lgtm: besides the two minor comments below

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[mobly/base_test.py, line 354 at r2](https://reviewable.io:443/reviews/google/mobly/344#-KwpNtetx7keoqKTZfNi:-KwpNteu2pQNoMLVZk-U:b-pmehwn) ([raw file](https://github.com/google/mobly/blob/7ecfa31f3f43194f2799da0b173188a7d3bf6a02/mobly/base_test.py#L354)):*
>         # Should the test fail solely because of errors recorded by `expects`.
>         fail_by_expect = False
fail_by_expect is unused now

---

*[mobly/expects.py, line 63 at r2](https://reviewable.io:443/reviews/google/mobly/344#-KwpQiMFK7XoDaz3oXW8:-KwpQiMGgeTLeHTxMvq_:b-w87br0) ([raw file](https://github.com/google/mobly/blob/7ecfa31f3f43194f2799da0b173188a7d3bf6a02/mobly/expects.py#L63)):*
> ```Python
>                                error)
> 
>     def reset_with_record(self, record):
> ```

nit.  seems like this can be refactored and combine with _reset_internal_states() by defaulting record to None.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/344#-:-KwpR2oFhazGU202zHuj:bsl5tzd)*

xpconanfan:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[mobly/base_test.py, line 354 at r2](https://reviewable.io:443/reviews/google/mobly/344#-KwpNtetx7keoqKTZfNi:-KwtUvR6ha-FNkBkWoa_:b-896fix) ([raw file](https://github.com/google/mobly/blob/7ecfa31f3f43194f2799da0b173188a7d3bf6a02/mobly/base_test.py#L354)):*

>         # Should the test fail solely because of errors recorded by `expects`.
>         fail_by_expect = False
fail_by_expect is unused now
</blockquote></details>

Done.

---

*[mobly/expects.py, line 63 at r2](https://reviewable.io:443/reviews/google/mobly/344#-KwpQiMFK7XoDaz3oXW8:-KwtUgA_04D4EKXThd4y:b-896fix) ([raw file](https://github.com/google/mobly/blob/7ecfa31f3f43194f2799da0b173188a7d3bf6a02/mobly/expects.py#L63)):*

nit.  seems like this can be refactored and combine with _reset_internal_states() by defaulting record to None.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/344)*

kdart:
:lgtm:
Just keep in mind that system tests, unlike unit tests, can change the state of the system. Some tests may depend on the state that a previous test leaves the system in.  It might actually be appropriate to abort the test run rather than continue on error. 

---

Reviewed 2 of 5 files at r1, 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/344#-:-KwuJirpSavDGGjxdHDU:b539qsc)*

xpconanfan:
@kdart, we generally don't encourage ppl to make tests depend on each other's states.
But for aborting all subsequent tests, we do have `asserts.abort_class` which will do exactly that :)

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/344#-:-KwwPuDwb-IQQE3ywAxm:b-r2h7op)*

PR #352, opened 2017-10-10
Move test_suite functionality to test_runner

mswest46:
test_runner can edit test methods belonging to classes that have already been added to the runner's execution plan, and can print a list of test methods that are in the runner's execution plan. Users of the test_runner must do their own parsing of command line arguments for test selection.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/352)

### DIFF
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -167,12 +167,11 @@ def _compute_selected_tests(selected_tests):
     """Computes tests to run for each class from selector strings.
 
     This function transforms a list of selector strings (such as FooTest or
-    FooTest.test_method_a) to a dict where keys are test classes (strings), and
+    FooTest.test_method_a) to an ordered dict where keys are test classes (strings), and
     values are lists of selected test methods (strings) in those classes. 
     None means all methods in that class are selected.
 
     Args:
-        test_classes: (list of class) all classes that are part of this suite.
         selected_tests: (list of string) list of tests to execute, eg:
             [
                 'FooTest',
@@ -183,8 +182,8 @@ def _compute_selected_tests(selected_tests):
             May be empty, in which case all tests in the class are selected.
 
     Returns:
-        dict: str class -> list(str method):
-        identifiers for TestRunner. For the above example:
+        dict: str class -> list(str method).
+        For above example: 
         {
             'FooTest': None,
             'BarTest': None,
@@ -210,7 +209,7 @@ def _compute_selected_tests(selected_tests):
 
 
 def _get_all_tests(test_class, config):
-    """Returns all test methods (including generated ones) belonging to test_class(config)."""
+    """Returns all test methods (including generated ones) belonging to  test_class(config)."""
     config_copy = config.copy()  # just in case.
     test_instance = test_class(config_copy)
     test_instance.setup_generated_tests
@@ -402,6 +401,7 @@ class TestRunner(object):
                             'Trying to selected nonexistent test methods %s from test class %s'
                             % (nonexistent_methods, test_class_name))
 
+                # class and method selections are valid. Add new test_run_info to plan.
                 test_run_info.tests = selected_test_methods
                 new_test_run_infos.append(test_run_info)
         if test_selector:
@@ -417,8 +417,6 @@ class TestRunner(object):
             FooTest.test_method_a
             FooTest.test_method_b
             BarTest.test_method_a
-        Note that this function will instantiate test classes in order to generate tests. 
-
         """
         test_names_to_print = []
         for test_run_info in self._test_run_infos:
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -17,6 +17,7 @@ from future import standard_library
 standard_library.install_aliases()
 
 import argparse
+import collections
 import copy
 import functools
 import inspect
@@ -161,6 +162,57 @@ def _print_test_names(test_class):
     for name in cls.get_existing_test_names():
         print(name)
 
+def _compute_selected_tests(selected_tests): 
+    """Computes tests to run for each class from selector strings.
+
+    This function transforms a list of selector strings (such as FooTest or
+    FooTest.test_method_a) to a dict where keys are test classes (strings), and
+    values are lists of selected test methods (strings) in those classes. 
+    None means all methods in that class are selected.
+
+    Args:
+        test_classes: (list of class) all classes that are part of this suite.
+        selected_tests: (list of string) list of tests to execute, eg:
+            [
+                'FooTest',
+                'BarTest',
+                'BazTest.test_method_a',
+                'BazTest.test_method_b'
+            ].
+            May be empty, in which case all tests in the class are selected.
+
+    Returns:
+        dict: str class -> list(str method):
+        identifiers for TestRunner. For the above example:
+        {
+            'FooTest': None,
+            'BarTest': None,
+            'BazTest': ['test_method_a', 'test_method_b'],
+        }
+    """
+    test_class_name_to_tests = collections.OrderedDict()
+    for test_name in selected_tests:
+        if '.' in test_name:  # Has a test method
+            (test_class_name, test_name) = test_name.split('.')
+            if test_class_name not in test_class_name_to_tests:
+                # Never seen this class before
+                test_class_name_to_tests[test_class_name] = [test_name]
+            elif test_class_name_to_tests[test_class_name] is None:
+                # Already running all tests in this class, so ignore this extra
+                # test.
+                pass
+            else:
+                test_class_name_to_tests[test_class_name].append(test_name)
+        else:  # No test method; run all tests in this class.
+            test_class_name_to_tests[test_name] = None
+    return test_class_name_to_tests
+
+def _get_all_tests(test_class, config): 
+    """Returns all test methods (including generated ones) belonging to test_class(config)."""
+    config_copy = config.copy()  # just in case. 
+    test_instance = test_class(config_copy)
+    test_instance.setup_generated_tests
+    return test_instance.get_existing_test_names()
 
 def verify_controller_module(module):
     """Verifies a module object follows the required interface for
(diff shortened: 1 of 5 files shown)


xpconanfan:
Review status: 0 of 5 files reviewed at latest revision, 7 unresolved discussions.

---

*[mobly/test_runner.py, line 166 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6HGkZR6kxHXYE0Ry8:-Kw6HGkZR6kxHXYE0Ry9:bsnhvmt) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L166)):*
> ```Python
> 
> 
> def _compute_selected_tests(selected_tests):
> ```

shouldn't this be in config parser?

---

*[mobly/test_runner.py, line 176 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6EUe_8Gr89nEa4PgK:-Kw6EUe_8Gr89nEa4PgL:b-voqavw) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L176)):*
> ```Python
>     Args:
>         selected_tests: (list of string) list of tests to execute, eg:
>             [
> ```

put the code example above args so the input and output are next to each other and it's easier to read.

Also pls follow the sphinx format here to mark code blocks properly
https://github.com/google/mobly/blob/b6ab47685d1edd2c6d2124470edd0f0e1f6276fc/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L18

---

*[mobly/test_runner.py, line 195 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6FJH_DgWvJ3f2XmZg:-Kw6FJH_DgWvJ3f2XmZh:b-wdq4b4) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L195)):*
> ```Python
>     test_class_name_to_tests = collections.OrderedDict()
>     for test_name in selected_tests:
>         if '.' in test_name:  # Has a test method
> ```

make a macro for '.'
say, `TEST_NAME_SEPARATOR`, with a comment saying what it separates

---

*[mobly/test_runner.py, line 200 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6FlDKVNxwJx5bOdWB:-Kw6FlDKVNxwJx5bOdWC:bzcshjj) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L200)):*
> ```Python
>                 # Never seen this class before
>                 test_class_name_to_tests[test_class_name] = [test_name]
>             elif test_class_name_to_tests[test_class_name] is None:
> ```

Seems odd to use `None` as a indicator, but no big deal.
Do make a macro for the indicator and make sure it's documented.

---

*[mobly/test_runner.py, line 212 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6G9MUpqFCUFdnQrK7:-Kw6G9MUpqFCUFdnQrK8:b-fs9fgg) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L212)):*
> ```Python
> 
> def _get_all_tests(test_class, config):
>     """Returns all test methods (including generated ones) belonging to  test_class(config)."""
> ```

longer than 80 char.
All doc strings should also follow the 80 char rule.
Unfortunately yapf can't handle this so you have to do it manually.

---

*[mobly/test_runner.py, line 352 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6GLuQe-wc4GgC-mRN:-Kw6GLuQe-wc4GgC-mRO:b-vmoh8o) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L352)):*
> ```Python
>                 (self._test_bed_name, config.test_bed_name))
> 
>         if not issubclass(test_class, base_test.BaseTestClass):
> ```

do we really need to add this?
In python, if it walks like a duck and it quacks like a duck, it is a duck
https://en.wikipedia.org/wiki/Duck_typing

---

*[mobly/test_runner.py, line 360 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6HRQpUcnY3zf5maGd:-Kw6HRQpUcnY3zf5maGe:b-2zgljq) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L360)):*
> ```Python
>                 config=config, test_class=test_class, tests=tests))
> 
>     def select_test_methods(self, test_selections):
> ```

`add_test_class` already lets you specify which tests to run, why do we need to add another method to specify what tests to run separately?
This adds complexity without clear benefit.
It also locks test runner to a particular format of test identifier.

Let's split the test token parsing into config_parser.
And in your suite, you can directly pass the test case names to `add_test_class`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/352)*

xpconanfan:
Review status: 0 of 5 files reviewed at latest revision, 8 unresolved discussions.

---

*[mobly/test_runner.py, line 413 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6IMZXWmminZjEnKEz:-Kw6IMZXWmminZjEnKF-:b-fg88db) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L413)):*
> ```Python
>         self._test_run_infos = new_test_run_infos
> 
>     def print_all_test_methods(self):
> ```

Let's not have a `print` method in `TestRunner`
You can add a `get_execution_plan`, which returns a dict of {ClassName: [test_name]}

then the caller (probably the suite file) can choose to print the info in whatever format they like

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/352)*

mswest46:
Review status: 0 of 5 files reviewed at latest revision, 8 unresolved discussions.

---

*[mobly/test_runner.py, line 166 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6HGkZR6kxHXYE0Ry8:-Kw6JlQLwYMK39_KzI-8:bc4vmhp) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L166)):*

shouldn't this be in config parser?
</blockquote></details>

Could be, I wasn't sure because it's not exactly parsing configs, it's parsing a list of strings. But happy to move it

---

*[mobly/test_runner.py, line 200 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6FlDKVNxwJx5bOdWB:-Kw6LSBhxl7wFdbwOgu1:b-5p4cxr) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L200)):*

Seems odd to use `None` as a indicator, but no big deal.
Do make a macro for the indicator and make sure it's documented.
</blockquote></details>

Sure, can make it 'all' if you want? 

---

*[mobly/test_runner.py, line 352 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6GLuQe-wc4GgC-mRN:-Kw6LdJOk9zB5_Eba5C-:bpkmq92) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L352)):*

do we really need to add this?
In python, if it walks like a duck and it quacks like a duck, it is a duck
https://en.wikipedia.org/wiki/Duck_typing
</blockquote></details>

I was bringing over the suite_runner check: https://github.com/google/mobly/blob/ec37cec8a66f929bafc1db12f76ac793bcb9a679/mobly/suite_runner.py#L79

---

*[mobly/test_runner.py, line 360 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6HRQpUcnY3zf5maGd:-Kw6M7BxBvkowrrIyWr5:b-dufuzp) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L360)):*

`add_test_class` already lets you specify which tests to run, why do we need to add another method to specify what tests to run separately?
This adds complexity without clear benefit.
It also locks test runner to a particular format of test identifier.

Let's split the test token parsing into config_parser.
And in your suite, you can directly pass the test case names to `add_test_class`
</blockquote></details>

I thought about how to do this, but couldn't quite make it work. 

Currently, the user (maker of suite) is responsible for adding the test classes and configs, and then selecting which tests they want to run with a list of strings, in that order. 

Your suggestion requires the user to call the parser method, retrieve the methods that belong to particular instances test_class(config), and pass those methods in when they call add_test_class(test_class, config, test_methods). This seems like extra work/thought from the user. 

An alternative is to write something that adds tests in bulk (kind of like the old suite_runner did, but with multi-config file options.) Something like: 
```
def add_test_classes_in_bulk(list(tuple) test_classes_and_configs, list(str) test_selections)
```
In this case, the user would just need to create a list of class/config pairs, and pass in the parsed test_selection strings. We would do the selection (and validation) before adding the classes to the test runner. 

Does that make sense? I might be being a bit thick here, but I don't see how to use the existing methods without making the creator of the suite think more than they have to. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/352)*

xpconanfan:
Review status: 0 of 5 files reviewed at latest revision, 8 unresolved discussions.

---

*[mobly/test_runner.py, line 360 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6HRQpUcnY3zf5maGd:-Kw6Tynlw-PRf1JsEiAi:bt13v1x) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L360)):*
<details><summary><i>Previously, mswest46 (Michael West) wrote…</i></summary><blockquote>

I thought about how to do this, but couldn't quite make it work. 

Currently, the user (maker of suite) is responsible for adding the test classes and configs, and then selecting which tests they want to run with a list of strings, in that order. 

Your suggestion requires the user to call the parser method, retrieve the methods that belong to particular instances test_class(config), and pass those methods in when they call add_test_class(test_class, config, test_methods). This seems like extra work/thought from the user. 

An alternative is to write something that adds tests in bulk (kind of like the old suite_runner did, but with multi-config file options.) Something like: 
```
def add_test_classes_in_bulk(list(tuple) test_classes_and_configs, list(str) test_selections)
```
In this case, the user would just need to create a list of class/config pairs, and pass in the parsed test_selection strings. We would do the selection (and validation) before adding the classes to the test runner. 

Does that make sense? I might be being a bit thick here, but I don't see how to use the existing methods without making the creator of the suite think more than they have to. 
</blockquote></details>

The idea is to keep the test runner generic so different users can use it differently.

Your parser function can return a format that makes it easy to call the TestRunner API with.
e.g. it can return a list of tuples with the first element being the class and the second element being the specific tests to run, then it's a simple loop to call `add_test_class`.

In pyunit if you want to create a suite, you follow a similar pattern.

Building the parsing directly into the test runner also means we are gonna have to deal with all kinds of filtering pattern within test runner, which can quickly become something hard to manage bc different users are gonna want to filter their test sets differently.

For the suite owner it is also easier to maintain python statements than a long string specifying what is what.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/352)*

xpconanfan:
Review status: 0 of 5 files reviewed at latest revision, 8 unresolved discussions.

---

*[mobly/test_runner.py, line 200 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6FlDKVNxwJx5bOdWB:-Kw6WHkDncjPM4j7uopz:b32wwp) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L200)):*
<details><summary><i>Previously, mswest46 (Michael West) wrote…</i></summary><blockquote>

Sure, can make it 'all' if you want? 
</blockquote></details>

sg

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/352)*

xpconanfan:
Review status: 0 of 5 files reviewed at latest revision, 8 unresolved discussions.

---

*[mobly/test_runner.py, line 166 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6HGkZR6kxHXYE0Ry8:-Kw6WO1KYYVX3-L1ewOL:bj829pd) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L166)):*
<details><summary><i>Previously, mswest46 (Michael West) wrote…</i></summary><blockquote>

Could be, I wasn't sure because it's not exactly parsing configs, it's parsing a list of strings. But happy to move it
</blockquote></details>

Yes, pls

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/352)*

mswest46:
Review status: 0 of 5 files reviewed at latest revision, 8 unresolved discussions.

---

*[mobly/test_runner.py, line 360 at r1](https://reviewable.io:443/reviews/google/mobly/352#-Kw6HRQpUcnY3zf5maGd:-Kw6X-ecJl1xeySAOacH:b-n77d6) ([raw file](https://github.com/google/mobly/blob/b568d9b2a9e4d513d97b42ac3b8d062abb877f49/mobly/test_runner.py#L360)):*

The idea is to keep the test runner generic so different users can use it differently.

Your parser function can return a format that makes it easy to call the TestRunner API with.
e.g. it can return a list of tuples with the first element being the class and the second element being the specific tests to run, then it's a simple loop to call `add_test_class`.

In pyunit if you want to create a suite, you follow a similar pattern.

Building the parsing directly into the test runner also means we are gonna have to deal with all kinds of filtering pattern within test runner, which can quickly become something hard to manage bc different users are gonna want to filter their test sets differently.

For the suite owner it is also easier to maintain python statements than a long string specifying what is what.
</blockquote></details>

sg. One final qualm I have is about the multiple configs. Suppose I have a test class FooTest that generates test methods based on config. So if I can call FooTest(configA).test_method_A and FooTest(configB).test_method_B, but not vice versa. When a user specifies a test via command line as 'FooTest.test_method_A', should we do any checking to make sure test_method_A is generated by FooTest, and if so, how do we know which config user is talking about? Or should we just trust the user to call add_test_class(FooTest, configA, ['test_method_A']) as opposed to add_test_class(FooTest, configB, ['test_method_A']). If the latter, we could help the user out by adding a public get_methods_belonging_to_test_instance(test_class, config) that does the generating of test_methods and returns them itself. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/352)*

PR #362, opened 2017-11-04
Adding special config key - 'start_logcat' in AndroidDevice controlle…

gauravkatoch007:
…r module

start_logcat param in AndroidDevice controller module decides whether to start logcat service or not. If not specified, logcat services will be started (current default Behavior).

Example-
Name: XyzTestBed
    Controllers:
        AndroidDevice:
          - serial: xyz
            start_logcat: False

In above case, for AndroidDevice with serial xyz, logcat service will not be started.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/362)

### DIFF
--- a/mobly/asserts.py
+++ b/mobly/asserts.py
@@ -31,13 +31,14 @@ _pyunit_proxy = _ProxyTest()
 
 
 def assert_equal(first, second, msg=None, extras=None):
-    """Assert an expression evaluates to true, otherwise fail the test.
+    """Assert the equality of objects, otherwise fail the test.
 
     Error message is "first != second" by default. Additional explanation can
     be supplied in the message.
 
     Args:
-        expr: The expression that is evaluated.
+        first: The first object to compare.
+        second: The second object to compare.
         msg: A string that adds additional info about the failure.
         extras: An optional field for extra information to be included in
             test result.
@@ -98,9 +99,8 @@ def assert_raises_regex(expected_exception,
         extras: An optional field for extra information to be included in
             test result.
     """
-    context = _AssertRaisesContext(expected_exception,
-                                   expected_regex,
-                                   extras=extras)
+    context = _AssertRaisesContext(
+        expected_exception, expected_regex, extras=extras)
     return context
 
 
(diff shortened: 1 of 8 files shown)


xpconanfan:

---

*[mobly/controllers/android_device.py, line 472 at r1](https://reviewable.io:443/reviews/google/mobly/362#-KyHPCpTCYCF0GuaS9tO:-KyHPCpTCYCF0GuaS9tP:bd0pmf1) ([raw file](https://github.com/google/mobly/blob/3e442b3d282d0cabc3459f51e3f38afe8012f33e/mobly/controllers/android_device.py#L472)):*
> ```Python
>         self._debug_tag = tag
>         self.log.extra['tag'] = tag
>         
> ```

random `...`? does this code even run?

---

*[mobly/controllers/android_device.py, line 481 at r1](https://reviewable.io:443/reviews/google/mobly/362#-KyHPhP_VsJz584Toho-:-KyHPhP_VsJz584Toho0:byhrfz6) ([raw file](https://github.com/google/mobly/blob/3e442b3d282d0cabc3459f51e3f38afe8012f33e/mobly/controllers/android_device.py#L481)):*
> ```Python
> 
>     @log_path.setter
>     def log_path(self, log_path_tag):
> ```

we don't want a setter for this right?
`log_path` should be a calculated value based on the initial config of the obj.
in our discussion, we did not mention setting `log_path` during the lifecycle of the obj as it involves merging existing logs to the new location which is very complicated.

---

*[mobly/controllers/android_device.py, line 499 at r1](https://reviewable.io:443/reviews/google/mobly/362#-KyHPL93KDayiju6vxdS:-KyHPL93KDayiju6vxdT:b-rpaner) ([raw file](https://github.com/google/mobly/blob/3e442b3d282d0cabc3459f51e3f38afe8012f33e/mobly/controllers/android_device.py#L499)):*
> ```Python
> 
>     def set_serial(self, serial):
>         """Update serial of android device."""
> ```

need proper docstring
pls follow style guide
https://google.github.io/styleguide/pyguide.html

---

*[mobly/controllers/android_device.py, line 885 at r1](https://reviewable.io:443/reviews/google/mobly/362#-KyHP0lJkZjBi37z2Sqj:-KyHP0lKOFmztp4TOyGh:b8elait) ([raw file](https://github.com/google/mobly/blob/3e442b3d282d0cabc3459f51e3f38afe8012f33e/mobly/controllers/android_device.py#L885)):*
> ```Python
>             clear: If True, clear device log before starting logcat.
>         """
>         if hasattr(self, KEY_START_LOGCAT) and not self.start_logcat:
> ```

this will cause the method to never be able to start logcat if the flag is set, which is not correct.
the flag should be processed in `create`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/362)*

xpconanfan:
Obsolete by #366

PR #363, opened 2017-11-07
Support remote devices whose serials can change during test.

xpconanfan:
The key problem is that the `log_path` is hard coded to include the device's serial number as part of it. Now that the serial number may change, we should allow users to decouple the log_path from the serial number itself.

So now we allow users to update the serial and log_path during the test to accommodate such devices.

The recommended approach is to skip adb logcat via the config key, then in `setup_class`, set the `log_path` to a location agnostic to the device's actual serial number, then when the device reboots with a different serial, call `update_serial` to use the new serial number on the object.

* Add API to properly update the serial number.
* Allow log_path to be modified.
* Allow not starting logcat collection by default.

@gauravkatoch007

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/363)

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -540,11 +540,10 @@ class AndroidDevice(object):
 
         Args:
             new_serial: string, the new serial number for the same device.
+
+        Raises:
+            DeviceError: tries to update serial when any service is running.
         """
-        if new_serial == self._serial:
-            raise DeviceError(
-                self, 'Serial is already %s, cannot set to the same value.' %
-                self._serial)
         if self.has_active_service:
             raise DeviceError(
                 self,
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -17,8 +17,10 @@ from builtins import open
 from past.builtins import basestring
 
 import contextlib
+import distutils
 import logging
 import os
+import shutil
 import time
 
 from mobly import logger as mobly_logger
@@ -44,7 +46,12 @@ ANDROID_DEVICE_EMPTY_CONFIG_MSG = 'Configuration is empty, abort!'
 ANDROID_DEVICE_NOT_LIST_CONFIG_MSG = 'Configuration should be a list, abort!'
 
 # Keys for attributes in configs that alternate the controller module behavior.
+# If this is False for a device, errors from that device will be ignored
+# during `create`. Default is True.
 KEY_DEVICE_REQUIRED = 'required'
+# If True, logcat collection will not be started during `create`.
+# Default is False.
+KEY_SKIP_LOGCAT = 'skip_logcat'
 
 # Default Timeout to wait for boot completion
 DEFAULT_TIMEOUT_BOOT_COMPLETION_SECOND = 15 * 60
@@ -142,6 +149,10 @@ def _start_services_on_ads(ads):
     running_ads = []
     for ad in ads:
         running_ads.append(ad)
+        skip_logcat = getattr(ad, KEY_SKIP_LOGCAT, False)
+        print('#######$$$$$$$', ad.serial, skip_logcat)
+        if skip_logcat:
+            continue
         try:
             ad.start_services()
         except Exception:
@@ -414,10 +425,12 @@ class AndroidDevice(object):
     """
 
     def __init__(self, serial=''):
-        self.serial = serial
+        self._serial = serial
         # logging.log_path only exists when this is used in an Mobly test run.
         self._log_path_base = getattr(logging, 'log_path', '/tmp/logs')
-        self._debug_tag = self.serial
+        self._log_path = os.path.join(self._log_path_base,
+                                      'AndroidDevice%s' % self._serial)
+        self._debug_tag = self._serial
         self.log = AndroidDeviceLoggerAdapter(logging.getLogger(),
                                               {'tag': self.debug_tag})
         self.sl4a = None
@@ -466,10 +479,71 @@ class AndroidDevice(object):
         self.log.extra['tag'] = tag
 
     @property
+    def has_active_service(self):
+        """True if any service is running on the device.
+
+        A service can be a snippet or logcat collection.
+        """
+        return any(
+            [self._snippet_clients, self._adb_logcat_process, self.sl4a])
+
+    @property
     def log_path(self):
         """A string that is the path for all logs collected from this device.
+
+        This by design should not be changed during the test.
         """
-        return os.path.join(self._log_path_base, 'AndroidDevice%s' % self.debug_tag)
+        return self._log_path
+
+    @log_path.setter
+    def log_path(self, new_path):
+        if self.has_active_service:
+            raise DeviceError(
+                self,
+                'Cannot change `log_path` when there is service running.')
+        old_path = self._log_path
+        utils.create_dir(new_path)
+        distutils.dir_util.copy_tree(self._log_path, new_path)
+        shutil.rmtree(self._log_path, ignore_errors=True)
+        self._log_path = new_path
+
+    @property
+    def serial(self):
+        """The serial number used to identify a device.
+
+        This is essentially the value used for adb's `-s` arg, which means it
+        can be a network address or USB bus number.
+        """
+        return self._serial
+
+    def update_serial(self, new_serial):
+        """Updates the serial number of a device.
+
+        The "serial number" used with adb's `-s` arg is not necessarily the
+        actual serial number. For remote devices, it could be a combination of
+        host names and port numbers.
+
+        This is used for when such identifier of remote devices changes during
+        a test. For example, when a remote device reboots, it may come back
+        with a different serial number.
+
+        This is NOT meant for switching the object to represent another device.
+
+        We intentionally did not make it a regular setter of the serial
+        property so people don't accidentally call this without understanding
+        the consequences.
+
+        Args:
+            new_serial: string, the new serial number for the same device.
+        """
+        if self.has_active_service:
+            raise DeviceError(
+                self,
+                'Cannot change device serial number when there is service running.'
+            )
+        if self._debug_tag == self.serial:
+            self._debug_tag = new_serial
+        self._serial = new_serial
 
     def start_services(self, clear_log=True):
         """Starts long running services on the android device, like adb logcat
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -416,8 +416,7 @@ class AndroidDevice(object):
     def __init__(self, serial=''):
         self.serial = serial
         # logging.log_path only exists when this is used in an Mobly test run.
-        log_path_base = getattr(logging, 'log_path', '/tmp/logs')
-        self.log_path = os.path.join(log_path_base, 'AndroidDevice%s' % serial)
+        self._log_path_base = getattr(logging, 'log_path', '/tmp/logs')
         self._debug_tag = self.serial
         self.log = AndroidDeviceLoggerAdapter(logging.getLogger(),
                                               {'tag': self.debug_tag})
@@ -465,6 +464,12 @@ class AndroidDevice(object):
         self.log.info('Logging debug tag set to "%s"', tag)
         self._debug_tag = tag
         self.log.extra['tag'] = tag
+
+    @property
+    def log_path(self):
+        """A string that is the path for all logs collected from this device.
+        """
+        return os.path.join(self._log_path_base, 'AndroidDevice%s' % self.debug_tag)
 
     def start_services(self, clear_log=True):
         """Starts long running services on the android device, like adb logcat
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -150,7 +150,6 @@ def _start_services_on_ads(ads):
     for ad in ads:
         running_ads.append(ad)
         skip_logcat = getattr(ad, KEY_SKIP_LOGCAT, False)
-        print('#######$$$$$$$', ad.serial, skip_logcat)
         if skip_logcat:
             continue
         try:
@@ -503,8 +502,9 @@ class AndroidDevice(object):
                 'Cannot change `log_path` when there is service running.')
         old_path = self._log_path
         utils.create_dir(new_path)
-        distutils.dir_util.copy_tree(self._log_path, new_path)
-        shutil.rmtree(self._log_path, ignore_errors=True)
+        if os.path.exists(old_path):
+            distutils.dir_util.copy_tree(self._log_path, new_path)
+            shutil.rmtree(self._log_path, ignore_errors=True)
         self._log_path = new_path
 
     @property
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -49,9 +49,11 @@ ANDROID_DEVICE_NOT_LIST_CONFIG_MSG = 'Configuration should be a list, abort!'
 # If this is False for a device, errors from that device will be ignored
 # during `create`. Default is True.
 KEY_DEVICE_REQUIRED = 'required'
+DEFAULT_VALUE_DEVICE_REQUIRED = True
 # If True, logcat collection will not be started during `create`.
 # Default is False.
 KEY_SKIP_LOGCAT = 'skip_logcat'
+DEFAULT_VALUE_SKIP_LOGCAT = False
 
 # Default Timeout to wait for boot completion
 DEFAULT_TIMEOUT_BOOT_COMPLETION_SECOND = 15 * 60
@@ -149,13 +151,14 @@ def _start_services_on_ads(ads):
     running_ads = []
     for ad in ads:
         running_ads.append(ad)
-        skip_logcat = getattr(ad, KEY_SKIP_LOGCAT, False)
+        skip_logcat = getattr(ad, KEY_SKIP_LOGCAT, DEFAULT_VALUE_SKIP_LOGCAT)
         if skip_logcat:
             continue
         try:
             ad.start_services()
         except Exception:
-            is_required = getattr(ad, KEY_DEVICE_REQUIRED, True)
+            is_required = getattr(ad, KEY_DEVICE_REQUIRED,
+                                  DEFAULT_VALUE_DEVICE_REQUIRED)
             if is_required:
                 ad.log.exception('Failed to start some services, abort!')
                 destroy(running_ads)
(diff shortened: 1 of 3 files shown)


mobly/controllers/android_device.py
```diff
@@ -142,6 +149,10 @@ def _start_services_on_ads(ads):
     running_ads = []
     for ad in ads:
         running_ads.append(ad)
+        skip_logcat = getattr(ad, KEY_SKIP_LOGCAT, False)
+        print('#######$$$$$$$', ad.serial, skip_logcat)
```
gauravkatoch007: self.log.info ?

mobly/controllers/android_device.py
```diff
@@ -466,6 +477,74 @@ def debug_tag(self, tag):
...
+    @property
+    def serial(self):
+        """The serial number used to identify a device.
+
+        This is essentially the value used for adb's `-s` arg, which means it
+        can be a network address or USB bus number.
+        """
+        return self._serial
+
+    def update_serial(self, new_serial):
```
gauravkatoch007: In function start_adb_logcat of android_device.py 
Please change fname to -
f_name = 'adblog,%s,%s.txt' % (self.model, self.debug_tag)
As otherwise after changing serial of remote device, this would result in another logfile with new serial.

For remote devices, users can setup custom device name with debug_tag, before starting logcat service.

xpconanfan: If the serial changed in the middle of the test, it seems good to have the adb logs split between before and after the change?

xpconanfan: the problem is, debug_tag can change at any time without guarding.
I don't want file names to be dependent on that.
If your device changes serial in the middle of a test, that's kind of what you have to suffer atm.

The alternative is to force all existing users to conform to the same approach described in this PR, which is more costly.

gauravkatoch007: Acknowledged.

k2fong:
Review status: 0 of 3 files reviewed at latest revision, 7 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 50 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyalTpYlrMcthjP7xy1:-KyalTpYlrMcthjP7xy2:b1ekaa9) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L50)):*
> Default is True.
instead of commenting the default, how create a constant to store the default.  Self documenting and can be reuse where the default gets used.

Similar for SKIP_LOGCAT

---

*[mobly/controllers/android_device.py, line 493 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyatXhvXM4e9tn1P_3z:-KyatXhvXM4e9tn1P_4-:b-nafzt) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L493)):*
> This by design should not be changed during the test.
is this statement contradicting the available setter for log_path below?

---

*[mobly/controllers/android_device.py, line 507 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyaoVC-IHjUnaMNT0XX:-KyaoVC-IHjUnaMNT0XY:b-b9q00r) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L507)):*
>             distutils.dir_util.copy_tree(self._log_path, new_path)
>             shutil.rmtree(self._log_path, ignore_errors=True)
nit, consider using the var old_path instead of self._log_path for readability

---

*[mobly/controllers/android_device.py, line 528 at r2](https://reviewable.io:443/reviews/google/mobly/363#-Kyas6qn91Fc14H4XfDQ:-Kyas6qn91Fc14H4XfDR:bme7j45) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L528)):*
>         This is used for when such identifier of remote devices changes during
>         a test. For example, when a remote device reboots, it may come back
>         with a different serial number.
Can serial change anytime?  or is it only after a reboot?  Sorry, I don't know much about this serial switching but its its the latter, consider putting serial update in reboot() so it doesn't feel like a chore that user have to maintain.  Also, is it possible that the device changes serial on its own (not orchestrated by the test) where we might get into a case where there is a mismatch?

---

*[mobly/controllers/android_device.py, line 546 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyatqIgUfAmogoQ02xj:-KyatqIgUfAmogoQ02xk:b-750h3x) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L546)):*
> ```Python
>         if self._debug_tag == self.serial:
>             self._debug_tag = new_serial
>         self._serial = new_serial
> ```

should log_path be update here when serial gets updated?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/363)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 7 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 153 at r1](https://reviewable.io:443/reviews/google/mobly/363#-KyNou8vsJiaCEg8hJhy-r1-153:-KybJNea6_YqYIsRoZkl:b-d3wvuy) ([raw file](https://github.com/google/mobly/blob/e017ee7eb6f4336c55edf7b79e00c9099179076a/mobly/controllers/android_device.py#L153)):*
<details><summary><i>Previously, gauravkatoch007 (Gaurav Katoch) wrote…</i></summary><blockquote>

self.log.info ?
</blockquote></details>

not sure what you mean
there is no `self.log` obj at this point

---

*[mobly/controllers/android_device.py, line 50 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyalTpYlrMcthjP7xy1:-KybLPHgAyFqCv6Rvq5i:b-896fix) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L50)):*

> Default is True.
instead of commenting the default, how create a constant to store the default.  Self documenting and can be reuse where the default gets used.

Similar for SKIP_LOGCAT
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 493 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyatXhvXM4e9tn1P_3z:-KybJT8TN5Cs6SFf67hg:b-896fix) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L493)):*

> This by design should not be changed during the test.
is this statement contradicting the available setter for log_path below?
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 507 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyaoVC-IHjUnaMNT0XX:-KybJoID8spmwydFDnoP:b1fuffa) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L507)):*

>             distutils.dir_util.copy_tree(self._log_path, new_path)
>             shutil.rmtree(self._log_path, ignore_errors=True)
nit, consider using the var old_path instead of self._log_path for readability
</blockquote></details>

I'm trying to avoid calling an attribute within its own setter.

---

*[mobly/controllers/android_device.py, line 528 at r2](https://reviewable.io:443/reviews/google/mobly/363#-Kyas6qn91Fc14H4XfDQ:-KybJv5HMF9xFe2LOlJR:bk9cupo) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L528)):*

>         This is used for when such identifier of remote devices changes during
>         a test. For example, when a remote device reboots, it may come back
>         with a different serial number.
Can serial change anytime?  or is it only after a reboot?  Sorry, I don't know much about this serial switching but its its the latter, consider putting serial update in reboot() so it doesn't feel like a chore that user have to maintain.  Also, is it possible that the device changes serial on its own (not orchestrated by the test) where we might get into a case where there is a mismatch?
</blockquote></details>

reboot does not always update the serial so it doesn't make sense to do this inside reboot.
a serial change is an uncommon case hence extra steps is required to handle it, this is intentional.
we do not want to build magic into reboot.

---

*[mobly/controllers/android_device.py, line 546 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyatqIgUfAmogoQ02xj:-KybKL1H7KbjTNZblHXY:b9hkpdt) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L546)):*

should log_path be update here when serial gets updated?
</blockquote></details>

No, only if the user intentionally want to change the path, which they should handle themselves.
We don't want to build too much magic into these APIs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/363)*

k2fong:
Review status: 0 of 3 files reviewed at latest revision, 6 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 507 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyaoVC-IHjUnaMNT0XX:-Kyc-dDBIJUZ4Ckb6-nH:b2w8fdn) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L507)):*

I'm trying to avoid calling an attribute within its own setter.
</blockquote></details>

the old_path variable created above?  not log_path.  just so its consistent with the if() check

---

*[mobly/controllers/android_device.py, line 546 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyatqIgUfAmogoQ02xj:-Kyc0kMG66yLXNutB2o7:b-da145b) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L546)):*

No, only if the user intentionally want to change the path, which they should handle themselves.
We don't want to build too much magic into these APIs.
</blockquote></details>

If that's the case does the check for active services even needed?  But if we still do, might be worthwhile to add comment as to why that check exist.

---

*[mobly/controllers/android_device.py, line 50 at r3](https://reviewable.io:443/reviews/google/mobly/363#-Kyc1O9ojIhJnExrIa5m:-Kyc1O9ojIhJnExrIa5n:b-nakwxy) ([raw file](https://github.com/google/mobly/blob/0759fb3476f68953b3839c3b2103638a2daf6a8c/mobly/controllers/android_device.py#L50)):*
>  Default is True.
nit, no need for this comment now

---

*[mobly/controllers/android_device.py, line 547 at r3](https://reviewable.io:443/reviews/google/mobly/363#-Kyc0aU1JVJYwxwOvY28:-Kyc0aU1JVJYwxwOvY29:bu3iscz) ([raw file](https://github.com/google/mobly/blob/0759fb3476f68953b3839c3b2103638a2daf6a8c/mobly/controllers/android_device.py#L547)):*
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>
>         if new_serial == self._serial:
>             raise DeviceError(
>                 self, 'Serial is already %s, cannot set to the same value.' %
>                 self._serial)
</details>
will just returning be more graceful then raising exception in the case that they are equal?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/363)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 6 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 507 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyaoVC-IHjUnaMNT0XX:-KyqxpvQr8hv9tXrUNHd:b-896fix) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L507)):*

the old_path variable created above?  not log_path.  just so its consistent with the if() check
</blockquote></details>

Done.

---

*[mobly/controllers/android_device.py, line 546 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyatqIgUfAmogoQ02xj:-KyqyHqomV5MFWpGZmyK:b-80nnlh) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L546)):*

If that's the case does the check for active services even needed?  But if we still do, might be worthwhile to add comment as to why that check exist.
</blockquote></details>

Because a device's serial can not be changed when there are active services.
Not sure how that's related to reboots.

---

*[mobly/controllers/android_device.py, line 547 at r3](https://reviewable.io:443/reviews/google/mobly/363#-Kyc0aU1JVJYwxwOvY28:-KyqxtXB00mKXm2Er3HZ:b83cig2) ([raw file](https://github.com/google/mobly/blob/0759fb3476f68953b3839c3b2103638a2daf6a8c/mobly/controllers/android_device.py#L547)):*

>         if new_serial == self._serial:
>             raise DeviceError(
>                 self, 'Serial is already %s, cannot set to the same value.' %
>                 self._serial)
will just returning be more graceful then raising exception in the case that they are equal?
</blockquote></details>

"updating" a serial means it should be a different one?
I'm trying to convey to people that they should not randomly call `update_serial`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/363)*

k2fong:
Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 546 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyatqIgUfAmogoQ02xj:-KyrNVjOdoUj69u0RrRo:b-4fm9xv) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L546)):*

Because a device's serial can not be changed when there are active services.
Not sure how that's related to reboots.
</blockquote></details>

serial can not change only because of the already open handle that has reference to the prior serial correct?  if so, what if user don't care to change log path after a serial change?  (As an example, the user might have previously set the log path to something that has no serial reference in it)  Should we still restrict user from setting the serial while there are active services?

---

*[mobly/controllers/android_device.py, line 547 at r3](https://reviewable.io:443/reviews/google/mobly/363#-Kyc0aU1JVJYwxwOvY28:-KyrJkFUtRzpvYkx4xW5:b-88mmh) ([raw file](https://github.com/google/mobly/blob/0759fb3476f68953b3839c3b2103638a2daf6a8c/mobly/controllers/android_device.py#L547)):*

"updating" a serial means it should be a different one?
I'm trying to convey to people that they should not randomly call `update_serial`.
</blockquote></details>

Acknowledged.

I don't disagree updating should be a different serial just not sure if raising an exception is necessary, but I could see that its probably not the user's intent to set the same serial so this will catch that and hopefully promote fail fast.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/363)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 546 at r2](https://reviewable.io:443/reviews/google/mobly/363#-KyatqIgUfAmogoQ02xj:-KyrR5eex6_5Pz9eiVmK:brp5lru) ([raw file](https://github.com/google/mobly/blob/766e848ec3f3cf6e21f1974ab75efc205c206f88/mobly/controllers/android_device.py#L546)):*

serial can not change only because of the already open handle that has reference to the prior serial correct?  if so, what if user don't care to change log path after a serial change?  (As an example, the user might have previously set the log path to something that has no serial reference in it)  Should we still restrict user from setting the serial while there are active services?
</blockquote></details>

log_path change is not the only thing affected by serial change.
Almost all ongoing device operations need to be restarted if the serial changes, hence not allowing serial change when there's ongoing service.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/363)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 547 at r3](https://reviewable.io:443/reviews/google/mobly/363#-Kyc0aU1JVJYwxwOvY28:-KyrRLUWb97ZjuWzeXme:b-dr860z) ([raw file](https://github.com/google/mobly/blob/0759fb3476f68953b3839c3b2103638a2daf6a8c/mobly/controllers/android_device.py#L547)):*

Acknowledged.

I don't disagree updating should be a different serial just not sure if raising an exception is necessary, but I could see that its probably not the user's intent to set the same serial so this will catch that and hopefully promote fail fast.
</blockquote></details>

Why would the test logic call `update_serial` when it's the same serial?
Sounds like the user needs to check their test logic ;) 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/363)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[mobly/controllers/android_device.py, line 50 at r3](https://reviewable.io:443/reviews/google/mobly/363#-Kyc1O9ojIhJnExrIa5m:-Kz107ClXKycmDyTOk84:b-y8r4dn) ([raw file](https://github.com/google/mobly/blob/0759fb3476f68953b3839c3b2103638a2daf6a8c/mobly/controllers/android_device.py#L50)):*

>  Default is True.
nit, no need for this comment now
</blockquote></details>

Let's keep it so it's more obvious.

---

*[mobly/controllers/android_device.py, line 547 at r3](https://reviewable.io:443/reviews/google/mobly/363#-Kyc0aU1JVJYwxwOvY28:-Kz10BiPMn6g7b3cxYlw:b-896fix) ([raw file](https://github.com/google/mobly/blob/0759fb3476f68953b3839c3b2103638a2daf6a8c/mobly/controllers/android_device.py#L547)):*

Why would the test logic call `update_serial` when it's the same serial?
Sounds like the user needs to check their test logic ;) 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/363)*

k2fong:
:lgtm:

---

Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/363#-:-Kz129-YWl6HOjh6GmrB:bnfp4nl)*

PR #366, opened 2017-11-21
Update serials for adb and fastboot in `update_serial`.

xpconanfan:
A bug from #363 

@gauravkatoch007

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/366)

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -554,6 +554,8 @@ class AndroidDevice(object):
         if self._debug_tag == self.serial:
             self._debug_tag = new_serial
         self._serial = new_serial
+        self.adb.serial = new_serial
+        self.fastboot.serial = new_serial
 
     def start_services(self, clear_log=True):
         """Starts long running services on the android device, like adb logcat
--- a/tests/mobly/controllers/android_device_test.py
+++ b/tests/mobly/controllers/android_device_test.py
@@ -519,6 +519,8 @@ class AndroidDeviceTest(unittest.TestCase):
         ad.update_serial(2)
         self.assertEqual(ad.serial, 2)
         self.assertEqual(ad.debug_tag, ad.serial)
+        self.assertEqual(ad.adb.serial, ad.serial)
+        self.assertEqual(ad.fastboot.serial, ad.serial)
 
     @mock.patch(
         'mobly.controllers.android_device_lib.adb.AdbProxy',


k2fong:

---

*[mobly/controllers/android_device.py, line 558 at r1](https://reviewable.io:443/reviews/google/mobly/366#-KzV6xiEV8NnqVZxx4Zj:-KzV6xiEV8NnqVZxx4Zk:b-fwfev5) ([raw file](https://github.com/google/mobly/blob/8a528d11279f11c52c3acf07dcfdc772a3215023/mobly/controllers/android_device.py#L558)):*
>         self.adb.serial = new_serial
>         self.fastboot.serial = new_serial
not sure if its the intent of the AdbProxy and Fastboot have serial to change after creation.  Will it be better to create new adb and fastboot object with the new serial instead of just updating serial directly?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/366)*

xpconanfan:
No it makes no difference.
If you look at the constructor of those classes, they do nothing but
setting the serial.
And neither of them hold any state.

k2fong:

---

*[mobly/controllers/android_device.py, line 558 at r1](https://reviewable.io:443/reviews/google/mobly/366#-KzV6xiEV8NnqVZxx4Zj:-KzWMDk9r93A73QRHeVR:bsmbt24) ([raw file](https://github.com/google/mobly/blob/8a528d11279f11c52c3acf07dcfdc772a3215023/mobly/controllers/android_device.py#L558)):*

>         self.adb.serial = new_serial
>         self.fastboot.serial = new_serial
not sure if its the intent of the AdbProxy and Fastboot have serial to change after creation.  Will it be better to create new adb and fastboot object with the new serial instead of just updating serial directly?
</blockquote></details>

yes, I agree, its not a problem currently but doing this could potentially cause problem later if in case adb & fastboot changes in the future where it might.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/366)*

xpconanfan:
The adb and fastboot code have not changed much for two years. Not sure
what kind of situation you're talking about. We intentionally kept these
classes stateless for simplicity.

Can you give examples on when we'd have to change that? Not sure if it's
worth blocking this for that.


On Nov 21, 2017 5:37 PM, "k2fong" <notifications@github.com> wrote:

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved
discussion.
------------------------------

*mobly/controllers/android_device.py, line 558 at r1
<https://reviewable.io:443/reviews/google/mobly/366#-KzV6xiEV8NnqVZxx4Zj:-KzWMDk9r93A73QRHeVR:bsmbt24>
(raw file
<https://github.com/google/mobly/blob/8a528d11279f11c52c3acf07dcfdc772a3215023/mobly/controllers/android_device.py#L558>):*
*Previously, k2fong wrote…*

    self.adb.serial = new_serial
    self.fastboot.serial = new_serial

not sure if its the intent of the AdbProxy and Fastboot have serial to
change after creation. Will it be better to create new adb and fastboot
object with the new serial instead of just updating serial directly?

yes, I agree, its not a problem currently but doing this could potentially
cause problem later if in case adb & fastboot changes in the future where
it might.
------------------------------

*Comments from Reviewable
<https://reviewable.io:443/reviews/google/mobly/366>*

—
You are receiving this because you were assigned.
Reply to this email directly, view it on GitHub
<https://github.com/google/mobly/pull/366#issuecomment-346216455>, or mute
the thread

k2fong:
This serial in these classes are states.  And as for example, I think its just good practice to encapsulate states (and not directly manipulating them).  If they are just simple data classes, I probably will feel different.  And the way I see it, these classes are abstraction of the adb and fastboot conduit, and not simply data classes.  Or if the serial member is intended as a publicly use member, I might be ok with that too.  But its not.   The serial is use only internally.  

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/366#-:-KzWqMygyZLJ37tj30pS:bhdoe55)*

xpconanfan:
Why do you say it's meant for internal use? As you can see the serial is
structured as public var.

And these classes are just simple wrappers. we actually use them even
without serial as well as adb and fastboot can work that way. So they are
intentionally thin and we don't put too much into them.

If you want to make them more traditional classes in Java style, feel free
to send a PR. Refactoring these classes are out of the scope of this PR.


On Nov 21, 2017 8:16 PM, "k2fong" <notifications@github.com> wrote:

This serial in these classes are states. And as for example, I think its
just good practice to encapsulate states (and not directly manipulating
them). If they are just simple data classes, I probably will feel
different. And the way I see it, these classes are abstraction of the adb
and fastboot conduit, and not simply data classes. Or if the serial member
is intended as a publicly use member, I might be ok with that too. But its
not. The serial is use only internally.
------------------------------

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved
discussion.
------------------------------

*Comments from Reviewable
<https://reviewable.io:443/reviews/google/mobly/366#-:-KzWqMygyZLJ37tj30pS:bhdoe55>*

—
You are receiving this because you were assigned.
Reply to this email directly, view it on GitHub
<https://github.com/google/mobly/pull/366#issuecomment-346238729>, or mute
the thread

k2fong:
I do see two other areas that uses the serial but I don't quite agree with their usage, seems more appropriate to get serial from ad instead.

my comment is not restructuring adb nor fastboot.   it was to create new adb and fastboot as such instead of directly setting the updated serial.

```
def update_serial(self, new_serial):
        ...
        self.adb = adb.AdbProxy(new_serial)
        self.fastboot = fastboot.FastbootProxy(new_serial)
```

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/366#-:-KzZg3ekLMR87t7x-BaU:b6fvvhw)*

k2fong:
:lgtm:

per offline chat regarding the suggestion will break the current UT structuring of mocking out the proxies.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/366#-:-KzaApl5Yomjo_ykzTQP:bdcpd67)*

PR #368, opened 2017-11-29
Restructure adb error classes.

xpconanfan:
* AdbTimeoutError should not be a subclass of AdbError as they are
  raised for different occasions.
* Improve log messages.
* Add proper docstrings.

@oprypin

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/368)

### DIFF
--- a/mobly/controllers/android_device_lib/adb.py
+++ b/mobly/controllers/android_device_lib/adb.py
@@ -16,6 +16,7 @@ from builtins import str
 from past.builtins import basestring
 
 import logging
+import pipes
 import psutil
 import subprocess
 import threading
@@ -53,7 +54,7 @@ class AdbError(Error):
 
     def __str__(self):
         return ('Error executing adb cmd "%s". ret: %d, stdout: %s, stderr: %s'
-                ) % (' '.join(self.cmd), self.ret_code, self.stdout,
+                ) % (cli_cmd_to_string(self.cmd), self.ret_code, self.stdout,
                      self.stderr)
 
 
@@ -71,7 +72,7 @@ class AdbTimeoutError(Error):
 
     def __str__(self):
         return 'Timed out executing command "%s" after %ss.' % (
-            ' '.join(self.cmd), self.timeout)
+            cli_cmd_to_string(self.cmd), self.timeout)
 
 
 def list_occupied_adb_ports():
@@ -94,6 +95,18 @@ def list_occupied_adb_ports():
             continue
         used_ports.append(int(tokens[1]))
     return used_ports
+
+
+def cli_cmd_to_string(args):
+    """Converts a cmd arg list to string.
+
+    Args:
+        args: list of strings, the arguments of a command.
+
+    Returns:
+        String representation of the command.
+    """
+    return ' '.join([pipes.quote(arg) for arg in args])
 
 
 class AdbProxy(object):
@@ -155,7 +168,7 @@ class AdbProxy(object):
         (out, err) = proc.communicate()
         ret = proc.returncode
         logging.debug('cmd: %s, stdout: %s, stderr: %s, ret: %s',
-                      ' '.join(args), out, err, ret)
+                      cli_cmd_to_string(args), out, err, ret)
         if ret == 0:
             return out
         else:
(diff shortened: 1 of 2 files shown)


mobly/controllers/android_device_lib/adb.py
```diff
@@ -121,27 +139,27 @@ def _exec_cmd(self, args, shell, timeout):
             AdbError: The adb command exit code is not 0.
             AdbTimeoutError: The adb command timed out.
         """
+        cmd_str = ' '.join(args)
```
oprypin: This changes the way command line arguments are displayed from `['cp', 'my file', 'other file']` to `"cp my file other file"`. The change is not strictly necessary for this pull request, but if you'd like to proceed with it, I think the proper way would be:

```python
def _quote_command(command):
  return ' '.join(pipes.quote(arg) for arg in command)
```
...returning `cp 'my file' 'other file'`

mobly/controllers/android_device_lib/adb.py
```diff
@@ -121,27 +139,27 @@ def _exec_cmd(self, args, shell, timeout):
...
         (out, err) = proc.communicate()
         ret = proc.returncode
-        logging.debug('cmd: %s, stdout: %s, stderr: %s, ret: %s', args, out,
+        logging.debug('cmd: %s, stdout: %s, stderr: %s, ret: %s', cmd_str, out,
                       err, ret)
         if ret == 0:
             return out
         else:
-            raise AdbError(cmd=args, stdout=out, stderr=err, ret_code=ret)
+            raise AdbError(cmd=cmd_str, stdout=out, stderr=err, ret_code=ret)
```
oprypin: The `cmd` member should stay as it was - a list. Changing it to a string breaks API without a real need. I would suggest only changing the way that the string representation of the exception works (or not changing it)

mobly/controllers/android_device_lib/adb.py
```diff
@@ -215,10 +233,7 @@ def instrument(self, package, options=None, runner=None):
         options_string = ' '.join(options_list)
 
         instrumentation_command = 'am instrument -r -w %s %s/%s' % (
-            options_string,
-            package,
-            runner,
-        )
+            options_string, package, runner, )
```
oprypin: This change with the comma at the end seems accidental

mobly/controllers/android_device_lib/adb.py
```diff
@@ -31,8 +31,19 @@
...
-    """Raised when there is an error in adb operations."""
+class Error(Exception):
+    """Base error type for adb proxy module."""
+
+
+class AdbError(Error):
+    """Raised when an adb command encounters an error.
+
+    Args:
+        cmd: string, the adb command that timed out
```
oprypin: Should stay a list probably, in both places.

oprypin:
Overall this is a good change. Thanks for it.

mobly/controllers/android_device_lib/adb.py
xpconanfan: In debugging a failure caused by AdbError, I find it pretty common to try to run the same cmd manually to see the results. Having this as a list makes pasting the cmd difficult.

So the motivation here is so we can directly copy the cmd from the error msg.

`pipes.quote` is used for constructing cmd to be passed to pipes, whereas we purely want something for displaying. So I don't think `pipes` is necessary. Also `pipes` only works on POSIX, which is not good as we need to support windows as well.

Is there an alternative?

mobly/controllers/android_device_lib/adb.py
xpconanfan: Done

mobly/controllers/android_device_lib/adb.py
xpconanfan: Done.
Comma has always been there, just linter merged the lines..

mobly/controllers/android_device_lib/adb.py
xpconanfan: Done

mobly/controllers/android_device_lib/adb.py
oprypin: I have explained why correct quoting is necessary, with it you actually can copy/paste the command unmodified, without fearing for any unusual characters in the command. ` ` `;` `$` `*` `(` `)` `{` `}` `'` `"` - any of these will make the command invalid. And not just for the purposes of copy pasting, but even seeing what's going on.

I believe `pipes.quote` is available on Windows.
https://github.com/python/cpython/blob/2.7/Lib/pipes.py#L267
https://github.com/python/cpython/blob/3.6/Lib/pipes.py#L65

xpconanfan: Done

PR #393, opened 2018-01-30
Set `end_time` when `teardown_test` fails.

xpconanfan:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/393)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -376,6 +376,7 @@ class BaseTestClass(object):
                     raise
                 except Exception as e:
                     logging.exception(e)
+                    tr_record.test_error()
                     tr_record.add_error('teardown_test', e)
                     teardown_test_failed = True
                 else:
--- a/tests/mobly/base_test_test.py
+++ b/tests/mobly/base_test_test.py
@@ -297,6 +297,7 @@ class BaseTestTest(unittest.TestCase):
         # Make sure the full stacktrace of `setup_test` is preserved.
         self.assertTrue('self.setup_test()' in actual_record.stacktrace)
         self.assertIsNone(actual_record.extras)
+        self.assertTrue(actual_record.end_time)
         expected_summary = ("Error 1, Executed 1, Failed 0, Passed 0, "
                             "Requested 1, Skipped 0")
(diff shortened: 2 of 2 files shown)


k2fong:

---

*[mobly/base_test.py, line 379 at r1](https://reviewable.io:443/reviews/google/mobly/393#-L483MHH4H4SNM5JyO6n:-L481ny9nicv-h0LU8KG:b-g2d4d4) ([raw file](https://github.com/google/mobly/blob/4ed86b8db4cec0671cf4bab3656db0262d053c91/mobly/base_test.py#L379)):*
> ```Python
>                 except Exception as e:
>                     logging.exception(e)
>                     tr_record.test_error()
> ```

looks like the underlying tr_record.result set here to TestResultEnums.TEST_RESULT_ERROR from calling test_error() will most like be override by the below test_() calls.  I don't think it hurts any but instead of reusing test_error() as a way to mark the end time, will it be more clear to expose a different api to mark test end?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/393)*

xpconanfan:

---

*[mobly/base_test.py, line 379 at r1](https://reviewable.io:443/reviews/google/mobly/393#-L483MHH4H4SNM5JyO6n:-L48Hi4PCUVTiLyAbEze:byxghyb) ([raw file](https://github.com/google/mobly/blob/4ed86b8db4cec0671cf4bab3656db0262d053c91/mobly/base_test.py#L379)):*

looks like the underlying tr_record.result set here to TestResultEnums.TEST_RESULT_ERROR from calling test_error() will most like be override by the below test_() calls.  I don't think it hurts any but instead of reusing test_error() as a way to mark the end time, will it be more clear to expose a different api to mark test end?
</blockquote></details>

The override is intentional.
The ultimate override actually occurs in `update_record` of the next `finally` block.
So this is how the test record is expected to be used.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/393)*

k2fong:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/393#-:-L4CU7QF1UJ6TKzsufx-:bnfp4nl)*

PR #400, opened 2018-02-13
Fix Windows Android M empty line issue with instrumentation parser

winterfroststrom:
On Android M on a Windows, each line of the instrumentation output has an extra line, this causes the instrumentation parser to erroneously assume that there is a non-empty malformed instrumentation block.

The fix for this is to simply, ignore whitespace only lines for the purposes of calculating empty blocks.

Additionally, when formatting, the extra new lines at the end of everything should be stripped otherwise stuff like the test name becomes malformed.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/400)

### DIFF
--- a/mobly/base_instrumentation_test.py
+++ b/mobly/base_instrumentation_test.py
@@ -411,7 +411,10 @@ class _InstrumentationBlock(object):
             line: string, the raw instrumentation line to append to the value
                 list.
         """
-        self._empty = False
+        # Don't count whitespace only lines.
+        if line.strip():
+            self._empty = False
+
         if self.current_key in self.known_keys:
             self.known_keys[self.current_key].append(line)
         else:
@@ -459,10 +462,10 @@ class _InstrumentationBlockFormatter(object):
         self._unknown_keys = {}
         for key, value in instrumentation_block.known_keys.items():
             self._known_keys[key] = '\n'.join(
-                instrumentation_block.known_keys[key])
+                instrumentation_block.known_keys[key]).rstrip()
         for key, value in instrumentation_block.unknown_keys.items():
             self._unknown_keys[key] = '\n'.join(
-                instrumentation_block.unknown_keys[key])
+                instrumentation_block.unknown_keys[key]).rstrip()
 
     def _get_name(self):
         """Gets the method name of the test method for the instrumentation
--- a/tests/mobly/base_instrumentation_test_test.py
+++ b/tests/mobly/base_instrumentation_test_test.py
@@ -260,6 +260,61 @@ INSTRUMENTATION_CODE: -1
             expected_executed=expected_executed,
             expected_completed_and_passed=True)
 
+    def test_run_instrumentation_test_with_random_whitespace(self):
+        instrumentation_output = """\
+
+INSTRUMENTATION_STATUS: numtests=1
+
+INSTRUMENTATION_STATUS: stream=
+
+com.my.package.test.BasicTest:
+
+INSTRUMENTATION_STATUS: id=AndroidJUnitRunner
+
+INSTRUMENTATION_STATUS: test=basicTest
+
+INSTRUMENTATION_STATUS: class=com.my.package.test.BasicTest
+
+INSTRUMENTATION_STATUS: current=1
+
+INSTRUMENTATION_STATUS_CODE: 1
+
+INSTRUMENTATION_STATUS: numtests=1
+
+INSTRUMENTATION_STATUS: stream=.
+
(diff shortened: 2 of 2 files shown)


xpconanfan:
Pls upload a clean PR without unrelated commits mixed in.

xpconanfan:
Looks ok now.
Had a different view on mobile

xpconanfan:

---

*[mobly/base_instrumentation_test.py, line 415 at r1](https://reviewable.io:443/reviews/google/mobly/400#-L5D0JXLmrZxJJGXV_lT:-L5D0JXLmrZxJJGXV_lU:b-tzu1i9) ([raw file](https://github.com/google/mobly/blob/1b0c350b03818a2927c492829262f3b06c6e9c5c/mobly/base_instrumentation_test.py#L415)):*
> ```Python
>         """
>         # Don't count whitespace only lines.
>         if line.strip():
> ```

It seems like we should have a parsing function that makes sure the input is ok before applying other logic?
Then we don't have to add special `if` and strips in various places, which makes them difficult to track and maintain.

---

*[mobly/base_instrumentation_test.py, line 465 at r1](https://reviewable.io:443/reviews/google/mobly/400#-L5D08JCpK5APDgi69t-:-L5D08JCpK5APDgi69t0:b-13y9wy) ([raw file](https://github.com/google/mobly/blob/1b0c350b03818a2927c492829262f3b06c6e9c5c/mobly/base_instrumentation_test.py#L465)):*
> ```Python
>         for key, value in instrumentation_block.known_keys.items():
>             self._known_keys[key] = '\n'.join(
>                 instrumentation_block.known_keys[key]).rstrip()
> ```

why only `rstrip`?

---

*[tests/mobly/base_instrumentation_test_test.py, line 263 at r1](https://reviewable.io:443/reviews/google/mobly/400#-L5D0-LUCVBrulZF5G7d:-L5D0-LVzR6dfzE03ay2:b-jkzwjc) ([raw file](https://github.com/google/mobly/blob/1b0c350b03818a2927c492829262f3b06c6e9c5c/tests/mobly/base_instrumentation_test_test.py#L263)):*
> ```Python
>             expected_completed_and_passed=True)
> 
>     def test_run_instrumentation_test_with_random_whitespace(self):
> ```

do you need to add extra line everywhere?

Perhaps a couple here and there is sufficient?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/400)*

winterfroststrom:

---

*[mobly/base_instrumentation_test.py, line 415 at r1](https://reviewable.io:443/reviews/google/mobly/400#-L5D0JXLmrZxJJGXV_lT:-L5FJTUC_vkNUGjua0QR:b-jqeyqd) ([raw file](https://github.com/google/mobly/blob/1b0c350b03818a2927c492829262f3b06c6e9c5c/mobly/base_instrumentation_test.py#L415)):*

It seems like we should have a parsing function that makes sure the input is ok before applying other logic?
Then we don't have to add special `if` and strips in various places, which makes them difficult to track and maintain.
</blockquote></details>

No? Like the easy solution to this is to strip all lines and format them, but that would basically break the display of stacktraces and custom formatting.

---

*[mobly/base_instrumentation_test.py, line 465 at r1](https://reviewable.io:443/reviews/google/mobly/400#-L5D08JCpK5APDgi69t-:-L5FJmJgujxqDr3CLQbd:b9qljbc) ([raw file](https://github.com/google/mobly/blob/1b0c350b03818a2927c492829262f3b06c6e9c5c/mobly/base_instrumentation_test.py#L465)):*

why only `rstrip`?
</blockquote></details>

In case someone wants to start with new lines, compare the following:

INSTRUMENTATION_STATUS: stream=blah1 blah2
blah3 blah4

With the following (where you start with a new line):

INSTRUMENTATION_STATUS: stream=
blah1 blah2
blah3 blah4

---

*[tests/mobly/base_instrumentation_test_test.py, line 263 at r1](https://reviewable.io:443/reviews/google/mobly/400#-L5D0-LUCVBrulZF5G7d:-L5FKF9P7L7Z3iOmkYRq:b-q8wae1) ([raw file](https://github.com/google/mobly/blob/1b0c350b03818a2927c492829262f3b06c6e9c5c/tests/mobly/base_instrumentation_test_test.py#L263)):*

do you need to add extra line everywhere?

Perhaps a couple here and there is sufficient?
</blockquote></details>

A couple would be sufficient, but there's one everywhere in the Android M/Windows case

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/400)*

PR #414, opened 2018-02-28
Add command check for logpersist

### LINKED ISSUE #411: logpersist.stop and logpersist.start scripts are not available on all devices
In addition to the root check, the actual logpersist.stop/start/cat script is not available on some devices, which can cause mobly to crash unexpectedly.

winterfroststrom:
Fixes #411

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -984,8 +984,16 @@ class AndroidDevice(object):
         # clear settings first because 'start' doesn't support --clear option
         # before Android N.
         if self.is_rootable:
-            self.adb.shell('logpersist.stop --clear')
-            self.adb.shell('logpersist.start')
+            # Android L and older versions do not have logpersist installed,
+            # so check that the logpersist scripts exists before trying to use
+            # them.
+            if (self.adb.has_shell_command('logpersist.stop')
+                    and self.adb.has_shell_command('logpersist.start')):
+                self.adb.shell('logpersist.stop --clear')
+                self.adb.shell('logpersist.start')
+            else:
+                logging.debug('%s is missing logpersist executables, '
+                              'logs may not get saved.', self)
         f_name = 'adblog,%s,%s.txt' % (self.model, self.serial)
         utils.create_dir(self.log_path)
         logcat_file_path = os.path.join(self.log_path, f_name)
--- a/mobly/controllers/android_device_lib/adb.py
+++ b/mobly/controllers/android_device_lib/adb.py
@@ -211,6 +211,24 @@ class AdbProxy(object):
         """
         return self.shell('getprop %s' % prop_name).decode('utf-8').strip()
 
+    def has_shell_command(self, command):
+        """Checks to see if a given check command exists on the device.
+
+        Args:
+            command: A string that is the name of the command to check.
+
+        Returns:
+            A boolean that is True if the command exists and False otherwise.
+        """
+        try:
+            output = self.shell(['command', '-v',
+                                 command]).decode('utf-8').strip()
+            return command in output
+        except AdbError:
+            # If the command doesn't exist, then 'command -v' can return
+            # an exit code > 1.
+            return False
+
     def forward(self, args=None, shell=False):
         with ADB_PORT_LOCK:
             return self._exec_adb_cmd('forward', args, shell, timeout=None)
--- a/tests/mobly/controllers/android_device_lib/adb_test.py
+++ b/tests/mobly/controllers/android_device_lib/adb_test.py
@@ -39,6 +39,10 @@ MOCK_OPTIONS_INSTRUMENTATION_COMMAND = ('am instrument -r -w -e option1 value1'
                                         '.instrumentation.tests/com.android'
                                         '.common.support.test.runner'
                                         '.AndroidJUnitRunner')
+# Mock Shell Command
+MOCK_SHELL_COMMAND = 'ls'
+MOCK_COMMAND_OUTPUT = '/system/bin/ls'
+MOCK_ADB_SHELL_COMMAND_CHECK = 'adb shell command -v ls'
 
 
 class AdbTest(unittest.TestCase):
@@ -175,6 +179,34 @@ class AdbTest(unittest.TestCase):
         self.assertEqual(adb.cli_cmd_to_string(cmd), '\'"adb"\' \'a b\' c//')
         cmd = 'adb -s meme do something ab_cd'
         self.assertEqual(adb.cli_cmd_to_string(cmd), cmd)
+
+    def test_has_shell_command_called_correctly(self):
+        with mock.patch.object(adb.AdbProxy, '_exec_cmd') as mock_exec_cmd:
+            adb.AdbProxy().has_shell_command(MOCK_SHELL_COMMAND)
+            mock_exec_cmd.assert_called_once_with(
+                ['adb', 'shell', 'command', '-v', MOCK_SHELL_COMMAND],
+                shell=False,
+                timeout=None)
+
+    def test_has_shell_command_with_existing_command(self):
+        with mock.patch.object(adb.AdbProxy, '_exec_cmd') as mock_exec_cmd:
+            mock_exec_cmd.return_value = MOCK_COMMAND_OUTPUT
+            self.assertTrue(
+                adb.AdbProxy().has_shell_command(MOCK_SHELL_COMMAND))
+
+    def test_has_shell_command_with_missing_command_on_older_devices(self):
+        with mock.patch.object(adb.AdbProxy, '_exec_cmd') as mock_exec_cmd:
+            mock_exec_cmd.side_effect = adb.AdbError(
+                MOCK_ADB_SHELL_COMMAND_CHECK, '', '', 0)
+            self.assertFalse(
+                adb.AdbProxy().has_shell_command(MOCK_SHELL_COMMAND))
+
+    def test_has_shell_command_with_missing_command_on_newer_devices(self):
+        with mock.patch.object(adb.AdbProxy, '_exec_cmd') as mock_exec_cmd:
+            mock_exec_cmd.side_effect = adb.AdbError(
+                MOCK_ADB_SHELL_COMMAND_CHECK, '', '', 1)
+            self.assertFalse(
+                adb.AdbProxy().has_shell_command(MOCK_SHELL_COMMAND))
 
 
 if __name__ == '__main__':
--- a/tests/mobly/controllers/android_device_test.py
+++ b/tests/mobly/controllers/android_device_test.py
@@ -299,7 +299,7 @@ class AndroidDeviceTest(unittest.TestCase):
         """Verifies that the serial is a primitive string type and serializable.
         """
         ad = android_device.AndroidDevice(serial=1)
-        # In py2, checks that ad.serial is not the backported py3 str type, 
+        # In py2, checks that ad.serial is not the backported py3 str type,
         # which is not dumpable by yaml in py2.
         # In py3, new_str is equivalent to str, so this check is not
         # appropirate in py3.
@@ -626,6 +626,54 @@ class AndroidDeviceTest(unittest.TestCase):
         self.assertEqual(actual_cat, ''.join(MOCK_ADB_LOGCAT_CAT_RESULT))
         # Stops adb logcat.
         ad.stop_adb_logcat()
+
+    @mock.patch(
+        'mobly.controllers.android_device_lib.adb.AdbProxy',
+        return_value=mock.MagicMock())
+    @mock.patch(
+        'mobly.controllers.android_device_lib.fastboot.FastbootProxy',
+        return_value=mock_android_device.MockFastbootProxy('1'))
+    @mock.patch(
+        'mobly.utils.start_standing_subprocess', return_value='process')
+    @mock.patch('mobly.utils.stop_standing_subprocess')
+    def test_AndroidDevice_start_adb_logcat_with_logpersist(
+            self, stop_proc_mock, start_proc_mock, MockFastboot, MockAdbProxy):
+        mock_serial = '1'
+        mock_adb_proxy = MockAdbProxy.return_value
+        mock_adb_proxy.getprop.return_value = 'userdebug'
+        mock_adb_proxy.has_shell_command.side_effect = lambda command: {
+            'logpersist.start': True,
+            'logpersist.stop': True, }[command]
+        ad = android_device.AndroidDevice(serial=mock_serial)
+        ad.start_adb_logcat()
+        mock_adb_proxy.shell.assert_has_calls([
+            mock.call('logpersist.stop --clear'),
+            mock.call('logpersist.start'),
+        ])
+
+    @mock.patch(
+        'mobly.controllers.android_device_lib.adb.AdbProxy',
+        return_value=mock.MagicMock())
+    @mock.patch(
+        'mobly.controllers.android_device_lib.fastboot.FastbootProxy',
+        return_value=mock_android_device.MockFastbootProxy('1'))
+    @mock.patch(
+        'mobly.utils.start_standing_subprocess', return_value='process')
+    @mock.patch('mobly.utils.stop_standing_subprocess')
+    def test_AndroidDevice_start_adb_logcat_with_missing_logpersist(
+            self, stop_proc_mock, start_proc_mock, MockFastboot, MockAdbProxy):
+        mock_serial = '1'
+        mock_adb_proxy = MockAdbProxy.return_value
+        mock_adb_proxy.getprop.return_value = 'userdebug'
+        mock_adb_proxy.has_shell_command.side_effect = lambda command: {
+            'logpersist.start': False,
+            'logpersist.stop': False, }[command]
+        ad = android_device.AndroidDevice(serial=mock_serial)
+        ad.start_adb_logcat()
+        mock_adb_shell_calls = mock_adb_proxy.shell.mock_calls
+        self.assertNotIn(
+            mock.call('logpersist.stop --clear'), mock_adb_shell_calls)
+        self.assertNotIn(mock.call('logpersist.start'), mock_adb_shell_calls)
 
     @mock.patch(
         'mobly.controllers.android_device_lib.adb.AdbProxy',


xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6Z_l9jtRQrTZmfs5Hm:bt8c3ov) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*
> ```Python
>                 self.adb.shell('logpersist.start')
>             else:
>                 logging.error('%s is missing logpersist executables, '
> ```

this is not accurate.

If either `logpersist.start` and `logpersist.stop` does not exist, does not mean `logpersist` itself doesn't exist.

Should we change to only check `logpersist`?

Or have you seen cases where `logpersist.xxx` exists yet `logpersist.yyy` does not?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

xpconanfan:

---

*[mobly/controllers/android_device_lib/adb.py, line 223 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6ZaaWazQdBXBtppyJq:-L6ZaaWazQdBXBtppyJr:b-zhlaii) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device_lib/adb.py#L223)):*
> ```Python
>             A boolean that is True if the command exists and False otherwise.
>         """
>         try:
> ```

can we log.debug the stdout/stderr before returning  `False` in both paths?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

winterfroststrom:

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6ZdhEST7_mtuU1bchn:b5g8lzi) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

this is not accurate.

If either `logpersist.start` and `logpersist.stop` does not exist, does not mean `logpersist` itself doesn't exist.

Should we change to only check `logpersist`?

Or have you seen cases where `logpersist.xxx` exists yet `logpersist.yyy` does not?
</blockquote></details>

logperist is the name of the binary in AOSP, but it's not actually installed as such.
The build rule that installs it actually installs it as logpersist.start and then symlinks logpersist.stop and logpersist.cat
This means that in a correct configuration, they should all be present or none of them should be present

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

winterfroststrom:

---

*[mobly/controllers/android_device_lib/adb.py, line 223 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6ZaaWazQdBXBtppyJq:-L6ZeLdM1cLyLj6YGNH_:b-db70b3) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device_lib/adb.py#L223)):*

can we log.debug the stdout/stderr before returning  `False` in both paths?
</blockquote></details>

Doesn't this already happen in self.shell? Or do you mean something else?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

winterfroststrom:

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6ZeXeXdsfi5GQgvpTN:b-85wqok) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

logperist is the name of the binary in AOSP, but it's not actually installed as such.
The build rule that installs it actually installs it as logpersist.start and then symlinks logpersist.stop and logpersist.cat
This means that in a correct configuration, they should all be present or none of them should be present

</blockquote></details>

Link for reference: https://android.googlesource.com/platform/system/core/+/master/logcat/Android.bp

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://beta.reviewable.io/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6Zf4hN8_leCrqWsMNV:b-w9szi1) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

Link for reference: https://android.googlesource.com/platform/system/core/+/master/logcat/Android.bp
</blockquote></details>

then we only need to check one of them?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/google/mobly/414)*

winterfroststrom:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://beta.reviewable.io/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6ZfY-F61pzTK0TeRQi:b-fv8glb) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

then we only need to check one of them?
</blockquote></details>

Technically, yes. But I would prefer not to for safety reasons in case of bad configurations.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/google/mobly/414)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://beta.reviewable.io/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6ZgDrE2okK_ZZEHwYH:bdeib2a) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

Technically, yes. But I would prefer not to for safety reasons in case of bad configurations.
</blockquote></details>

If we are concerned about the possibility of one of them missing, we should log accordingly; if not, we should only need to check one.

You can't have the cake and eat it too :P

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/google/mobly/414)*

winterfroststrom:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6ZjvkqsG4J8ltiC78U:b-7y4qx0) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

If we are concerned about the possibility of one of them missing, we should log accordingly; if not, we should only need to check one.

You can't have the cake and eat it too :P
</blockquote></details>

Okay, added better logging

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6cLJbNAb4sBEl5lsBa:b-xb0j3m) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

Okay, added better logging
</blockquote></details>

This doesn't look very pythonic...
Can we simply call the cmd and catch `AdbError` instead of doing the checks and stuff?

```
try:
  self.adb.shell('logpersist.stop --clear')
  self.adb.shell('logpersist.start')
except adb.AdbError:
  log.warning(blah)
```

`has_shell_command` can be useful for other cases, so we can keep it.
But it is not needed in this case.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

winterfroststrom:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6dC9rAnA4XeK-VvewN:b-k5dphr) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

This doesn't look very pythonic...
Can we simply call the cmd and catch `AdbError` instead of doing the checks and stuff?

```
try:
  self.adb.shell('logpersist.stop --clear')
  self.adb.shell('logpersist.start')
except adb.AdbError:
  log.warning(blah)
```

`has_shell_command` can be useful for other cases, so we can keep it.
But it is not needed in this case.
</blockquote></details>

Would this be problematic on Android M and older devices since they don't set ret code?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6dDBA1rodxjLZ_JGS8:b-2csjj3) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

Would this be problematic on Android M and older devices since they don't set ret code?
</blockquote></details>

Why would it be a problem?
`logpersist` doesn't exist on those builds anyway right?

This way the behavior largely stays the same except it doesn't crash the test when `logpersist` doesn't exist on M+ builds.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

winterfroststrom:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6dIlNdJKpna-BM-67y:baqo3by) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

Why would it be a problem?
`logpersist` doesn't exist on those builds anyway right?

This way the behavior largely stays the same except it doesn't crash the test when `logpersist` doesn't exist on M+ builds.
</blockquote></details>

Logpersist exists on Android M, it's Android L and before where it's not present.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6dIzf3qKbNXpgZCkXa:bvnz00u) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

Logpersist exists on Android M, it's Android L and before where it's not present.
</blockquote></details>

Still don't see what the concern is. Can you be more specific

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

winterfroststrom:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6dJ3pkewBhh6RgsmA8:b-n5bd5b) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

Still don't see what the concern is. Can you be more specific
</blockquote></details>

If an Android M device is missing logpersist, which it's supposed to have, nothing will get logged.
So, you'd only find out about the issue after realizing that your logs are missing, which seems bad.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6dK6FQ7MUXpBm_xvRj:bn7k7gq) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

If an Android M device is missing logpersist, which it's supposed to have, nothing will get logged.
So, you'd only find out about the issue after realizing that your logs are missing, which seems bad.
</blockquote></details>

maybe we can do a specific check of `logpersist.start`, log a waning, and leave the rest to try...except?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

winterfroststrom:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6dKvLDzB6tTb5gy505:b-syctjm) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

maybe we can do a specific check of `logpersist.start`, log a waning, and leave the rest to try...except?
</blockquote></details>

I'm not sure why you don't like this implementation of the check, but okay, I'm will to do what you just suggested

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6dLb0BjdV1Kosc-1V9:bghm7bp) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

I'm not sure why you don't like this implementation of the check, but okay, I'm will to do what you just suggested
</blockquote></details>

It's a python thing
https://docs.quantifiedcode.com/python-anti-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html

Not a big fan of cascading `if` statements "asking for permission"...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

winterfroststrom:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 995 at r2](https://reviewable.io:443/reviews/google/mobly/414#-L6Z_l9jtRQrTZmfs5Hl:-L6dRCseAaO0c31X6pBQ:b-896fix) ([raw file](https://github.com/google/mobly/blob/59aff146a418e810c72f3fc4b8fc8c37ad177902/mobly/controllers/android_device.py#L995)):*

It's a python thing
https://docs.quantifiedcode.com/python-anti-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html

Not a big fan of cascading `if` statements "asking for permission"...
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

winterfroststrom:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[mobly/controllers/android_device.py, line 983 at r3](https://reviewable.io:443/reviews/google/mobly/414#-L6dRlXZSVS_Gvia9C3X:-L6dTniPwiDvhM-hmyzk:b-896fix) ([raw file](https://github.com/google/mobly/blob/dd2da82ecbef2bb490cd47eed39ef9e414ff7f9e/mobly/controllers/android_device.py#L983)):*

since it's pretty self-contained, can we factor the logpersist stuff into its own function?
like `_enable_logpersist`.

Then we can directly `return` and get rid of some nesting ;)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/mobly/controllers/android_device_test.py, line 690 at r4](https://reviewable.io:443/reviews/google/mobly/414#-L6dZ757Ph_50ve3dDRP:-L6dZ757Ph_50ve3dDRQ:bc1ut1h) ([raw file](https://github.com/google/mobly/blob/5a955dd5e30a8928af7d46aad8974e38f49e939d/tests/mobly/controllers/android_device_test.py#L690)):*
> ```Python
>         mock_adb_proxy.shell.side_effect = adb_shell_helper
>         ad = android_device.AndroidDevice(serial=mock_serial)
>         ad.start_adb_logcat()
> ```

we can actually call the logpersist method in unit tests now right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

winterfroststrom:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/mobly/controllers/android_device_test.py, line 690 at r4](https://reviewable.io:443/reviews/google/mobly/414#-L6dZ757Ph_50ve3dDRP:-L6d_g_ciuAZik2cOIUo:b-896fix) ([raw file](https://github.com/google/mobly/blob/5a955dd5e30a8928af7d46aad8974e38f49e939d/tests/mobly/controllers/android_device_test.py#L690)):*

we can actually call the logpersist method in unit tests now right?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414)*

xpconanfan:
:lgtm:

---

Reviewed 2 of 4 files at r1, 1 of 1 files at r4, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/414#-:-L6dceIP5sRZcSRSbOw1:bnfp4nl)*

PR #416, opened 2018-03-06
Fix key value unpacking issue.

winterfroststrom:
Tested by reproducing with unit test first, then fixing parser for unit test

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/416)

### DIFF
--- a/mobly/base_instrumentation_test.py
+++ b/mobly/base_instrumentation_test.py
@@ -390,7 +390,7 @@ class _InstrumentationBlock(object):
             key_line,
         )
         if '=' in key_value:
-            (key, value) = key_value.split('=')
+            (key, value) = key_value.split('=', 1)
             self.current_key = key
             if key in self.known_keys:
                 self.known_keys[key].append(value)
--- a/tests/mobly/base_instrumentation_test_test.py
+++ b/tests/mobly/base_instrumentation_test_test.py
@@ -1049,7 +1049,7 @@ INSTRUMENTATION_CODE: -1"""
         parsed_line = "INSTRUMENTATION_STATUS: stack=%s" % value
         block = _InstrumentationBlock()
         block.set_key(_InstrumentationStructurePrefixes.STATUS, parsed_line)
-        self.assertIn(value.rstrip(),
+        self.assertIn(value,
                       block.known_keys[_InstrumentationKnownStatusKeys.STACK])
 
 
--- a/tests/mobly/base_instrumentation_test_test.py
+++ b/tests/mobly/base_instrumentation_test_test.py
@@ -19,6 +19,9 @@ import tempfile
 
 from future.tests.base import unittest
 
+from mobly.base_instrumentation_test import _InstrumentationBlock
+from mobly.base_instrumentation_test import _InstrumentationKnownStatusKeys
+from mobly.base_instrumentation_test import _InstrumentationStructurePrefixes
 from mobly.base_instrumentation_test import BaseInstrumentationTestClass
 from mobly import config_parser
 from mobly import signals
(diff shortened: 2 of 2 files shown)


xpconanfan:

---

*[tests/mobly/base_instrumentation_test_test.py, line 318 at r1](https://reviewable.io:443/reviews/google/mobly/416#-L6xJuZphPMTqsfErs6A:-L6xJuZphPMTqsfErs6B:btiz2cg) ([raw file](https://github.com/google/mobly/blob/99c897c0d1a1b357ff21f263c5db29877df93752/tests/mobly/base_instrumentation_test_test.py#L318)):*
> ```Python
>             expected_completed_and_passed=True)
> 
>     def test_run_instrumentation_test_with_equals_sign_key_value_pair(self):
> ```

this is getting a bit out of hand...

Ideally the unit test should be very easy to read and the intention should be clear to readers.
If we are trying to cover the split '=' once case, the unit test should be clear on that.
Currently the actual part we want to test is buried in this giant string.

Can we improve the testability of the logic and the readability of the test cases?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/416)*

xpconanfan:

---

*[tests/mobly/base_instrumentation_test_test.py, line 318 at r1](https://reviewable.io:443/reviews/google/mobly/416#-L6xJuZphPMTqsfErs6A:-L6xKh2hC54do64EfiGl:bo4s0p1) ([raw file](https://github.com/google/mobly/blob/99c897c0d1a1b357ff21f263c5db29877df93752/tests/mobly/base_instrumentation_test_test.py#L318)):*

this is getting a bit out of hand...

Ideally the unit test should be very easy to read and the intention should be clear to readers.
If we are trying to cover the split '=' once case, the unit test should be clear on that.
Currently the actual part we want to test is buried in this giant string.

Can we improve the testability of the logic and the readability of the test cases?
</blockquote></details>

Maybe we can add a test just for `set_key` instead of the entire runner here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/416)*

winterfroststrom:

---

*[tests/mobly/base_instrumentation_test_test.py, line 318 at r1](https://reviewable.io:443/reviews/google/mobly/416#-L6xJuZphPMTqsfErs6A:-L6xP_9uYvUi-de0h9eY:b48c31s) ([raw file](https://github.com/google/mobly/blob/99c897c0d1a1b357ff21f263c5db29877df93752/tests/mobly/base_instrumentation_test_test.py#L318)):*

Maybe we can add a test just for `set_key` instead of the entire runner here.
</blockquote></details>

I'm willing to only test the set_key method because the test runner is covered by the other tests.

I'm not really willing to refactor the other tests because then there's no real guarantee that the tests are even correct.
I would be willing to add tests against the underlying structures though
although any change to the state management would probably end up invalidating most of them.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/416)*

xpconanfan:

---

*[tests/mobly/base_instrumentation_test_test.py, line 318 at r1](https://reviewable.io:443/reviews/google/mobly/416#-L6xJuZphPMTqsfErs6A:-L70dg7APICV7A76vUN_:b9vd6ul) ([raw file](https://github.com/google/mobly/blob/99c897c0d1a1b357ff21f263c5db29877df93752/tests/mobly/base_instrumentation_test_test.py#L318)):*
> there's no real guarantee that the tests are even correct
This is concerning...
Should we remove these tests that may not be correct?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/416)*

winterfroststrom:

---

*[tests/mobly/base_instrumentation_test_test.py, line 318 at r1](https://reviewable.io:443/reviews/google/mobly/416#-L6xJuZphPMTqsfErs6A:-L70kr8H8s0JJXIMZpu_:b-kkxs8v) ([raw file](https://github.com/google/mobly/blob/99c897c0d1a1b357ff21f263c5db29877df93752/tests/mobly/base_instrumentation_test_test.py#L318)):*

> there's no real guarantee that the tests are even correct
This is concerning...
Should we remove these tests that may not be correct?
</blockquote></details>

You misunderstand, what I mean is that tests which only test the interfaces between the individual parser components could very easily become invalided by changes, especially those which fundamentally change parser state management such as supporting streamed execution (at least some of the other approaches) or concurrent execution. The current tests are guaranteed to be correct because they simulate the actual parsing.

So, I mean yes, the current tests are pretty large and hard to read for edge cases, but that's because they're behaviour-based integration tests rather than unit tests. So, if I were to delete the existing tests and replace them with all the equivalent unit tests, then I think I could easily end up in a situation where I either have to delete and re-implement most of the unit tests or they fail to test real parser execution, which is why I went with this approach for testing to begin with aside from that I thought this was clearer since it's fairly obvious what user-facing behaviour each test is testing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/416)*

winterfroststrom:

---

*[tests/mobly/base_instrumentation_test_test.py, line 318 at r1](https://reviewable.io:443/reviews/google/mobly/416#-L6xJuZphPMTqsfErs6A:-L70uxT6XaTz_nXIrYxs:bq6usx0) ([raw file](https://github.com/google/mobly/blob/99c897c0d1a1b357ff21f263c5db29877df93752/tests/mobly/base_instrumentation_test_test.py#L318)):*

You misunderstand, what I mean is that tests which only test the interfaces between the individual parser components could very easily become invalided by changes, especially those which fundamentally change parser state management such as supporting streamed execution (at least some of the other approaches) or concurrent execution. The current tests are guaranteed to be correct because they simulate the actual parsing.

So, I mean yes, the current tests are pretty large and hard to read for edge cases, but that's because they're behaviour-based integration tests rather than unit tests. So, if I were to delete the existing tests and replace them with all the equivalent unit tests, then I think I could easily end up in a situation where I either have to delete and re-implement most of the unit tests or they fail to test real parser execution, which is why I went with this approach for testing to begin with aside from that I thought this was clearer since it's fairly obvious what user-facing behaviour each test is testing.
</blockquote></details>

To clarify, I'm not against adding all of the equivalent unit tests, but I didn't implement them as such originally because I expected a high amount of near-future churn on them. And, I wouldn't trust them without corresponding integration tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/416)*

xpconanfan:

---

*[tests/mobly/base_instrumentation_test_test.py, line 318 at r1](https://reviewable.io:443/reviews/google/mobly/416#-L6xJuZphPMTqsfErs6A:-L71CCTVxTuRJYbrHQR6:b-88mvqy) ([raw file](https://github.com/google/mobly/blob/99c897c0d1a1b357ff21f263c5db29877df93752/tests/mobly/base_instrumentation_test_test.py#L318)):*

To clarify, I'm not against adding all of the equivalent unit tests, but I didn't implement them as such originally because I expected a high amount of near-future churn on them. And, I wouldn't trust them without corresponding integration tests.
</blockquote></details>

Ok.
For this particular case, can we have a more target unit test instead of an integration test like this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/416)*

winterfroststrom:

---

*[tests/mobly/base_instrumentation_test_test.py, line 318 at r1](https://reviewable.io:443/reviews/google/mobly/416#-L6xJuZphPMTqsfErs6A:-L71Fn8TF2s5_JnNzvwl:bsb2e0s) ([raw file](https://github.com/google/mobly/blob/99c897c0d1a1b357ff21f263c5db29877df93752/tests/mobly/base_instrumentation_test_test.py#L318)):*

Ok.
For this particular case, can we have a more target unit test instead of an integration test like this?
</blockquote></details>

I'm not sure I understand, I already changed it to a unit test?
Or do you mean that you want me to trim down the input string still to just the multi-split case?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/416)*

xpconanfan:

---

*[tests/mobly/base_instrumentation_test_test.py, line 318 at r1](https://reviewable.io:443/reviews/google/mobly/416#-L6xJuZphPMTqsfErs6A:-L71Rs-b7ImI5SaOpm6V:b-7e8phe) ([raw file](https://github.com/google/mobly/blob/99c897c0d1a1b357ff21f263c5db29877df93752/tests/mobly/base_instrumentation_test_test.py#L318)):*

I'm not sure I understand, I already changed it to a unit test?
Or do you mean that you want me to trim down the input string still to just the multi-split case?
</blockquote></details>

I don't think this is a unit test?
Can we only test the parsing part instead of passing in this giant input that is hard to tell what key splitting is being tested?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/416)*

winterfroststrom:

---

*[tests/mobly/base_instrumentation_test_test.py, line 318 at r1](https://reviewable.io:443/reviews/google/mobly/416#-L6xJuZphPMTqsfErs6A:-L71TqdVT8mRTbtK9S4o:bvxly2x) ([raw file](https://github.com/google/mobly/blob/99c897c0d1a1b357ff21f263c5db29877df93752/tests/mobly/base_instrumentation_test_test.py#L318)):*

I don't think this is a unit test?
Can we only test the parsing part instead of passing in this giant input that is hard to tell what key splitting is being tested?
</blockquote></details>

then you mean the latter, so done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/416)*

xpconanfan:

---

*[tests/mobly/base_instrumentation_test_test.py, line 1052 at r2](https://reviewable.io:443/reviews/google/mobly/416#-L724h8wukuX7zAku5W_:-L724h8wukuX7zAku5Wa:bfgvg8p) ([raw file](https://github.com/google/mobly/blob/701ec73b67604204cf350219bdf7cb51c7d7689c/tests/mobly/base_instrumentation_test_test.py#L1052)):*
> ```Python
>         block = _InstrumentationBlock()
>         block.set_key(_InstrumentationStructurePrefixes.STATUS, parsed_line)
>         self.assertIn(value.rstrip(),
> ```

why do we need `rstrip` here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/416)*

winterfroststrom:

---

*[tests/mobly/base_instrumentation_test_test.py, line 1052 at r2](https://beta.reviewable.io/reviews/google/mobly/416#-L724h8wukuX7zAku5W_:-L725idV3RuYj20aQZ7N:bepiyd9) ([raw file](https://github.com/google/mobly/blob/701ec73b67604204cf350219bdf7cb51c7d7689c/tests/mobly/base_instrumentation_test_test.py#L1052)):*

why do we need `rstrip` here?
</blockquote></details>

Removed.

This is because set_key does an rstrip, but it's not needed in this case since there are no lines that it'll strip..

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/google/mobly/416)*

winterfroststrom:

---

*[tests/mobly/base_instrumentation_test_test.py, line 1052 at r2](https://beta.reviewable.io/reviews/google/mobly/416#-L724h8wukuX7zAku5W_:-L727W5hC7pPNrW7Awf7:b-kcl4y3) ([raw file](https://github.com/google/mobly/blob/701ec73b67604204cf350219bdf7cb51c7d7689c/tests/mobly/base_instrumentation_test_test.py#L1052)):*

Removed.

This is because set_key does an rstrip, but it's not needed in this case since there are no lines that it'll strip..
</blockquote></details>

Specifically, _remove_structure_prefix does a strip

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/google/mobly/416)*

xpconanfan:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly/416#-:-L7SPnGpXyrIe6qycCQR:bnfp4nl)*

PR #422, opened 2018-04-06
disable hidden API blacklist before starting apps

mswest46:
Note that the blacklist will only affect hidden API usage in P, and the workaround will only disable the blacklist on debug builds. Snippets that use hidden APIs on user P builds will still error in runtime.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/422)

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -877,6 +877,7 @@ class AndroidDevice(object):
                     'Snippet package "%s" has already been loaded under name'
                     ' "%s".' % (package, client_name))
         client = snippet_client.SnippetClient(package=package, ad=self)
+        self._maybe_disable_hidden_api_blacklist()
         try:
             client.start_app_and_connect()
         except snippet_client.AppStartPreCheckError:
@@ -907,6 +908,7 @@ class AndroidDevice(object):
         EventDispatcher obj (self.ed) with the other connection.
         """
         self.sl4a = sl4a_client.Sl4aClient(ad=self)
+        self._maybe_disable_hidden_api_blacklist()
         self.sl4a.start_app_and_connect()
         # Unpack the 'ed' attribute for compatibility.
         self.ed = self.sl4a.ed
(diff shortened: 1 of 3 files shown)


xpconanfan:

---

*[mobly/controllers/android_device_lib/sl4a_client.py, line 66 at r1](https://beta.reviewable.io/reviews/google/mobly/422#-L9RM6O42eD1ggMBPEVm:-L9RM6O42eD1ggMBPEVn:bco9p7e) ([raw file](https://github.com/google/mobly/blob/973ecb3a36a5a8f37168c84722e01a2a720c72cb/mobly/controllers/android_device_lib/sl4a_client.py#L66)):*
> ```Python
> 
>         # sl4a uses hidden APIs, so disable hidden API blacklist.
>         self._adb.shell(
> ```

Can we add this to the base client instead of two different places?
Or make a util function and call in both places.
Also, we should only call this for P+ builds that are rootable.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/google/mobly/422)*

mswest46:

---

*[mobly/controllers/android_device.py, line 1202 at r2](https://beta.reviewable.io/reviews/google/mobly/422#-L9RmP0BAyyVB9RrzA9x:-L9RmP0BAyyVB9RrzA9y:bt4fxva) ([raw file](https://github.com/google/mobly/blob/c04a7644a711debf41823dfb76d510032ecee45a/mobly/controllers/android_device.py#L1202)):*
> ```Python
>             self.adb.reboot()
> 
>     def _is_at_least_P(self):
> ```

I'm mimicking the logic in androidx.core.os.BuildCompat#isAtLeastP(). Seems unnecessarily complicated so let me know if there's a better way.

---

*[mobly/controllers/android_device.py, line 1204 at r2](https://beta.reviewable.io/reviews/google/mobly/422#-L9RmzMH0mjs5fAyOw73:-L9RmzMH0mjs5fAyOw74:b-998tjo) ([raw file](https://github.com/google/mobly/blob/c04a7644a711debf41823dfb76d510032ecee45a/mobly/controllers/android_device.py#L1204)):*
> ```Python
>     def _is_at_least_P(self):
>         """Checks if the device is running on version at least P."""
>         version_codename = self.adb.getprop('ro.build.version.codename')
> ```

can also put this in the build_info attribute if you want

---

*[mobly/controllers/android_device_lib/sl4a_client.py, line 66 at r1](https://beta.reviewable.io/reviews/google/mobly/422#-L9RM6O42eD1ggMBPEVm:-L9RlCzm6E5OJFKTbDzY:bo4hy2e) ([raw file](https://github.com/google/mobly/blob/973ecb3a36a5a8f37168c84722e01a2a720c72cb/mobly/controllers/android_device_lib/sl4a_client.py#L66)):*

Can we add this to the base client instead of two different places?
Or make a util function and call in both places.
Also, we should only call this for P+ builds that are rootable.
</blockquote></details>

Have moved it to android_device. If that's no good let me know.
Have also made it only for P+ builds. FYI the shell command won't make a difference on <P builds, so it's not strictly necessary to do the P+ logic. But your call.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/google/mobly/422)*

PR #432, opened 2018-04-10
Fix instrumentation output on python3.

### LINKED ISSUE #431: Instrumentation parser output is concatenated in python3.
Python3 outputs newline characters as literals for byte strings.
This causes an issue because `run_instrumentation_test` outputs the raw instrumentation output, which ends up being concatenated since the newlines doesn't properly display with python3.

When fixing https://github.com/google/mobly/issues/429, I did not actually run against a real device with python3 (only ran unit tests with python 2 and 3), which meant that I did not actually make sure the output looked correct.

winterfroststrom:
Fixes #431

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/432)

### DIFF
--- a/mobly/base_instrumentation_test.py
+++ b/mobly/base_instrumentation_test.py
@@ -927,7 +927,7 @@ class BaseInstrumentationTestClass(base_test.BaseTestClass):
             package=package,
             options=options,
             runner=runner,
-        )
+        ).decode('utf-8')
         logging.info('Outputting instrumentation test log...')
         logging.info(instrumentation_output)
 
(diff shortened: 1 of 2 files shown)


xpconanfan:
Is there any test we can add for this?

winterfroststrom:
I did write a unit test for this, but I originally decided against including it because it's kinda specific.

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/google/mobly/432#-:-L9l4tLx5kob-0O5JS3w:bgipfst)*

PR #439, opened 2018-04-20
Change UserData timestamp from a float to an int

matty3:
Changed UserData 'timestamp' to use the same util as 'Begin Time' and 'End Time'.

Currently, the UserData 'timestamp' looks like so:
`timestamp: 1524253402.126622`
This changes it so it will look like:
`timestamp: 1524253402`

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/439)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -26,6 +26,7 @@ from mobly import expects
 from mobly import records
 from mobly import signals
 from mobly import runtime_test_info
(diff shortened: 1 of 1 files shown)


matty3:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/google/mobly/439)*

PR #447, opened 2018-05-11
Refactor format adb command

winterfroststrom:


winterfroststrom:

---

*[mobly/controllers/android_device_lib/adb.py, line 184 at r1](https://reviewable.io/reviews/google/mobly/447#-LCWYjxu6Yw4mp3FU56x:-LCWhyYc2zuNCBxzhS6W:b-896fix) ([raw file](https://github.com/google/mobly/blob/25b4fab425e1e2aac0aca08b65e6bbd748a51cf8/mobly/controllers/android_device_lib/adb.py#L184)):*

this is not just "format" right? as it also adds args like `-s` to it.
Does the name . `_construct_adb_cmd` make sense?
</blockquote></details>

Done.

---

*[mobly/controllers/android_device_lib/adb.py, line 196 at r1](https://reviewable.io/reviews/google/mobly/447#-LCW_YL1CK-qMOYOIpUD:-LCWhz1T6oR1JzPAwART:b-896fix) ([raw file](https://github.com/google/mobly/blob/25b4fab425e1e2aac0aca08b65e6bbd748a51cf8/mobly/controllers/android_device_lib/adb.py#L196)):*

can we be more specific here?
seems like we always return a str when `shell` is true, and a list of strings otherwise?
</blockquote></details>

Done.

---

*[mobly/controllers/android_device_lib/adb.py, line 201 at r1](https://reviewable.io/reviews/google/mobly/447#-LCW_B3VD_5H14gG7Fvg:-LCWiBJy8kfp310yUSO-:b-fxzoqz) ([raw file](https://github.com/google/mobly/blob/25b4fab425e1e2aac0aca08b65e6bbd748a51cf8/mobly/controllers/android_device_lib/adb.py#L201)):*

could we use `cli_cmd_to_string` and reorder to simplify the logic of this section?
</blockquote></details>

I don't see how this can be reordered to be simplified?

---

*[tests/mobly/controllers/android_device_lib/adb_test.py, line 126 at r1](https://reviewable.io/reviews/google/mobly/447#-LCW_wwh2o93a_eAQeL7:-LCWiLHM9SJcYWrDJU5m:b-896fix) ([raw file](https://github.com/google/mobly/blob/25b4fab425e1e2aac0aca08b65e6bbd748a51cf8/tests/mobly/controllers/android_device_lib/adb_test.py#L126)):*

let's add some cases where the input is a whole string and some of the args in that string require auto quoting?

e.g. 'shell ls /asdafsfd/asdf-asfd/asa'
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/google/mobly/447)*

xpconanfan:
:lgtm:

---

---

*[mobly/controllers/android_device_lib/adb.py, line 201 at r1](https://reviewable.io/reviews/google/mobly/447#-LCW_B3VD_5H14gG7Fvg:-LC_FGna0nO-yjA5zpDe:b2pnemj) ([raw file](https://github.com/google/mobly/blob/25b4fab425e1e2aac0aca08b65e6bbd748a51cf8/mobly/controllers/android_device_lib/adb.py#L201)):*

I don't see how this can be reordered to be simplified?
</blockquote></details>

so all the quoting and adding '-s' part is the same right? it's just the final output format is either a list or a `join` of the list items.
I was thinking maybe we don't need these two separate branches inside.

But it's not a huge deal. Up to you.

---


*Comments from [Reviewable](https://reviewable.io/reviews/google/mobly/447#-:-LC_FzdtCe6NXxrqrm_e:bnfp4nl)*

winterfroststrom:

---

*[mobly/controllers/android_device_lib/adb.py, line 201 at r1](https://reviewable.io/reviews/google/mobly/447#-LCW_B3VD_5H14gG7Fvg:-LC_Lkby-MtRFb_sqdub:bpar1m0) ([raw file](https://github.com/google/mobly/blob/25b4fab425e1e2aac0aca08b65e6bbd748a51cf8/mobly/controllers/android_device_lib/adb.py#L201)):*

so all the quoting and adding '-s' part is the same right? it's just the final output format is either a list or a `join` of the list items.
I was thinking maybe we don't need these two separate branches inside.

But it's not a huge deal. Up to you.
</blockquote></details>

Not really?
The shell=False branch doesn't do any quoting, presumably subprocess handles this already.

I could probably flatten the branches, but the resulting logic is probably less readable and more error-prone.
This is because there's not really a common data representation that works elegantly.
For example, if the common representation is string, shell=False needs to parse stuff out of it.
If the common representation is a list, you don't want to quote name or self.serial when shell=True
and you don't want to quote adb or serial when shell=False;
but then you need logic to determine how to handle that (otherwise, it changes behaviour, which may break stuff).
Then, it needs to be something more complicated, which is also not really ideal (e.g. tagged tuples, specific class).

So I think the logic could be like:
1. serial logic
2. argument logic
3. concat logic
But each phase would actually have to check both serial and shell, so instead of two separate branches, you have 3 sets of branches/metadata.

Maybe I'm missing something simpler, but that's my evaluation of it.

---


*Comments from [Reviewable](https://reviewable.io/reviews/google/mobly/447)*

PR #473, opened 2018-07-19
Log each rpc msg in snippet client for better debugging.

### LINKED ISSUE #436: Log snippet client calls at debug level
Right now there is no log for calls made to snippet client side, which makes it difficult to debug should a snippet client call hangs.

On debug level, we should log messages sent and received by the client, same way we log messages received and sent by snippet server on devices.

xpconanfan:
Fixes #436

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/473)


xpconanfan:
It was a mistake. Encoding arg has been removed.

PR #487, opened 2018-08-08
Fix max timeout for `CallbackHandler.waitForEvent`.

xpconanfan:
This method should not be bound by the max timeout of a single Rpc call.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/487)

### DIFF
--- a/mobly/controllers/android_device_lib/callback_handler.py
+++ b/mobly/controllers/android_device_lib/callback_handler.py
@@ -1,11 +1,11 @@
 # Copyright 2017 Google Inc.
-# 
+#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


mobly/controllers/android_device_lib/callback_handler.py
```diff
@@ -134,6 +136,8 @@ def waitForEvent(self, event_name, predicate, timeout=DEFAULT_TIMEOUT):
             rpc_timeout = deadline - time.time()
             if rpc_timeout < 0:
                 break
+            # A single RPC call cannot exceed MAX_TIMEOUT.
+            rpc_timeout = min(rpc_timeout, MAX_TIMEOUT)
```
winterfroststrom: this seems like user error though?

xpconanfan: Why is it a user error?

PR #489, opened 2018-08-16
Clean up dep requirements in `setup.py`.

xpconanfan:
* Move test-only deps out of package dep list.
* Remove unnecessary deps.

`setuptools` version in ubuntu dist is no longer a concern.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/489)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,8 +62,8 @@ def main():
         include_package_data=False,
         scripts=['tools/sl4a_shell.py', 'tools/snippet_shell.py'],
         tests_require=[
-            'pytest',
             'mock',
+            'pytest',
             'pytz',
         ],
         install_requires=install_requires,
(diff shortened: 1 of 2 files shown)


winterfroststrom:
apparently timeout decorator is used in monsoon.py

xpconanfan:
Whaa, how come localing running unit tests didn't catch this..

winterfroststrom:
because our monsoon unit test have like no coverage

winterfroststrom:
also, you probably already have the dependency installed

xpconanfan:
the test sets up deps from scratch every time.
I think it's just we don't have coverage for monsoon...

winterfroststrom:
no, I mean that the reason why running it locally passed for you was probably because a previous run of it installed the deps, where restinalling isn't actually necessary, otherwise, if it was never installed, and the reinstall logic didn't install it, I have a hard time imagining it passing

xpconanfan:
Oh wait, it's the other way around...

PR #493, opened 2018-08-27
Move `register_controller` into `base_test`.

### LINKED ISSUE #492: Move `register_controller` into `base_test`
For historic reasons, `register_controller` lives under `test_runner`.
Back then the consideration was, one day we may want to let controller objects persist across multiple test classes.

This causes multiple organizational problems and incurs complexity.

In hindsight that is a very bad idea and controller object lifecycles should be contained within each class. So it is time to push controller management into `base_test`.

xpconanfan:
Now the controller objects of a class is properly encapsulated within the test class itself.

This has many benefits:
* makes the structure cleaner and easier to understand
* no more awkward passing of a partial function from test runner to each test class.
* docstring of the controller interface is properly located now.

Fixes #492

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/493)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -335,6 +335,8 @@ class BaseTestClass(object):
 
         This will be called after each test class.
         """
+        # TODO(xpconanfan): actually record these errors instead of just
+        # logging them.
         for name, destroy in self._controller_destructors.items():
             try:
                 logging.debug('Destroying %s.', name)
@@ -900,6 +902,7 @@ class BaseTestClass(object):
             raise e
         finally:
             self._teardown_class()
+            self._unregister_controllers()
             logging.info('Summary for test class %s: %s', self.TAG,
                          self.results.summary_str())
 
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -46,6 +46,30 @@ class Error(Exception):
     """Raised for exceptions that occured in BaseTestClass."""
 
 
+def _verify_controller_module(module):
+    """Verifies a module object follows the required interface for
+    controllers.
+
+    Args:
+        module: An object that is a controller module. This is usually
+            imported with import statements or loaded by importlib.
+
+    Raises:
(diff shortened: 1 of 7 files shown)


mobly/base_test.py
```diff
@@ -46,6 +46,30 @@ class Error(Exception):
...
+        ControllerError: if the module does not match the Mobly controller
+            interface, or one of the required members is null.
+    """
+    required_attributes = ('create', 'destroy', 'MOBLY_CONTROLLER_CONFIG_NAME')
+    for attr in required_attributes:
+        if not hasattr(module, attr):
+            raise signals.ControllerError(
+                'Module %s missing required controller module attribute'
+                ' %s.' % (module.__name__, attr))
+        if not getattr(module, attr):
```
jeffherman: Should we check if `callable(getattr(module, 'create'))`, `callable(getattr(module, 'destroy'))`, and `isinstance(getattr(module, 'MOBLY_CONTROLLER_CONFIG_NAME'), str)`?

xpconanfan:
CI failure it not reproducible locally in py2.7, 3.4, 3.5, 3.6
I don't know what Travis is doing...

dthkao:
https://stackoverflow.com/questions/24802740/unnamed-python-objects-have-the-same-id

Looks like the test was based on a bad assumption to begin wth. The two objects you're id()-ing have non-overlapping scope so no guarantee of unique mem address.

Try making them attributes of the outer test (although this relies on the mock controller destroy not actually doing anything)?

winterfroststrom:
I re-ran the build, so it's probably a test flake. Probably per the reason dthkao@ mentioned.
Is there an easy way to rewrite the test to avoid this?

mobly/base_test.py
xpconanfan: A bit of an overkill?
If it's not callable, it'll fail anyway. The python "if it quacks like a duck" principle :P

PR #495, opened 2018-08-30
Properly record controller info.

### LINKED ISSUE #484: Custom controller info may not be recorded in summary yaml
Controller modules may choose to allow customization of controller info during the test. E.g. `AndroidDevice.add_device_info` adds extra info of the controller objects to the summary yaml.

But `TestRunner` only record the controller info at the controller registration time, so the customized info may never get recorded depending on the implementation of the controller module.

I.e. If the info is a static string, we it will be set at registration time, and subsequent changes to the info field won't be picked up.

### LINKED ISSUE #485: Inconsistency in controller handling for multi-class run
In a test run with multiple classes (like a suite), the controller objects are created and destroyed for every class. But the controller info is only dumped once at the very end of the test as a test run level block.

Recording class-level info as a run-level info causes information loss and inconsistency in reporting.

E.g. we allow users to add custom info to `AndroidDevice` obj in test. Right now if two classes are executed and they add different custom info to their `AndroidDevice` objects, some info is lost.

The root cause comes from this comment
https://github.com/google/mobly/blob/811e52a95fd8c35253dbd918cd7fcbda8a78369b/mobly/records.py#L490

The controller info is **not** the same across all classes in the same run, even if the same controller units are used in reality.

E.g. the same Android devices are used in a test run. But they can have different controller info at the end of each class due to customized controller info.

xpconanfan:
* Support multiple groups of controller info in  test result record.
* Record controller info at the end instead of the beginning of a
  test class run.
* Introduce a proper data class for controller info.

This changes public API `mobly.records.TestResult.controller_info`.

Fixes #484 
Fixes #485

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/495)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -331,7 +331,7 @@ class BaseTestClass(object):
                 module.destroy(self._controller_registry[name])
             except:
                 logging.exception('Exception occurred destroying %s.', name)
-        self._controller_registry = OrderedDict()
+        self._controller_registry = collections.OrderedDict()
         self._controller_modules = {}
 
     def _record_controller_info(self):
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -134,8 +134,8 @@ class BaseTestClass(object):
         self.current_test_name = None
         self._generated_test_table = collections.OrderedDict()
         # Controller object management.
-        self._controller_registry = {}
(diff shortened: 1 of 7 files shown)


winterfroststrom:
build failure?

Issue #486, opened 2018-08-01
Refactoring Android logcat to allow for publishing and subscribing


jeffherman:
The [mobly.controller.android_device](https://github.com/google/mobly/blob/master/mobly/controllers/android_device.py) logcat functionality currently creates a file in `/tmp/logs` ([populated by the Mobly test runner](https://github.com/google/mobly/blob/master/mobly/logger.py#L135)).

While this allows for test artifact preservation, it doesn't allow the test to easily make use of the live logcat stream. A test could attach a Unix `tail -f` subprocess but due to an [issue](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/882147) with `inotify` on OverlayFS on Linux, it is often the case that any particular logcat line may be delayed and, therefore, cannot be used for realtime processing.

Another approach would be to create an entirely new logcat publisher class (a subclass of `threading.Thread`) and subscriber with `subscribe()` and `unsubscribe()` methods along with a logcat subscriber class (an abstract base class) that would handle new log lines. An example of a subscriber class would something that simply saves all raw logcat lines to a file (such as the current system) or something that's monitoring a particular regular expression and updating some state variables.

xpconanfan:
I don't think we will change the current logcat mechanism to a different way.
We will likely preserve the current logcat and provide new APIs to make it easier to consume the content.

PR #514, opened 2018-10-03
Logcat publisher/subscriber service

jeffherman:
Implementation of https://github.com/google/mobly/issues/486. This feature allows users of Mobly's AndroidDevice controller to define their own logcat subscriber to asynchronously track log output as they get produced and update a test state. This saves test writers from manually parsing the saved logcat (typically in `/tmp/logs/AndroidDevice`) for a specific log line.

Right now, there's only one concrete class implementation of `LogcatSubscriber` which is created ephemerally when the user calls the `logcat_event` context manager:

      >>> from mobly.controllers import android_device
      >>> ad = android_device.AndroidDevice('0123456789')
      >>> with ad.logcat_event(pattern='writeEvent (.*) (.*)') as event:
             event.wait()
             print('time: %s' % event.trigger.time)
             print('pid: %d' % event.trigger.pid)
             print('tid: %d' % event.trigger.tid)
             print('tag: %s' % event.trigger.tag)
             print('message: %s' % event.trigger.message)
             print('regex match groups: %s' % str(event.match.groups()))
      time: 2018-10-02 16:19:00.408000
      pid: 1463
      tid: 1463
      tag: vol.Events
      message: writeEvent dismiss_dialog volume_controller
      regex match groups: ('dismiss_dialog', 'volume_controller')

However, if the user wants to create their own `LogcatSubscriber`, it would simply be a matter of subclassing `LogcatSubscriber` and registering that subscriber with the publisher. For example, suppose you had a service that was dumping out logcat lines such as:

    01-02 03:45:02.300  2000  2001 I MyService: metric_a=20 metric_b=80

We can define a subclass `MyLogcatSubscriber` and register that with the `AndroidDevice` instance: 

    from mobly.controllers import android_device
    from mobly.controllers.android_device_lib.services import logcat_pubsub
    
    class MyLogcatSubscriber(logcat_pubsub.LogcatSubscriber):
        def __init__(self):
            self.metric_a = None
            self.metric_b = None
        
        def handle(self, data):
            match = re.match('metric_a=(\d+) metric_b=(\d+)', data.message)
            if match:
                self.metric_a = int(match.group(1))
                self.metric_b = int(match.group(2))
    
    ad = android_device.AndroidDevice('0123456789')
    my_logcat_subscriber = MyLogcatSubscriber()
    ad.register_logcat_subscriber(my_logcat_subscriber)

When the log line above appears in the logcat, the subscriber's internal state will get updated:

    >>> print(my_logcat_subscriber.metric_a)
    20
    >>> print(my_logcat_subscriber.metric_b)
    80

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/514)


winterfroststrom:
build failed? make sure this works with python3?

jeffherman:
___
*[mobly/controllers/android_device.py, line 1026 at r4](https://reviewable.io/reviews/google/mobly/514#-LNvhWy-1Bp5JjFi5AKt:-LNvotv07quclX3rOuuQ:b6ufpot) ([raw file](https://github.com/google/mobly/blob/8d4968dc61859b2a5adcbc206638e60c5e5bfeac/mobly/controllers/android_device.py#L1026)):*

> ```
>       """Register a logcat subscriber with the logcat publisher.
>  
>       Check if an existing logcat publisher is active. If not then lazy
>       instantiate a new one and register the subscriber to that publisher.
>  
>       Args:
>           subscriber: LogcatSubscriber, subscriber to register
>       """
>       try:
>           logcat_publisher = self.services.logcat_publisher
>       except KeyError:
>           self.services.register('logcat_publisher',
>                                  logcat_pubsub.LogcatPublisher)
>           logcat_publisher = self.services.logcat_publisher
>       subscriber.subscribe(self.services.logcat_publisher)
>  
>     def logcat_event(self, pattern='.*', tag='*', level='V'):
>         """Context manager object for a logcat event.
>  
>         Args:
>             pattern: str, Regular expression pattern to trigger on.
>             tag: str, Tag portion of filterspec string.
>             level: str, Level portion of filterspec string.
>         """
>         subscriber = logcat_pubsub.LogcatEventSubscriber(
>             pattern=pattern, tag=tag, level=level)
>         self.register_logcat_subscriber(subscriber)
> ```

I don't see why this is better than going through the service manager?
Like if someone wants to subclass/change the publisher,
this implementation will make it less convenient
because you're tightly coupling the publisher with the android device
</blockquote></details>

Not sure there's any value of subclassing the publisher. Re: registering a subscriber to the single publisher, I thought about directly calling the `ad.services.logcat_publisher` method directly but that's not guaranteed to exist. The only service to exist by default is `snippets` which is registered in the `__init__()` method and even the `logcat` service is registered after a manual call to `start_services()`.

This implementation allows us to lazy-register the service since most people won't need to use this and we don't want to burn up CPU cycles running a logcat publisher for those that don't.

jeffherman:
___
*[mobly/controllers/android_device.py, line 1026 at r4](https://reviewable.io/reviews/google/mobly/514#-LNvhWy-1Bp5JjFi5AKt:-LO-brVI18j8-48COQii:b-dzsp9z) ([raw file](https://github.com/google/mobly/blob/8d4968dc61859b2a5adcbc206638e60c5e5bfeac/mobly/controllers/android_device.py#L1026)):*

+1 to making this a service.
We don't want to add more feature-specific method to ad.
</blockquote></details>

Okay, I changed the API a bit. There's a little bit of awkwardness in creating a logcat event context manager (a publisher method is temporarily creating its own subscriber) but it's not too bad.

jeffherman:
___
*[mobly/controllers/android_device_lib/services/logcat_pubsub.py, line 67 at r6](https://reviewable.io/reviews/google/mobly/514#-LO-d-Nc0xBiaRW3GM-Q:-LO5mTG46oELKlSWBPnJ:b-kj1fif) ([raw file](https://github.com/google/mobly/blob/732fe46db269b8195e0087f140d6f66772608961/mobly/controllers/android_device_lib/services/logcat_pubsub.py#L67)):*
<details><summary><i>Previously, jeffherman wrote…</i></summary><blockquote>

Right now, mobly's built-in logcat service dumps to `/tmp` which has issues with `inotify` as described [here](https://github.com/google/mobly/issues/486). This makes asynchronously tracking a growing file unreliable (we've seen it fail to pick up changes to the logcat stream).
</blockquote></details>

Is there a know issue with running multiple logcats simultaneously? Haven't ran into this myself.

jeffherman:
___
*[tests/mobly/controllers/android_device_lib/services/logcat_pubsub_test.py, line 52 at r6](https://reviewable.io/reviews/google/mobly/514#-LO-dV4e7P-YKvBKiq5m:-LO5nod-BqF5KGxpMinB:b-zho6lo) ([raw file](https://github.com/google/mobly/blob/732fe46db269b8195e0087f140d6f66772608961/tests/mobly/controllers/android_device_lib/services/logcat_pubsub_test.py#L52)):*

can you add tests cases where non-ascii strings are being subscribed for?
Like a unicode, special chars.

</blockquote></details>

Looks like mocking an I/O stream using `os.pipe()` with Unicode is causing  test issues that only manifest in Windows.

    In [1]: import os
       ...: fd_s, fd_m = os.pipe()
       ...: stdout = os.fdopen(fd_s, 'r')
       ...: mock_stdout = os.fdopen(fd_m, 'w')
       ...: mock_stdout.write('\xcf\x80\n')
       ...: mock_stdout.flush()
       ...: stdout.readline()
       ...:
    Out[1]: '\xcf\x80\n'

The same code in windows results in

        def encode(self, input, final=False):
    >       return codecs.charmap_encode(input,self.errors,encoding_table)[0]
    E       UnicodeEncodeError: 'charmap' codec can't encode character '\x80' in position 58: character maps to <undefined>

I think we should just MagicMock a new stdout that handles special characters.

jeffherman:
___
*[mobly/controllers/android_device_lib/services/logcat_pubsub.py, line 67 at r6](https://reviewable.io/reviews/google/mobly/514#-LO-d-Nc0xBiaRW3GM-Q:-LOEz8tqEvSJz8b_MoIV:b-xyd77v) ([raw file](https://github.com/google/mobly/blob/732fe46db269b8195e0087f140d6f66772608961/mobly/controllers/android_device_lib/services/logcat_pubsub.py#L67)):*

Yes, many OEM devices cannot support multiple logcat streams.
It is one of the challenges of making generic utils for Android.
We have to consider many cases.

I don't think mobly's logcat service dumps to /tmp?
</blockquote></details>

Actually, it does:

    def __init__(self, serial=''):
        self._serial = str(serial)
        # logging.log_path only exists when this is used in an Mobly test run.
        self._log_path_base = getattr(logging, 'log_path', '/tmp/logs')
        self._log_path = os.path.join(
            self._log_path_base, 'AndroidDevice%s' % self._normalized_serial)

If you trace `_log_path`, it eventually makes its way to the current logcat module. That was how I found out about the `inotify` bug in OverlayFS.

The other idea I had was to create another subscriber class that simply saved out the data to a file but now we would have multiple things accomplishing the exact same function.

jeffherman:
___
*[mobly/controllers/android_device_lib/services/logcat_pubsub.py, line 67 at r6](https://reviewable.io/reviews/google/mobly/514#-LO-d-Nc0xBiaRW3GM-Q:-LOtixkEC3Wu9bAZ5iLk:b-21e1ep) ([raw file](https://github.com/google/mobly/blob/732fe46db269b8195e0087f140d6f66772608961/mobly/controllers/android_device_lib/services/logcat_pubsub.py#L67)):*

Is it possible to design some simple APIs to add to the existing `logcat` service?
Do we really need another service for this considering is so closely tied with logcat?
</blockquote></details>

Probably not without rearchitecting the existing logcat to not directly dump out to a file but rather to let Python handle the writing of individual lines to the destination file to prevent it from writing to `/tmp` before it has had a chance to parse its output (so avoiding using the `>>` redirection). Is that something we're willing to do?

jeffherman:
___
*[mobly/controllers/android_device_lib/services/logcat_pubsub.py, line 67 at r6](https://reviewable.io/reviews/google/mobly/514#-LO-d-Nc0xBiaRW3GM-Q:-LOzCSu61FG13B7cgwdZ:b-l54r6n) ([raw file](https://github.com/google/mobly/blob/732fe46db269b8195e0087f140d6f66772608961/mobly/controllers/android_device_lib/services/logcat_pubsub.py#L67)):*

Sounds like this is something that should have been resolved during design doc review as opposed to code review.
I don't believe we ever signed off on the design?
</blockquote></details>

We had some OpenHTF usage way back when that had this issue. Maybe we can add a doc to suggest calling `mobly.logger.setup_test_runner` with a non-`/tmp` directory.

Issue #533, opened 2018-10-12
Errors in `teardown_class` are not properly recorded


xpconanfan:
If Mobly encounters an error in the cleanup stage of `teardown_class`, the error would not be recorded anywhere except the cli output, which makes debugging difficult.

This points to a bigger problem: the errors occurred in base test and runner themselves are not clearly handled.

xpconanfan:
```
  File ".../third_party/py/mobly/test_runner.py", line 348, in run
    tests=test_run_info.tests)
  File ".../third_party/py/mobly/test_runner.py", line 315, in _run_test_class
    cls_result = test_instance.run(tests)
  File ".../third_party/py/mobly/base_test.py", line 816, in run
    self._teardown_class()
  File ".../third_party/py/mobly/base_test.py", line 332, in _teardown_class
    self._record_controller_info()
  File ".../third_party/py/mobly/base_test.py", line 252, in _record_controller_info
    for record in self._controller_manager.get_controller_info_records():
  File ".../third_party/py/mobly/controller_manager.py", line 208, in get_controller_info_records
    controller_module_name)
  File ".../third_party/py/mobly/controller_manager.py", line 179, in _create_controller_info_record
    copy.copy(self._controller_objects[controller_module_name]))
  File ".../third_party/py/mobly/controllers/android_device.py", line 124, in get_info
    return [ad.device_info for ad in ads]
  File ".../third_party/py/mobly/controllers/android_device.py", line 478, in device_info
    'model': self.model,
  File ".../third_party/py/mobly/controllers/android_device.py", line 789, in model
    return self.adb.getprop('ro.product.name').lower()
  File ".../third_party/py/mobly/controllers/android_device_lib/adb.py", line 294, in getprop
    return self.shell('getprop %s' % prop_name).decode('utf-8').strip()
  File ".../third_party/py/mobly/controllers/android_device_lib/adb.py", line 394, in adb_call
    name, args, shell=shell, timeout=timeout, stderr=stderr)
  File ".../third_party/py/mobly/controllers/android_device_lib/adb.py", line 273, in _exec_adb_cmd
    adb_cmd, shell=shell, timeout=timeout, stderr=stderr)
  File ".../third_party/py/mobly/controllers/android_device_lib/adb.py", line 182, in _exec_cmd
    raise AdbError(cmd=args, stdout=out, stderr=err, ret_code=ret)
[...]

PR #540, opened 2018-10-22
Fix unit test flake.

xpconanfan:
Followup for #533

### DIFF
--- a/mobly/controller_manager.py
+++ b/mobly/controller_manager.py
@@ -50,7 +50,7 @@ def verify_controller_module(module):
 
 
 class ControllerManager(object):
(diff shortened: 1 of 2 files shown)


winterfroststrom:
Actually, an assumption here is that controller manager is never called outside of basetest, how reasonable is that?

xpconanfan:
Good point.
It is definitely a reasonable behavior.
The question is, do we want to use `controller_manager` in other context?

Right now there is no use case afaik, but is there any future concerns?

xpconanfan:
Clarified in docstring that this shall be used within a Mobly test context.
We can change it later if needed

PR #542, opened 2018-11-01
Feature 514 refactor

jeffherman:
Refactor of https://github.com/google/mobly/pull/514 for #486. This implementation puts the publisher directly in the original `Logcat` class and adds the `subscribe`, `unsubscribe`, and `event` methods. The publisher will automagically start whenever there is anyone subscribed.


jeffherman:
___
*[setup.py, line 21 at r11](https://reviewable.io/reviews/google/mobly/542#-LQGbtCq8_KKg3iaQ4Da:-LQGfTIP57ikmtWBGL7S:b-1x9e6b) ([raw file](https://github.com/google/mobly/blob/9921422f26a70a70348bdd4a9e694170cd51b7d4/setup.py#L21)):*

is there no way to avoid using dateutil?
</blockquote></details>

I could write my own parser.

jeffherman:
___
*[mobly/controllers/android_device_lib/services/logcat.py, line 360 at r11](https://reviewable.io/reviews/google/mobly/542#-LQGcMoc1MBtXKuKEA9J:-LQGfnQ-71pny0j0w7lX:baxkw) ([raw file](https://github.com/google/mobly/blob/9921422f26a70a70348bdd4a9e694170cd51b7d4/mobly/controllers/android_device_lib/services/logcat.py#L360)):*

if you assume there's a set datetime or log line signature, then we are in trouble.
there is actually no guarantee for the format of each log line in logcat.

e.g. kernel log could be mixed with system log which has different formats. and different OEM devices may output logs in different formats.
And users can alter the logline format by passing in extra args to logcat cmd.

So I'd suggest not assuming what the logline format is.


Shall we simplify the design of this feature and simply make it a pattern matcher?
</blockquote></details>

Is that actually true? From the existing OEM devices I've seen, the log lines appear to follow a standard format up until the tag, at which point it's free form (that seems to fit what the [doc describes](https://developer.android.com/studio/command-line/logcat)).

jeffherman:
___
*[docs/logcat_pubsub_tutorial.md, line 72 at r10](https://reviewable.io/reviews/google/mobly/542#-LQGaltj5C_Wc0ffFKC4:-LQKl3Of9XO9ir3Jl7S-:b-73os6m) ([raw file](https://github.com/google/mobly/blob/5801255091e7825c4b4fc431c00cafea5a627ba4/docs/logcat_pubsub_tutorial.md#L72)):*

What "data" would be fed to other places? is it all the data that has been passed in so far?
Users can still use a buffer as part of the condition func to save whatever they want.

I don't see a compelling reason for the complexity involved in these APIs.
</blockquote></details>

Well, if that's the case (and we're trying to catch all prior and future use cases) then perhaps we should drop this PR since the main purpose was to bring some of our own generic code into the open source package. But if we're removing most of what we've implemented internally then there's not really much left (we would still have to keep the bulk of this code in our own code).

PR #546, opened 2018-11-28
Fix help() rpc to display new lines properly be default in snippet sh…

winterfroststrom:
…ell.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/546)

### DIFF
--- a/mobly/controllers/android_device_lib/snippet_client.py
+++ b/mobly/controllers/android_device_lib/snippet_client.py
@@ -339,7 +339,11 @@ class SnippetClient(jsonrpc_client_base.JsonRpcClientBase):
         return ''
 
     def help(self, print_output=True):
-        """Calls the help RPC, which returns list of RPC calls available.
+        """Calls the help RPC, which returns the list of RPC calls available.
+
+        This RPC should normally be used in an interactive console environment
+        where the output should be printed instead of returned. Otherwise,
+        newlines will be escaped, which will make the output difficult to read.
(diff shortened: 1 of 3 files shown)


tools/snippet_shell.py
```diff
@@ -34,13 +34,30 @@
 from mobly.controllers.android_device_lib import jsonrpc_shell_base
 
 
+def _patch_print_rpc(snippet, name):
```
xpconanfan: i was thinking simply defining the `help` method in the snippet client class...
would be less hacky?

winterfroststrom: Doesn't that potentially cause issues if someone actually wants the help() response?
I could add a default argument to disable printing in that case I guess.

winterfroststrom: Changed.

This is a bit more dangerous to do, but it's probably okay.

mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 """JSON RPC interface to Mobly Snippet Lib."""
 
+from __future__ import print_function
```
xpconanfan: `print_function` is not used?
is this a magical line?

winterfroststrom: it's for python2 compatibility?
print vs print()

PR #564, opened 2019-03-01
Remove `tests` module from installation.

mderu:
This change also makes the script paths absolute instead of relative,
which allows Mobly to be installed without being in the base directory.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/564)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ def main():
         license='Apache2.0',
         url='https://github.com/google/mobly',
         download_url='https://github.com/google/mobly/tarball/1.8.1',
-        packages=setuptools.find_packages(),
+        packages=setuptools.find_packages(exclude=['tests']),
         include_package_data=False,
         scripts=['tools/sl4a_shell.py', 'tools/snippet_shell.py'],
         tests_require=[


setup.py
```diff
@@ -59,9 +61,11 @@ def main():
         license='Apache2.0',
         url='https://github.com/google/mobly',
         download_url='https://github.com/google/mobly/tarball/1.8.1',
-        packages=setuptools.find_packages(),
+        package_dir={'mobly':'mobly'},
+        packages=setuptools.find_packages(exclude=['tests']),
```
xpconanfan: I think these two lines are sufficient.
I don't think we need to do the absolute path thing?

xpconanfan: Actually I'm still seeing the "tests" files included in the distribution with this change when using cmd:
```
$ python  setup.py sdist --formats=zip
```

mderu: When I generate this, I'm doing it within a virtualenv. Are you sure it isn't already a part of your PYTHONPATH/installed?

Here are the full steps I am using:

```
$ mobly_dir="~/mobly"
$ python3 -m virtualenv /tmp/venv >&2 >/dev/null
$ source /tmp/venv/bin/activate
$ cp -r $mobly_dir /tmp/venv/
$ cd /tmp/venv/mobly/
$ python3 setup.py sdist --formats=zip
...
$ python3 -m easy_install dist/mobly-1.8.1.zip
...
$ cd / # Somewhere not next to the tests/ folder in the mobly base directory
$ python3.5
Type "help", "copyright", "credits" or "license" for more information.
>>> import mobly
>>> import tests
ImportError: No module named 'tests'
>>> 
```
I confirmed this works with only making a single line change. I'll update the PR now.

xpconanfan: I see.
Seems like the tests will be added to sdist regardless but when someone installs from that zip, it won't install the tests.

mderu: Change is up. Confirmed installation from the zip does not install tests.

PR #565, opened 2019-03-01
Fix assert message to have a single stacktrace

mderu:
In Python3, the stacktrace is duplicated when an exception is caught and raised as a different Exception object. We can get around this by determining the assertion failed, and re-raising the assertion as a TestFailure.

Also, AssertionError is caught instead of Exception. It is the base class name for the errors pyunit raises in both python2 and python3.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/565)

### DIFF
--- a/mobly/asserts.py
+++ b/mobly/asserts.py
@@ -51,6 +51,8 @@ def assert_equal(first, second, msg=None, extras=None):
         if msg:
             my_msg = "%s %s" % (my_msg, msg)
 
+    # This raise statement is outside of the above except statement to prevent
+    # Python3's exception message from having two tracebacks.
     if my_msg is not None:
         raise signals.TestFailure(my_msg, extras=extras)
 
--- a/mobly/asserts.py
+++ b/mobly/asserts.py
@@ -43,18 +43,16 @@ def assert_equal(first, second, msg=None, extras=None):
         extras: An optional field for extra information to be included in
             test result.
     """
+    my_msg = None
     try:
         _pyunit_proxy.assertEqual(first, second)
-    except Exception as e:
-        # We have to catch all here for py2/py3 compatibility.
-        # In py2, assertEqual throws exceptions.AssertionError, which does not
-        # exist in py3. In py3, it throws unittest.case.failureException, which
-        # does not exist in py2. To accommodate using explicit catch complicates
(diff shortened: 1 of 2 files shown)


mderu:
No, it's far less severe than that. When you run this in python2 you get the following:

```
$ python2
>>> from mobly import asserts
>>> asserts.assert_equal(1, 2, 'msg')
  File "mobly/asserts.py", line 55, in assert_equal
    raise signals.TestFailure(my_msg, extras=extras)
mobly.signals.TestFailure: Details=1 != 2 msg, Extras=None
```

When you run this in Python3, you get:

```
$ python3
>>> from mobly import asserts
>>> asserts.assert_equal(1, 2, 'msg')
  File "/usr/local/google/home/markdr/mobly/mobly/asserts.py", line 47, in assert_equal
    _pyunit_proxy.assertEqual(first, second)
  File "/usr/lib/python3.6/unittest/case.py", line 829, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib/python3.6/unittest/case.py", line 822, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 1 != 2

During handling of the above exception, another exception occurred:

  File "/usr/local/google/home/markdr/mobly/mobly/asserts.py", line 57, in assert_equal
    fail(my_msg, extras=extras)
  File "/usr/local/google/home/markdr/mobly/mobly/asserts.py", line 239, in fail
    raise signals.TestFailure(msg, extras)
mobly.signals.TestFailure: Details=1 != 2 msg, Extras=None
```

With this change, the stacktrace is simplified to more closely represent what it looks like in pyunit:

```
markdr@markdr:~/mobly$ python3
>>> from mobly import asserts
>>> asserts.assert_equal(1, 2, 'msg')
  File "/usr/local/google/home/markdr/mobly/mobly/asserts.py", line 55, in assert_equal
    raise signals.TestFailure(my_msg, extras=extras)
mobly.signals.TestFailure: Details=1 != 2 msg, Extras=None
```

Because the framework intends to make the assert functions appear to be native, it makes more sense to hide the fact that we are raising an AssertionError with pyunit, and re-raising it as a test failure. End-users shouldn't be exposed to this quirk, which ultimately makes the stacktrace in Python3 more confusing.

PR #570, opened 2019-03-06
Moving the util and updating references in tests

stroz:
The motivation is so this function can be better reused by other places.

This change is [Reviewable](https://reviewable.io/reviews/google/mobly/570)

### DIFF
--- a/mobly/controllers/android_device_lib/adb.py
+++ b/mobly/controllers/android_device_lib/adb.py
@@ -16,7 +16,6 @@ from builtins import str
 from past.builtins import basestring
 
 import logging
-import pipes
 import psutil
 import subprocess
 import threading
--- a/mobly/controllers/android_device_lib/adb.py
+++ b/mobly/controllers/android_device_lib/adb.py
@@ -63,7 +63,7 @@ class AdbError(Error):
 
     def __str__(self):
         return ('Error executing adb cmd "%s". ret: %d, stdout: %s, stderr: %s'
-                ) % (cli_cmd_to_string(self.cmd), self.ret_code, self.stdout,
+                ) % (utils.cli_cmd_to_string(self.cmd), self.ret_code, self.stdout,
                      self.stderr)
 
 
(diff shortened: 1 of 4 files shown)


stroz:
___
*[mobly/controllers/android_device_lib/adb.py, line 19 at r1](https://reviewable.io/reviews/google/mobly/570#-L_UWE25A8zO9lAofmFs:-L_hs7cvDxc7V0yDihz_:b-ktdozw) ([raw file](https://github.com/google/mobly/blob/538e7423bc78c812931fd37c990f08bee85d0a58/mobly/controllers/android_device_lib/adb.py#L19)):*

> ```
> import pipes
> ```
no longer used?
</blockquote></details>

Looks like it's used here in cli_cmd_to_string:
  return ' '.join([pipes.quote(arg) for arg in args])

stroz:
___
*[mobly/controllers/android_device_lib/adb.py, line 19 at r1](https://reviewable.io/reviews/google/mobly/570#-L_UWE25A8zO9lAofmFs:-L_iaxnX2nU45lg2j2vt:b-ubda8l) ([raw file](https://github.com/google/mobly/blob/538e7423bc78c812931fd37c990f08bee85d0a58/mobly/controllers/android_device_lib/adb.py#L19)):*

But you moved that function to `utils` right?
so we don't need this import in `adb` any more?
</blockquote></details>

Right, I believe we can remove this from adb.

stroz:
___
*[mobly/controllers/android_device_lib/adb.py, line 19 at r1](https://reviewable.io/reviews/google/mobly/570#-L_UWE25A8zO9lAofmFs:-L_ibkc446XaBAeBLjar:b-896fix) ([raw file](https://github.com/google/mobly/blob/538e7423bc78c812931fd37c990f08bee85d0a58/mobly/controllers/android_device_lib/adb.py#L19)):*
<details><summary><i>Previously, stroz (Tim Strother) wrote…</i></summary><blockquote>

Right, I believe we can remove this from adb.
</blockquote></details>

Done.

PR #642, opened 2019-10-10
Set TestRunConfig.test_bed_name as a property that tracks testbed_name.

xianyuanjia:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/642)

### DIFF
--- a/mobly/config_parser.py
+++ b/mobly/config_parser.py
@@ -185,6 +185,16 @@ class TestRunConfig(object):
         self.summary_writer = None
         self.test_class_name_suffix = None
 
+    @property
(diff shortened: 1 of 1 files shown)


mobly/config_parser.py
```diff
@@ -185,6 +185,16 @@ def __init__(self):
         self.summary_writer = None
         self.test_class_name_suffix = None
 
+    @property
+    def test_bed_name(self):
+        """Deprecated, use testbed_name directly."""
```
xpconanfan: why do we need to add a "Deprecated" property?

PR #647, opened 2019-10-15
Always operate on the current user in `snippet_client`.

xpconanfan:
It does not make sense to attempt to launch snippet in a non-active user. So we should always do package check and snippet launch in the context of the current user.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/647)

### DIFF
--- a/mobly/controllers/android_device_lib/snippet_client.py
+++ b/mobly/controllers/android_device_lib/snippet_client.py
@@ -39,8 +39,9 @@ _PROTOCOL_MAJOR_VERSION = 1
 # existing clients.
 _PROTOCOL_MINOR_VERSION = 0
 
-_LAUNCH_CMD = ('%s am instrument -w -e action start %s/' +
-               _INSTRUMENTATION_RUNNER_PACKAGE)
+_LAUNCH_CMD = (
+    '{shell_cmd} am instrument --user {user} -w -e action start {snippet_package}/'
+    + _INSTRUMENTATION_RUNNER_PACKAGE)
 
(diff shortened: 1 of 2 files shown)


xpconanfan:
Unfortunately cmds like `pm list package` defaults to user 0 instead of the current user... Hence we get errors like "snippet not installed" when attempting to launch on a device whose current user is not user 0..

mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -39,8 +39,9 @@
 # existing clients.
 _PROTOCOL_MINOR_VERSION = 0
 
-_LAUNCH_CMD = ('%s am instrument -w -e action start %s/' +
-               _INSTRUMENTATION_RUNNER_PACKAGE)
+_LAUNCH_CMD = (
+    '{shell_cmd} am instrument --user {user} -w -e action start {snippet_package}/'
```
vitiral: note: current user is the default for `am instrument`. However, personally I still think it's better practice to be explicit (so no recommended change)

winterfroststrom:
Can you make sure this works on a variety of devices?

I've seen issues on some devices where user id doesn't work properly

xpconanfan:
Yes we've tried on lower API versions and Auto devices.
Do you have specific device models that we should try on?

winterfroststrom:
the devices I thinking about are android go devices

xpconanfan:
Tried both running the adb cmd directly and running the snippet client within a test. Seems fine.

PR #649, opened 2019-10-18
Initialize dicts for TestRunConfig's dict values.

mderu:
Defaults the controller and user_params to empty dictionaries in TestRunConfig.
    
These defaults prevent users from needing to check if they are None
first, in case the runner left these values unset.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/649)

### DIFF
--- a/mobly/controllers/android_device_lib/snippet_client.py
+++ b/mobly/controllers/android_device_lib/snippet_client.py
@@ -39,8 +39,9 @@ _PROTOCOL_MAJOR_VERSION = 1
 # existing clients.
 _PROTOCOL_MINOR_VERSION = 0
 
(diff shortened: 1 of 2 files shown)


mderu:
___
*[mobly/config_parser.py, line 183 at r1](https://reviewable.io/reviews/google/mobly/649#-LrVyIPh7ZNxcY0afNzP:-LrVyIPh7ZNxcY0afNzQ:br6kl23) ([raw file](https://github.com/google/mobly/blob/12ead33d32af83b499a524fc9e6513590be78c30/mobly/config_parser.py#L183)):*
> ```Python
>         self.log_path = None
>         # Deprecated, use 'testbed_name'
>         self.test_bed_name = None
> ```

Bad merge resolution. Fixing.

PR #668, opened 2020-03-18
Fix adb service can NOT start in mobly when debugging with Intelij IDEA on gLinux.

eric100lin:
When 'debugging' Mobly PY test with Intelij IDEA on gLinux,
logcat service in Mobly can NOT start with the following error
  b'/bin/sh: /path/to/adb -s 9C041FQC200408 logcat -v threadtime -T 1  >>
  /path/to/logcat,9C041FQC200408,sunfish,03-17-2020_10-29-23-299.txt: No such file or directory\n'
and end the test with 'Timeout while waiting for logcat file to be created.' error.
(mobly.controllers.android_device_lib.services.logcat.Error)
But the same Mobly PY test can run correctly with normal python run or
'run' with Intelij IDEA.

I provide my solution as
add single quote instead of double quote,
  \'%s\' -s %s logcat -v threadtime -T 1 %s >> \'%s\'
it's reasonable to use single quote here to avoid
substitution, expansion or word splitting for file name and adb path.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/668)

### DIFF
--- a/mobly/controllers/android_device_lib/services/logcat.py
+++ b/mobly/controllers/android_device_lib/services/logcat.py
@@ -292,7 +292,7 @@ class Logcat(base_service.BaseService):
             logcat_file_path = os.path.join(self._ad.log_path, f_name)
             self.adb_logcat_file_path = logcat_file_path
         utils.create_dir(os.path.dirname(self.adb_logcat_file_path))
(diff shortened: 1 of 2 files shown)


eric100lin:
Re-work as new solution.

PR #669, opened 2020-03-18
Fix adb service can NOT start in mobly when debugging with Intelij IDEA on gLinux/Mac/Windows.

eric100lin:
When 'debugging' Mobly PY test with Intelij IDEA on gLinux/Mac/Windows,
logcat service in Mobly can NOT start with the following error
  b'/bin/sh: /path/to/adb -s 9C041FQC200408 logcat -v threadtime -T 1  >>
  /path/to/logcat,9C041FQC200408,sunfish,03-17-2020_10-29-23-299.txt: No such file or directory\n'
and end the test with 'Timeout while waiting for logcat file to be created.' error.
(mobly.controllers.android_device_lib.services.logcat.Error)
But the same Mobly PY test can execute correctly with direct python command or
'run' with Intelij IDEA.

I trace the codes of Popen API and found
there is a "patch_args" in pydev/.../pydev_monkey.py.
  https://github.com/JetBrains/intellij-community/blob/master/python/helpers/pydev/_pydev_bundle/pydev_monkey.py
This function remove double quote in args if arg string
start and end with double quote.
Therefore, I provide this solution
add space at beginning and at last,
  ' "%s" -s %s logcat -v threadtime -T 1 %s >> "%s" '
this break the "patch_args" rule and can
still run correctly on Windows, Linux and MacOS.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/669)

### DIFF
--- a/mobly/controllers/android_device_lib/services/logcat.py
+++ b/mobly/controllers/android_device_lib/services/logcat.py
@@ -292,7 +292,7 @@ class Logcat(base_service.BaseService):
             logcat_file_path = os.path.join(self._ad.log_path, f_name)
             self.adb_logcat_file_path = logcat_file_path
         utils.create_dir(os.path.dirname(self.adb_logcat_file_path))
-        cmd = '"%s" -s %s logcat -v threadtime -T 1 %s >> "%s"' % (
+        cmd = ' "%s" -s %s logcat -v threadtime -T 1 %s >> "%s" ' % (
             adb.ADB, self._ad.serial, self._config.logcat_params,
             self.adb_logcat_file_path)
         process = utils.start_standing_subprocess(cmd, shell=True)
--- a/mobly/controllers/android_device_lib/services/logcat.py
+++ b/mobly/controllers/android_device_lib/services/logcat.py
@@ -292,6 +292,9 @@ class Logcat(base_service.BaseService):
             logcat_file_path = os.path.join(self._ad.log_path, f_name)
             self.adb_logcat_file_path = logcat_file_path
         utils.create_dir(os.path.dirname(self.adb_logcat_file_path))
+        # In debugging mode of IntelijIDEA, "patch_args" remove
+        # double quotes in args if starting and ending with it.
+        # Add spaces at beginning and at last to fix this issue.
         cmd = ' "%s" -s %s logcat -v threadtime -T 1 %s >> "%s" ' % (
             adb.ADB, self._ad.serial, self._config.logcat_params,
             self.adb_logcat_file_path)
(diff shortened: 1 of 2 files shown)


eric100lin:
Mobly with this change, works correctly and test pass on Windows.
[Mobly on Windows test pass and debug with IntelijIDEA screenshot](https://photos.app.goo.gl/u4xUH1h7B9TVqWiBA)

eric100lin:
Mobly with this change, works correctly and test pass on MacOS.
[Mobly on MacOS test pass and debug with IntelijIDEA screenshot](https://photos.app.goo.gl/cZKejNxgGLLL91xU7)

eric100lin:
@mderu @xpconanfan Could I invite you for the review?
I will be very appreciated for your help.

mobly/controllers/android_device_lib/services/logcat.py
```diff
@@ -292,7 +292,7 @@ def _start(self):
             logcat_file_path = os.path.join(self._ad.log_path, f_name)
             self.adb_logcat_file_path = logcat_file_path
         utils.create_dir(os.path.dirname(self.adb_logcat_file_path))
-        cmd = '"%s" -s %s logcat -v threadtime -T 1 %s >> "%s"' % (
+        cmd = ' "%s" -s %s logcat -v threadtime -T 1 %s >> "%s" ' % (
```
xpconanfan: so the only change is adding extra spaces?
does this actually solve your problem?

eric100lin: Yes! Only change is adding extra spaces!
My explanation copy from top of this thread:

> I trace the codes of Popen API and found there is a "patch_args" in pydev/.../pydev_monkey.py.
> https://github.com/JetBrains/intellij-community/blob/master/python/helpers/pydev/_pydev_bundle/pydev_monkey.py
> This function remove double quote in args if arg string start and end with double quote.
> Therefore, I provide this solution add space at beginning and at last,
> ' "%s" -s %s logcat -v threadtime -T 1 %s >> "%s" '
> this break the "patch_args" rule and can still run correctly on Windows, Linux and MacOS.

eric100lin: This solution is really simple and actually solve IntelijIDEA debugging issue.

xpconanfan: Could add an inline comment noting the extra spaces are required?

xpconanfan: Could you add an inline comment noting the extra spaces are required?

eric100lin: Thank you for your suggestion.
I added inline comment.
Thank you.

eric100lin:
@mderu Could I invite you for the review?
I will be very appreciated for your help.

PR #671, opened 2020-04-09
Change an arg name to be consistent.

xpconanfan:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/671)


xpconanfan:
CI passed, but the status is not reported properly
https://travis-ci.org/github/google/mobly/builds/672779215

PR #680, opened 2020-09-02
Use an empty string as the default TestRunConfig.log_path. This fixes…

dthkao:
… an issue When using the --list_tests flag. In that codepath os.path.join is called on this value which has not been set by a TestRunner, resulting in the error.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/680)

### DIFF
--- a/mobly/config_parser.py
+++ b/mobly/config_parser.py
@@ -178,7 +178,7 @@ class TestRunConfig(object):
     """
 
     def __init__(self):
-        self.log_path = None
+        self.log_path = ''
         # Deprecated, use 'testbed_name'
         self.test_bed_name = None
         self.testbed_name = None


xpconanfan (approved):
:lgtm:

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/mobly/680#-:-MGE2WKF7loruoR91RWT:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

mobly/config_parser.py
```diff
@@ -178,7 +178,7 @@ class TestRunConfig(object):
     """
 
     def __init__(self):
-        self.log_path = None
+        self.log_path = ''
```
xpconanfan, line 181: Can you add a comment here specifying why this has to be an empty string?

xpconanfan (changes requested):

xpconanfan:
Fixed in another PR

PR #683, opened 2020-10-15
Sets default not to do verbose logging

syaoranx:
Do not do verbose logging in RPC return messages.
Verbose logging could be enabled by user if you want.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/683)

### DIFF
--- a/mobly/controllers/android_device_lib/jsonrpc_client_base.py
+++ b/mobly/controllers/android_device_lib/jsonrpc_client_base.py
@@ -70,10 +70,6 @@ _SOCKET_CONNECTION_TIMEOUT = 60
 # Maximum time to wait for a response message on the socket.
 _SOCKET_READ_TIMEOUT = callback_handler.MAX_TIMEOUT
 
-# Logging RPC response with full string or truncated by given length
-_VERBOSE_LOGGING = False
-MAX_RPC_RETURN_VALUE_LENGTH = 1024
-
 
 class Error(errors.DeviceError):
     pass
@@ -143,7 +139,7 @@ class JsonRpcClientBase(object):
         self._lock = threading.Lock()
         self._event_client = None
         self.verbose_logging = False   # Logging RPC response with full string or truncated by given length
-        self._max_rpc_return_value_length = 1024
+        self.max_rpc_return_value_length = 1024
 
     def __del__(self):
         self.disconnect()
@@ -389,11 +385,17 @@ class JsonRpcClientBase(object):
             i += 1
 
     def set_snippet_client_verbose_logging(self, verbose=False):
-        # type: (bool) -> None
         """Switches verbose logging. True for logging full RPC response.
 
+        By default it will only write max_rpc_return_value_length for Rpc return
+        strings. If you need to see full message returned from Rpc, please turn
+        on verbose logging.
+
+        max_rpc_return_value_length will set to 1024 by default, the length
+        contains full Rpc response in Json format, included 1st element "id".
+
         Args:
-            verbose: bool, True for full logging full RPC response in DEBUG level.
+            verbose: bool. If True, turns on verbose logging, if False turns off
         """
         self._ad.log.info('Set verbose logging to %s.', verbose)
         self.verbose_logging = verbose
--- a/mobly/controllers/android_device_lib/jsonrpc_client_base.py
+++ b/mobly/controllers/android_device_lib/jsonrpc_client_base.py
@@ -70,6 +70,10 @@ _SOCKET_CONNECTION_TIMEOUT = 60
 # Maximum time to wait for a response message on the socket.
 _SOCKET_READ_TIMEOUT = callback_handler.MAX_TIMEOUT
 
+# Logging RPC response with full string or truncated by given length
+_VERBOSE_LOGGING = False
+MAX_RPC_RETURN_VALUE_LENGTH = 1024
+
 
 class Error(errors.DeviceError):
     pass
@@ -292,7 +296,9 @@ class JsonRpcClientBase(object):
         """
         try:
             response = self._client.readline()
-            self.log.debug('Snippet received: %s', response)
+            self.log.debug('Snippet received: %s',
+                           response if _VERBOSE_LOGGING else
+                           response[:MAX_RPC_RETURN_VALUE_LENGTH])
             return response
         except socket.error as e:
             raise Error(
@@ -377,3 +383,13 @@ class JsonRpcClientBase(object):
         while True:
             yield i
             i += 1
+
+    def set_verbose_logging(self, verbose: bool = False):
+        """Switches verbose logging. True for logging full RPC response.
+
+        Args:
+            verbose: bool, True for full logging full RPC response in DEBUG level.
+        """
+        self._ad.log.info('Set verbose logging to %s.', bool)
+        global _VERBOSE_LOGGING
+        _VERBOSE_LOGGING = verbose
--- a/mobly/controllers/android_device_lib/jsonrpc_client_base.py
+++ b/mobly/controllers/android_device_lib/jsonrpc_client_base.py
@@ -142,6 +142,8 @@ class JsonRpcClientBase(object):
         self._counter = None
         self._lock = threading.Lock()
         self._event_client = None
+        self.verbose_logging = False   # Logging RPC response with full string or truncated by given length
+        self._max_rpc_return_value_length = 1024
 
     def __del__(self):
         self.disconnect()
@@ -296,9 +298,11 @@ class JsonRpcClientBase(object):
         """
         try:
             response = self._client.readline()
-            self.log.debug('Snippet received: %s',
-                           response if _VERBOSE_LOGGING else
-                           response[:MAX_RPC_RETURN_VALUE_LENGTH])
+            if self.verbose_logging:
+                self.log.debug('Snippet received: %s', response)
+            else:
+                self.log.debug('Snippet received: %s',
+                               response[:self._max_rpc_return_value_length])
             return response
         except socket.error as e:
             raise Error(
@@ -384,12 +388,12 @@ class JsonRpcClientBase(object):
             yield i
             i += 1
 
-    def set_verbose_logging(self, verbose: bool = False):
+    def set_snippet_client_verbose_logging(self, verbose=False):
+        # type: (bool) -> None
         """Switches verbose logging. True for logging full RPC response.
 
         Args:
             verbose: bool, True for full logging full RPC response in DEBUG level.
         """
-        self._ad.log.info('Set verbose logging to %s.', bool)
-        global _VERBOSE_LOGGING
-        _VERBOSE_LOGGING = verbose
+        self._ad.log.info('Set verbose logging to %s.', verbose)
+        self.verbose_logging = verbose
--- a/mobly/controllers/android_device_lib/jsonrpc_client_base.py
+++ b/mobly/controllers/android_device_lib/jsonrpc_client_base.py
@@ -70,8 +70,9 @@ _SOCKET_CONNECTION_TIMEOUT = 60
 # Maximum time to wait for a response message on the socket.
 _SOCKET_READ_TIMEOUT = callback_handler.MAX_TIMEOUT
 
-# Maximum logging length of Rpc response in DEBUG level.
-_MAX_RPC_RETURN_LENGTH_IN_DEBUG_LOG = 1024
+# Maximum logging length of Rpc response in DEBUG level when verbose logging is
+# off.
+_MAX_RPC_RESP_LOGGING_LENGTH = 1024
 
 
 class Error(errors.DeviceError):
@@ -297,14 +298,14 @@ class JsonRpcClientBase(object):
         try:
             response = self._client.readline()
             response_log = str(response)
-            if self.verbose_logging or\
-                _MAX_RPC_RETURN_LENGTH_IN_DEBUG_LOG >= len(response):
+            if self.verbose_logging or _MAX_RPC_RESP_LOGGING_LENGTH >= len(
+                response):
               self.log.debug('Snippet received: %s', response)
             else:
               self.log.debug(
                   'Snippet received: %s...(%d bytes are truncated)',
-                  response_log[:_MAX_RPC_RETURN_LENGTH_IN_DEBUG_LOG],
-                  len(response) - _MAX_RPC_RETURN_LENGTH_IN_DEBUG_LOG)
+                  response_log[:_MAX_RPC_RESP_LOGGING_LENGTH],
+                  len(response) - _MAX_RPC_RESP_LOGGING_LENGTH)
             return response
         except socket.error as e:
             raise Error(
@@ -390,7 +391,7 @@ class JsonRpcClientBase(object):
             yield i
             i += 1
 
-    def set_snippet_client_verbose_logging(self, verbose=True):
+    def set_snippet_client_verbose_logging(self, verbose):
         """Switches verbose logging. True for logging full RPC response.
 
         By default it will only write max_rpc_return_value_length for Rpc return
--- a/mobly/controllers/android_device_lib/jsonrpc_client_base.py
+++ b/mobly/controllers/android_device_lib/jsonrpc_client_base.py
@@ -298,7 +298,7 @@ class JsonRpcClientBase(object):
                 self.log.debug('Snippet received: %s', response)
             else:
                 self.log.debug('Snippet received: %s',
-                               response[:self._max_rpc_return_value_length])
+                               response[:int(self.max_rpc_return_value_length)])
             return response
         except socket.error as e:
             raise Error(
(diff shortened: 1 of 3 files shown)


xpconanfan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/683)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @syaoranx)

___
*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 300 at r1](https://reviewable.io/reviews/google/mobly/683#-MJkbxjKC4byl2fCePuM:-MJkbxjKC4byl2fCePuN:bfm7nm9) ([raw file](https://github.com/google/mobly/blob/ea0f24181a175445430d76d9ca1a7933c0d0a4a5/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L300)):*
> ```Python
>             response = self._client.readline()
>             self.log.debug('Snippet received: %s',
>                            response if _VERBOSE_LOGGING else
> ```

do this determination in separate lines instead of all in one line
___
*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 387 at r1](https://reviewable.io/reviews/google/mobly/683#-MJkbjAgDZKxM4HSATwp:-MJkbjAgDZKxM4HSATwq:b-s6y5ew) ([raw file](https://github.com/google/mobly/blob/ea0f24181a175445430d76d9ca1a7933c0d0a4a5/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L387)):*
> ```Python
>             i += 1
> 
>     def set_verbose_logging(self, verbose: bool = False):
> ```

`set_snippet_client_verbose_logging`
___
*[mobly/controllers/android_device_lib/jsonrpc_client_base.py, line 394 at r1](https://reviewable.io/reviews/google/mobly/683#-MJkbbmKCh-ZWQWtMRz7:-MJkbbmL3Lw6El8wJMFR:b-p7lutb) ([raw file](https://github.com/google/mobly/blob/ea0f24181a175445430d76d9ca1a7933c0d0a4a5/mobly/controllers/android_device_lib/jsonrpc_client_base.py#L394)):*
> ```Python
>         """
>         self._ad.log.info('Set verbose logging to %s.', bool)
>         global _VERBOSE_LOGGING
> ```

`global` is not appropriate here.
seems like a simple class var is sufficient?

xpconanfan (changes requested):
Need to add unit test for this behavior as well.
Can mock the response

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -377,3 +383,13 @@ def _id_counter(self):
         while True:
             yield i
             i += 1
+
+    def set_verbose_logging(self, verbose: bool = False):
```
eric100lin, line 387: Need to consider python 2 compatibility.

xpconanfan: Can you be more specific?
What's not py2 compatible here?

eric100lin: Could found `SyntaxError: invalid syntax` in the [py2 CI logs](https://travis-ci.org/github/google/mobly/jobs/735993600) below:    
```
E     File "/home/travis/build/google/mobly/mobly/controllers/android_device_lib/jsonrpc_client_base.py", line 387
E       def set_verbose_logging(self, verbose: bool = False):
E                                            ^
E   SyntaxError: invalid syntax
```
I think maybe need to use this kind of approach to workaround the python 2 compatibility issue:   
```
    def set_snippet_client_verbose_logging(self, verbose = False):
    # type: (bool) -> None
```

By the way, [CI also fail in py3](https://travis-ci.org/github/google/mobly/jobs/735993600) due to "'NoneType' object is not subscriptable":
```
>                          response[:MAX_RPC_RETURN_VALUE_LENGTH])
E                          TypeError: 'NoneType' object is not subscriptable
```

xpconanfan: I see. two things:
1. In this PR, let's not use py type
2. Eric, can you fix the CI asap :)

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -142,6 +142,8 @@ def __init__(self, app_name, ad):
         self._counter = None
         self._lock = threading.Lock()
         self._event_client = None
+        self.verbose_logging = False   # Logging RPC response with full string or truncated by given length
+        self._max_rpc_return_value_length = 1024
```
xpconanfan, line 146: this should remain a const?

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -384,12 +388,12 @@ def _id_counter(self):
             yield i
             i += 1
 
-    def set_verbose_logging(self, verbose: bool = False):
+    def set_snippet_client_verbose_logging(self, verbose=False):
+        # type: (bool) -> None
```
xpconanfan, line 392: not sure what this means...remove?

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -384,12 +388,12 @@ def _id_counter(self):
...
         """Switches verbose logging. True for logging full RPC response.
```
xpconanfan, line 394: need to explain what "verbose" means in this docstring.
Especially need to highlight the difference being whether the `value` field of the response message from snippet server is fully logged.

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -384,12 +388,12 @@ def _id_counter(self):
...
 
         Args:
             verbose: bool, True for full logging full RPC response in DEBUG level.
```
xpconanfan, line 396: this sentence is a bit broken, rephrase.
Something like, "if True, turns on verbose logging; if False, turns off verbose loggin.

xpconanfan (changes requested):

mobly/controllers/android_device_lib/jsonrpc_client_base.py
syaoranx: Well, it could be modify by user, not really need a const.

mobly/controllers/android_device_lib/jsonrpc_client_base.py
syaoranx: Add more description in docstring.

mobly/controllers/android_device_lib/jsonrpc_client_base.py
syaoranx: It's annotation for python 2 capability.
Remove according previous comment

mobly/controllers/android_device_lib/jsonrpc_client_base.py
eric100lin: I created [PR 684](https://github.com/google/mobly/pull/684) with:
* all PASS CI
* unit test cases
* as-is behavior
* configurable `jsonrpc_client_base.SNIPPET_RESPONSE_LOGGING_LENGTH_LIMITATION`

syaoranx: I modify it based on my version with
- Let length in class variable and configurable
- More unit tests

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -292,7 +294,12 @@ def _client_receive(self):
         """
         try:
             response = self._client.readline()
-            self.log.debug('Snippet received: %s', response)
+            response_log = str(response)
+            if self.verbose_logging:
+                self.log.debug('Snippet received: %s', response)
+            else:
+                self.log.debug('Snippet received: %s',
```
xpconanfan, line 301: the truncated log should indicate that it's truncated.
E.g. in Android logcat, truncated msgs follows "blah blah blah...[xxx bytes were truncated]"

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -138,6 +138,8 @@ def __init__(self, app_name, ad):
         self._counter = None
         self._lock = threading.Lock()
         self._event_client = None
+        self.verbose_logging = False   # Logging RPC response with full string or truncated by given length
+        self.max_rpc_return_value_length = 1024
```
xpconanfan, line 142: make this a const

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -138,6 +138,8 @@ def __init__(self, app_name, ad):
         self._counter = None
         self._lock = threading.Lock()
         self._event_client = None
+        self.verbose_logging = False   # Logging RPC response with full string or truncated by given length
```
xpconanfan, line 141: line too long. don't forget to run yapf ;)

tests/lib/jsonrpc_client_test_base.py
```diff
@@ -39,6 +39,9 @@ class JsonRpcClientTestBase(unittest.TestCase):
         b'{"id": 0, "result": 123, "error": null, "status": 1, "uid": 1, '
         b'"callback": "1-0"}')
     MOCK_RESP_WITH_ERROR = b'{"id": 0, "error": 1, "status": 1, "uid": 1}'
+    MOCK_REST_LONG_STRINGS = bytes((
```
xpconanfan, line 42: `bytes('{"id": 0, "result": "' + 'Very long long long long response. ' * 40 + '", "error": null, "status": 0, "callback": null}', 'utf-8')`
we prefer to use this format, but it doesn't work in py2.
add a TODO for py2 deprecation later.

tests/mobly/controllers/android_device_lib/jsonrpc_client_base_test.py
```diff
@@ -228,6 +228,48 @@ def test_rpc_call_increment_counter(self, mock_create_connection):
...
+            client.max_rpc_return_value_length = test_length
+            response = client._client_receive()
+
+            # Response should contain full response.
+            self.assertEqual(response, self.MOCK_REST_LONG_STRINGS)
+
+            # DEBUG level log should truncated by given length.
+            client.log.debug.assert_called_with(
+                'Snippet received: %s',
+                str(self.MOCK_REST_LONG_STRINGS)[:int(test_length)])
```
xpconanfan, line 271: remove `int`

tests/mobly/controllers/android_device_lib/jsonrpc_client_base_test.py
```diff
@@ -228,6 +228,48 @@ def test_rpc_call_increment_counter(self, mock_create_connection):
...
+        fake_file = self.setup_mock_socket_file(mock_create_connection)
+        fake_file.resp = self.MOCK_REST_LONG_STRINGS
+
+        test_truncate_lengths = [1024, 100, 10000]
+
+        client = FakeRpcClient()
+        client.connect()
+        client.set_snippet_client_verbose_logging(False)
+
+        for test_length in test_truncate_lengths:
```
xpconanfan, line 261: split this into separate test cases.

xpconanfan (changes requested):

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -295,11 +297,14 @@ def _client_receive(self):
         try:
             response = self._client.readline()
             response_log = str(response)
-            if self.verbose_logging:
-                self.log.debug('Snippet received: %s', response)
+            if self.verbose_logging or\
```
xpconanfan, line 300: avoid using `\`

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -70,6 +70,9 @@
 # Maximum time to wait for a response message on the socket.
 _SOCKET_READ_TIMEOUT = callback_handler.MAX_TIMEOUT
 
+# Maximum logging length of Rpc response in DEBUG level.
```
xpconanfan, line 73: this does not seem accurate
should be something like "when verbose logging is off`?

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -70,6 +70,9 @@
...
+_MAX_RPC_RETURN_LENGTH_IN_DEBUG_LOG = 1024
```
xpconanfan, line 74: In debug log is less relevant than verbose logging is off right?
the name of this const should be modified

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -385,7 +390,7 @@ def _id_counter(self):
             yield i
             i += 1
 
-    def set_snippet_client_verbose_logging(self, verbose=False):
+    def set_snippet_client_verbose_logging(self, verbose=True):
```
xpconanfan, line 393: i don't think we need a default value for `verbose`
since this is the setter function, which does not determine the default behavior, right?

xpconanfan (changes requested):

mobly/controllers/android_device_lib/jsonrpc_client_base.py
syaoranx: Change to "_MAX_RPC_RESP_LOGGING_LENGTH"

xpconanfan: I don't see any changes, did you push the commit?

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -297,14 +298,14 @@ def _client_receive(self):
         try:
             response = self._client.readline()
             response_log = str(response)
-            if self.verbose_logging or\
-                _MAX_RPC_RETURN_LENGTH_IN_DEBUG_LOG >= len(response):
+            if self.verbose_logging or _MAX_RPC_RESP_LOGGING_LENGTH >= len(
```
xpconanfan, line 301: I don't see why we need the `or  _MAX_RPC_RESP_LOGGING_LENGTH...` part.
if verbose is on, we simply log everything right?

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -297,14 +298,14 @@ def _client_receive(self):
...
+                response):
               self.log.debug('Snippet received: %s', response)
             else:
               self.log.debug(
                   'Snippet received: %s...(%d bytes are truncated)',
```
xpconanfan, line 306: "...chars are truncated"

xpconanfan (changes requested):

mobly/controllers/android_device_lib/jsonrpc_client_base.py
syaoranx: If verbose is on, we simply log everything, that's right.

And if the verbose is off, but the rpc return less then 1024 chars, we also log everything, that's the `or _MAX_RPC_RESP_LOGGING_LENGTH...` part do

xpconanfan: You should explicitly check for this only when verbose logging is off so the logic is more clear?

xpconanfan: Or just not check it at all?

mobly/controllers/android_device_lib/jsonrpc_client_base.py
syaoranx: Sorry I forgot to commit it before leave the comment. New changes are commit.

mobly/controllers/android_device_lib/jsonrpc_client_base.py
```diff
@@ -292,7 +297,17 @@ def _client_receive(self):
         """
         try:
             response = self._client.readline()
-            self.log.debug('Snippet received: %s', response)
+            response_log = str(response)
+            if self.verbose_logging:
+                self.log.debug('Snippet received: %s', response)
+            else:
+                if _MAX_RPC_RESP_LOGGING_LENGTH >= len(response):
```
xpconanfan, line 304: i don't think this check is needed
does any of the unit test fail if you remove this `if` branch?

xpconanfan (changes requested):

mobly/controllers/android_device_lib/jsonrpc_client_base.py
syaoranx: Yes, 2 tests are failed.
For example, if response length is 512, it will display 
`Snippet received: ...... **_-512_** chars are truncated`
because the length of response is less then `_MAX_RPC_RESP_LOGGING_LENGTH`.

xpconanfan (approved):

PR #684, opened 2020-10-19
Make snippet response logging length can be restricted.

eric100lin:
A revised version of PR 683.     
We keep the behavior as-is and make snippet response logging length can be restricted.    
User can set `jsonrpc_client_base.SNIPPET_RESPONSE_LOGGING_LENGTH_LIMITATION` through module wide variable.    

This is similar to usage of `adb.DEFAULT_GETPROP_TIMEOUT_SEC`:    
https://github.com/google/mobly/blob/master/mobly/controllers/android_device_lib/adb.py#38

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/684)


xpconanfan:
Combining the points in this PR into #683

PR #700, opened 2021-01-09
Reformat python files of mobly tests/lib.

eric100lin:
Make the indentation style of mobly consistent with that of google pyformat.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/700)

### DIFF
--- a/tests/lib/integration2_test.py
+++ b/tests/lib/integration2_test.py
@@ -18,10 +18,11 @@ from tests.lib import integration_test
 
 
 class Integration2Test(integration_test.IntegrationTest):
-    """Same as the IntegrationTest class, created this so we have two
-    'different' test classes to use in unit tests.
-    """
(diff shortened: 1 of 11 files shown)


tests/lib/integration2_test.py
```diff
@@ -14,14 +14,15 @@
 
 from mobly import test_runner
 
-from tests.lib import integration_test
```
xpconanfan, line 17: why is this line changed?
we should strictly do indentation change first, then fix tidbits

xpconanfan (changes requested):

tests/lib/integration2_test.py
eric100lin: This line is accidentally changed. I reverted it.

eric100lin:
Rework and create another new PR.

PR #719, opened 2021-02-11
Make quote delimiter consistent in mobly/utils.py.

eric100lin:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/719)

### DIFF
--- a/mobly/utils.py
+++ b/mobly/utils.py
@@ -39,34 +39,34 @@ MAX_FILENAME_LEN = 255
 MAX_PORT_ALLOCATION_RETRY = 50
 
 ascii_letters_and_digits = string.ascii_letters + string.digits
-valid_filename_chars = "-_." + ascii_letters_and_digits
+valid_filename_chars = f'-_.{ascii_letters_and_digits}'
 
 GMT_to_olson = {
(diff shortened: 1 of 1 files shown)


mobly/utils.py
```diff
@@ -291,8 +290,8 @@ def concurrent_exec(func, param_list, max_workers=30, raise_on_exception=False):
       try:
         return_vals.append(future.result())
       except Exception as exc:
-        logging.exception("{} generated an exception: {}".format(
-            params, traceback.format_exc()))
+        logging.exception('%s generated an exception: %s', params,
+                          traceback.format_exc())
```
xpconanfan, line 294: can you double check to see if the `traceback.format_exc()` is still needed in py3?

xpconanfan: Should probably fix in a separate PR. Let's merge this one :)

xpconanfan (approved):

mobly/utils.py
eric100lin: I think this is still needed:
https://docs.python.org/3/library/traceback.html#traceback.format_exc

PR #725, opened 2021-02-14
Prefer monotonic time from perf_counter for time measurement.

eric100lin:
Reference:
https://docs.python.org/3/library/time.html#time.perf_counter
https://www.webucator.com/blog/2015/08/python-clocks-explained/

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/725)

### DIFF
--- a/mobly/controllers/android_device_lib/snippet_client.py
+++ b/mobly/controllers/android_device_lib/snippet_client.py
@@ -173,7 +173,7 @@ class SnippetClient(jsonrpc_client_base.JsonRpcClientBase):
     cmd = _LAUNCH_CMD.format(shell_cmd=persists_shell_cmd,
                              user=self._get_user_command_string(),
                              snippet_package=self.package)
-    start_time = time.monotonic()
+    start_time = time.perf_counter()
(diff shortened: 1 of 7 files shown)


mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -173,7 +173,7 @@ def _start_app_and_connect(self):
     cmd = _LAUNCH_CMD.format(shell_cmd=persists_shell_cmd,
                              user=self._get_user_command_string(),
                              snippet_package=self.package)
-    start_time = time.time()
+    start_time = time.monotonic()
```
xpconanfan, line 176: do we want to use `perf_counter` instead since it's measuring time diff here?

xpconanfan (changes requested):

mobly/controllers/android_device_lib/snippet_client.py
eric100lin: `time.perf_counter()` would change about 33,491 times more often than `time.monotonic()`, it's more accurate.

xpconanfan (approved):

PR #736, opened 2021-04-26
Use "GitHub actions" to run CI tests for google/mobly

eric100lin:
Because Travis CI seems down and we can't get CI signal for google/mobly.     
We want to replace the "Travis CI" with "GitHub actions".
It not only can cover Linux but also Windows and MacOS with simple config files.
This PR add "GitHub actions" and delete "Travis CI" in the next PR.

The sample run results can be found at:    
https://github.com/eric100lin/mobly/runs/2436275239

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/736)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: [3.6, 3.7, 3.8]
     steps:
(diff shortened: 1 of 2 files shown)


xpconanfan (dismissed):

xpconanfan:
Let's also remove Travid CI configs in a separate PR

PR #762, opened 2021-08-16
Inspect class instead of class instance to get test methods list.

xianyuanjia:
This prevents inspect.getmembers from calling the properties of given test class. 

e.g.
```
>>>class Foo():
...   @property
...   def bar(self):
...       print('Called')

>>> inspect.getmembers(Foo(), callable)
Called
[...]
>>> getattr(Foo(), 'bar')
Called
'baz'

>>> inspect.getmembers(Foo, callable)
[...]
>>> getattr(Foo, 'bar')
<property object at 0x7f7c7463cb30>
```

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/762)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -905,7 +905,7 @@ class BaseTestClass:
       A list of strings, each is a test method name.
     """
     test_names = []
-    for name, _ in inspect.getmembers(self, callable):
+    for name, _ in inspect.getmembers(type(self), callable):
       if name.startswith('test_'):
         test_names.append(name)
     return test_names + list(self._generated_test_table.keys())
--- a/tests/mobly/base_test_test.py
+++ b/tests/mobly/base_test_test.py
@@ -274,6 +274,21 @@ class BaseTestTest(unittest.TestCase):
     mock_decorated.assert_called_once_with('test_decorated')
     mock_undecorated.assert_called_once_with('test_undecorated')
 
+  def test_get_existing_tests_do_not_call_properties(self):
+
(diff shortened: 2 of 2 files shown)


xpconanfan (changes requested):
Can we add a test for the use case we are trying to enable here?

xianyuanjia (commented):
Done.

> *[Reviewable](https://reviewable.io/reviews/google/mobly/762#-:-Mh_D4YYDcDBq5wrUq8u:b-896fix)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @eric100lin)

eric100lin (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/mobly/762#-:-MhkhksQ1q0u1wW-ikm1:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @xianyuanjia)

eric100lin:
According to the cpython source codes:    
https://github.com/python/cpython/blob/3.9/Lib/inspect.py#L351

When using inspect with class instance:`inspect.getmembers(Foo(), callable)`
It will call `getattr(foo_instance, key)` on all keys from `dir(foo_instance)`, therefore the `bar` property getter will be called.

When using  inspect with class type: `inspect.getmembers(Foo, callable)`
It will call `getattr(Foo, key)` on all keys from `dir(Foo)`, the `bar` property getter will not be called.

xpconanfan (approved):

PR #768, opened 2021-11-04
Add the ACTS sl4a library to Mobly.

xianyuanjia:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/768)


xpconanfan:
This is obsolete as we are discouraging the use of sl4a as is.
One can build Mobly snippet with Android sys privilege to achieve similar effect as sl4a.

PR #786, opened 2021-12-23
Add a test case to cover kill subprocess and descendants behavior.

eric100lin:
Add this test case before doing the psutil dependency removal
for utils.stop_standing_subprocess.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/786)

PR #787, opened 2021-12-23
Replace psutil usages in utils.stop_standing_subprocess.

eric100lin:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/787)

This PR is based on PR #786 .

### DIFF
--- a/mobly/utils.py
+++ b/mobly/utils.py
@@ -258,7 +258,7 @@ def rand_ascii_str(length):
 
 
 # Thead/Process related functions.
-def collect_process_tree(starting_pid):
+def _collect_process_tree(starting_pid):
   """Collects PID list of the descendant processes from the given PID.
 
   This function only available on Unix like system.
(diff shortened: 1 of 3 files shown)


mobly/utils.py
```diff
@@ -257,6 +258,75 @@ def rand_ascii_str(length):
...
+    A list of pid of the descendant processes.
+  """
+  ret = []
+  stack = [starting_pid]
+
+  while stack:
+    pid = stack.pop()
+    try:
+      ps_results = subprocess.check_output([
+          'ps',
```
xpconanfan, line 279: I see a couple of raw commands added.
do these commands work on windows?
and are they the same between mac and linux?

mobly/utils.py
```diff
@@ -257,6 +258,75 @@ def rand_ascii_str(length):
 
 
 # Thead/Process related functions.
+def collect_process_tree(starting_pid):
```
xpconanfan, line 261: make this module "private"?

xpconanfan (changes requested):

mobly/utils.py
eric100lin: Done

mobly/utils.py
eric100lin: As the doc string, this function only available on Unix like system.
The command works on Unix like system as you can see all tests pass on Windows, Mac and Linux.

xpconanfan (approved):

Issue #793, opened 2022-02-08
Refactor Snippet Client


mhaoli:
Mobly supports testing devices of multiple platforms, including Android, iOS and Windows. Mobly has one snippet client for each platform, and these clients share common patterns. So we want to consolidate current clients that they inherit from the same base class.

Currently, the father class of Android snippet client has too much Android-specific code, so it's better to add a new general base class. We will add a base class and one client V2 for each platform and we won't modify existing clients.

This issue is also a prerequisite of making iOS related code open source.

PR #800, opened 2022-03-06
Add a flag for switching to use snippet client v2

mhaoli:
This PR is a part of #793. 

This PR adds a flag for switching to use snippet client v2, including:
* Adds a command line argument `use_mobly_snippet_client_v2`.
* If the argument is specified, the flag will be passed to `SnippetManagementService`, which is responsible for creating, stopping and resuming snippet client.

For this switch, this PR only sets `SnippetManagementService._use_client_v2` to True instead of really using snippet client v2, because client v2 is not ready now.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/800)

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -883,7 +883,7 @@ class AndroidDevice:
       if k == config_parser.USE_SNIPPET_CLIENT_V2:
         # This is a special config related to snippet client, do not set it
         # to AndroidDevice itself.
-        self._set_snippet_client_v2(v)
+        self._set_snippet_client_v2_flag(v)
         continue
 
       if hasattr(self, k) and k not in _ANDROID_DEVICE_SETTABLE_PROPS:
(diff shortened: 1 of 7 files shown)


mobly/controllers/android_device.py
```diff
@@ -879,12 +880,22 @@ def load_config(self, config):
       Error: The config is trying to overwrite an existing attribute.
     """
     for k, v in config.items():
+      if k == config_parser.USE_SNIPPET_CLIENT_V2:
```
xpconanfan, line 883: this is way too complicated../
I don't think a complete new channel needs to be added just to pass this one param..
Can we not simplify this to minimize the amount of code changes made in the existing code?

xpconanfan (changes requested):

mobly/controllers/android_device.py
mhaoli: Yes, I find that the `BaseTest` can directly set the param to `SnippetManagementService`, adding a new PR #803. 

That PR mainly modifies 2 places. Although it still has 140 lines of addition, but most of them are unit tests and comments.

xpconanfan:
Should this PR be closed now?

mhaoli:
Yes, please close this PR. Thanks.

PR #803, opened 2022-03-14
Add a flag for switching to use snippet client v2 - Simpler Version 

mhaoli:
This PR is a part of #793. 

This PR adds a flag for switching to use snippet client v2, including:
* Adds a command line argument `use_mobly_snippet_client_v2`.
* If the arg is specified, the `BaseTest` class will directly set the flag to `SnippetManagementService`, which is responsible for creating, stopping and resuming snippet client.

For this switch, this PR only sets `SnippetManagementService._use_client_v2` to True instead of really using snippet client v2, because client v2 is not ready now.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/803)


xpconanfan:
Discussed offline. Try a simpler approach

PR #807, opened 2022-04-11
Refactor snippet client v2's stop server method

mhaoli:
There are two methods `stop_server` and `do_stop_server` in `ClientBase`, which sounds strange.

This is because `stop_server` is the proxy function of `do_stop_server` but both of them should be public methods. 

Currently, compared with `do_stop_sever`, `stop_server` is only responsible for printing log information. So we can remove the proxy function for now. 

We may refine this to use a decorator or something else when we need to collect metadata of snippet clients.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/807)

### DIFF
--- a/mobly/snippet/client_base.py
+++ b/mobly/snippet/client_base.py
@@ -420,7 +420,7 @@ class ClientBase(abc.ABC):
 
   @abc.abstractmethod
   def stop(self):
-    """Stops the snippet server and releases all the requested resources."""
+    """Releases all the resources acquired in `initialize`."""
 
   @abc.abstractmethod
   def close_connection(self):
--- a/mobly/snippet/client_base.py
+++ b/mobly/snippet/client_base.py
@@ -62,8 +62,8 @@ class ClientBase(abc.ABC):
 
   Connects to a remote device running a JSON RPC compatible server. Users call
   the function `start_server` to start the server on the remote device before
-  sending any RPC. After sending all RPCs, users call the function `stop_server`
-  to stop all the running instances.
+  sending any RPC. After sending all RPCs, users call the function `stop`
+  to stop the snippet server and release all the requested resources.
 
   Attributes:
     package: str, the user-visible name of the snippet library being
@@ -135,11 +135,11 @@ class ClientBase(abc.ABC):
           'Error occurred trying to start and connect to the snippet server '
           'of %s.', self.package)
       try:
-        self.stop_server()
+        self.stop()
       except Exception:  # pylint: disable=broad-except
         # Only prints this exception and re-raises the original exception
         self.log.exception(
-            'Failed to stop the snippet server of %s after failure to start '
+            'Failed to stop the snippet package %s after failure to start '
             'and connect.', self.package)
 
       raise
@@ -419,8 +419,8 @@ class ClientBase(abc.ABC):
     """
 
   @abc.abstractmethod
-  def stop_server(self):
-    """Kills any running instance of the server."""
+  def stop(self):
+    """Stops the snippet server and releases all the requested resources."""
 
   @abc.abstractmethod
   def close_connection(self):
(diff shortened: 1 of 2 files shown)


mobly/snippet/client_base.py
```diff
@@ -418,14 +418,8 @@ def handle_callback(self, callback_id, ret_value, rpc_func_name):
...
 
-  def stop_server(self):
-    """Proxy function of do_stop_server."""
-    self.log.debug('Stopping snippet %s.', self.package)
-    self.do_stop_server()
-    self.log.debug('Snippet %s stopped.', self.package)
-
   @abc.abstractmethod
-  def do_stop_server(self):
+  def stop_server(self):
```
xpconanfan, line 422: should this only kill the server or should this also clean up any socket/connection on the local host?

xpconanfan (changes requested):

mobly/snippet/client_base.py
mhaoli: It also kills connection. Actually, it releases all the requested resources. 

So may be better to rename it to `stop` or `kill`?

xpconanfan: The docstring should clarify this?

mhaoli: Modified.

mobly/snippet/client_base.py
```diff
@@ -418,15 +418,9 @@ def handle_callback(self, callback_id, ret_value, rpc_func_name):
...
-    """Proxy function of do_stop_server."""
-    self.log.debug('Stopping snippet %s.', self.package)
-    self.do_stop_server()
-    self.log.debug('Snippet %s stopped.', self.package)
-
   @abc.abstractmethod
-  def do_stop_server(self):
-    """Kills any running instance of the server."""
+  def stop(self):
+    """Stops the snippet server and releases all the requested resources."""
```
xpconanfan, line 423: Something like "releases all the resource acquired in `start_server`"?

mobly/snippet/client_base.py
```diff
@@ -418,15 +418,9 @@ def handle_callback(self, callback_id, ret_value, rpc_func_name):
...
-  def stop_server(self):
-    """Proxy function of do_stop_server."""
-    self.log.debug('Stopping snippet %s.', self.package)
-    self.do_stop_server()
-    self.log.debug('Snippet %s stopped.', self.package)
-
   @abc.abstractmethod
-  def do_stop_server(self):
-    """Kills any running instance of the server."""
+  def stop(self):
```
xpconanfan, line 422: I'm not sure about changing this name, since it pairs with `start_server`?
If we want to change this, should we change `start_server` as well?
Would it be easier to just keep it as `stop_server`?

xpconanfan (changes requested):

mobly/snippet/client_base.py
mhaoli: Done. Changed `start_server` to `initialize`.

mobly/snippet/client_base.py
xpconanfan: ok, so this is really `clean_up`, which calls `stop_server`? lol
If we have a dedicated `start_server` as part of `initialize` in the interface, should we have a `stop_server` as part of `clean_up`?..
Basically, 有始有终？

Or is it not worth the trouble?

mhaoli: There will be a `stop_server` in the subclass implementation. 

But I think we do not need to have it as an interface in base class. As we decided to make `clean_up` as an abstract method, it would be strange to have two abstract methods `clean_up` and `stop_server`, and their comments reminds that subclass must call `stop_server` in the implementation of `clean_up`.

We can just clarify that `clean_up` should clean all the acquired resources. Does that make sense?

Also, do you suggest to rename this function to `clean_up`?

xpconanfan: ok, let's follow the current plan

xpconanfan (approved):

PR #808, opened 2022-04-20
Add Snippet Client V2 for Android, Step 2

mhaoli:
Part of issue #793, the second step of adding snippet client v2. The previous step is PR #804.

This step includes building socket connection, sending RPC and restoring connection after device got reconnected.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/808)

### DIFF
--- a/mobly/controllers/android_device_lib/snippet_client_v2.py
+++ b/mobly/controllers/android_device_lib/snippet_client_v2.py
@@ -457,7 +457,7 @@ class SnippetClientV2(client_base.ClientBase):
     """Receives the server's response of an RPC message.
 
     Returns:
-      Raw byte string of the response.
+      Raw bytes of the response.
 
     Raises:
       errors.Error: if a socket error occurred during the read.
@@ -470,6 +470,17 @@ class SnippetClientV2(client_base.ClientBase):
           f'Encountered socket error "{e}" reading RPC response') from e
 
   def _decode_socket_response_bytes(self, response):
+    """Returns a string decoded from the socket response bytes.
+
+    Args:
+      response: bytes, the response to be decoded.
+
+    Returns:
+      The string decoded from the given bytes.
+
+    Raises:
+      UnicodeError: if failed to decode the given bytes using encoding utf8.
+    """
     try:
       return str(response, encoding='utf8')
     except UnicodeError:
--- a/mobly/controllers/android_device_lib/snippet_client_v2.py
+++ b/mobly/controllers/android_device_lib/snippet_client_v2.py
@@ -435,17 +435,33 @@ class SnippetClientV2(client_base.ClientBase):
                                  errors.ProtocolError.NO_RESPONSE_FROM_SERVER)
     return self._decode_socket_response_bytes(response)
 
-  def _client_send(self, request):
+  def _client_send(self, message):
+    """Sends an RPC message through the connection.
+
+    Args:
+      msg: str, the message to send.
+
+    Raises:
+      errors.Error: if a socket error occurred during the send.
+    """
     try:
-      self._client.write(f'{request}\n'.encode('utf8'))
+      self._client.write(f'{message}\n'.encode('utf8'))
       self._client.flush()
     except socket.error as e:
       raise errors.Error(
           self._device,
-          f'Encountered socket error "{e}" sending RPC message "{request}"'
+          f'Encountered socket error "{e}" sending RPC message "{message}"'
       ) from e
 
   def _client_receive(self):
+    """Receives the server's response of an RPC message.
+
+    Returns:
+      Raw byte string of the response.
+
+    Raises:
+      errors.Error: if a socket error occurred during the read.
+    """
     try:
       return self._client.readline()
     except socket.error as e:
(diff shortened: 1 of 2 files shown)


xpconanfan:
Resolve conflicts pls

mhaoli:
Resolved.

mobly/controllers/android_device_lib/snippet_client_v2.py
```diff
@@ -58,6 +61,31 @@
...
+UNKNOWN_UID = -1
+
+# Maximum time to wait for the socket to open on the device.
+_SOCKET_CONNECTION_TIMEOUT = 60
+
+# Maximum time to wait for a response message on the socket.
+_SOCKET_READ_TIMEOUT = callback_handler.MAX_TIMEOUT
+
+
+class ConnectionHandshakeCommand:
```
xpconanfan, line 75: should this be a str enum class?

mobly/controllers/android_device_lib/snippet_client_v2.py
```diff
@@ -276,24 +316,242 @@ def _read_protocol_line(self):
...
+
+    Raises:
+      errors.ProtocolError: something went wrong when sending the handshake
+        request.
+    """
+    try:
+      request = json.dumps({'cmd': cmd, 'uid': uid})
+      self.log.debug('Sending handshake request %s.', request)
+      response = self.send_rpc_request(request)
+    except errors.ProtocolError as e:
```
xpconanfan, line 396: since we own `send_rpc_request`, why do we need to catch what it raise then reraise, and check what it returns then raise?

seems like we can just do the checks in `send_rpc_request` itself?

xpconanfan (changes requested):

mobly/controllers/android_device_lib/snippet_client_v2.py
mhaoli: Done.

mobly/controllers/android_device_lib/snippet_client_v2.py
mhaoli: Done.

But instead of doing checks in the `send_rpc_request` method, I let `send_handshake_request` and `send_rpc_request` raise their own exceptions.

mobly/controllers/android_device_lib/snippet_client_v2.py
```diff
@@ -322,27 +613,72 @@ def _stop_server(self):
...
-  def restore_server_connection(self, port=None):
-    raise NotImplementedError('To be implemented.')
+    This function restores the connection of the previously created event
+    client, or creates a new client and makes a connection if it didn't
+    exist before.
+
+    The event client to restore reuses the same host port and device port
+    with the client on which function is called.
+    """
+    if not self._event_client:
```
xpconanfan, line 661: Not clear why there is a branching here.
In which situations would `_event_client` be None or not None?

mobly/controllers/android_device_lib/snippet_client_v2.py
```diff
@@ -276,24 +317,274 @@ def _read_protocol_line(self):
...
+    Args:
+      uid: int, the uid of the server session to continue. It will be ignored
+        if the `cmd` requires the server to create a new session.
+      cmd: ConnectionHandshakeCommand, the handshake command Enum for the
+        server, which requires the server to create a new session or use the
+        current session.
+    """
+    self._event_client.host_port = self.host_port
+    self._event_client.device_port = self.device_port
+    self._event_client._counter = self._event_client._id_counter()  # pylint: disable=protected-access
```
xpconanfan, line 546: there are many other access of private members, but only this one has `pylint` comment.
is this necessary?

xpconanfan (changes requested):

mobly/controllers/android_device_lib/snippet_client_v2.py
mhaoli: Here I followed the same branch as in client v1. `self._event_client` is None until the snippet client has sent the first async RPC. After that, `self._event_client` won't be None.

mobly/controllers/android_device_lib/snippet_client_v2.py
mhaoli: Yes, it's necessary. Because the `_counter` is a private member of the event client object, I think elsewhere we only access the private member of `self`. 

Also verified that gpylint will warn about it if delete this pylint comment.

xpconanfan: if `_id_counter` needs to be accessed, why not just make it "public"?

mobly/controllers/android_device_lib/snippet_client_v2.py
xpconanfan: If the `_event_client` is not even initialized, why do you need to create it in restoration? shouldn't the restoration just be no-op?

mobly/controllers/android_device_lib/snippet_client_v2.py
mhaoli: Seems that `_counter` is better to be private as it is just the counter of RPC ID and users should not care about it. So refactored this function to avoid access to this private member.

mobly/controllers/android_device_lib/snippet_client_v2.py
mhaoli: Removed this branch. 

It seems more reasonable to be no-op. Also checked the git PR which introduced this logic and didn't find any reason to create it here.

xpconanfan (approved):

PR #810, opened 2022-04-21
Fix typo in run_command

nkprasad12:
Currently, if there's a timeout, the `cmd` field of the exception is set to `cwd`, which results in confusing error messages like:

```
Command 'None' timed out after 20 seconds
```

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/810)


xpconanfan (changes requested):
Let's add a unit test for this

nkprasad12:
Added a unit test. The new unit test would fail on the old run_command because it just says "None".

xpconanfan (approved):

PR #812, opened 2022-04-28
Update tutorial.md

dush1729:
remove expect field from tutorial.md because these is no testcase order given via `--test_case`

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/812)


xpconanfan:
The order is actually deterministic because `inspect.getmembers` returns an ordered list.
So the order of the expected results just need to be flipped.

PR #813, opened 2022-05-05
ClientBase should call "stop" if "before_starting_server" failed

mhaoli:
Part of issue #793. This change **is related to** the implementation of **Windows client**.

When implementing the Windows client, we found that we need to improve some behaviors of `ClientBase`.

**Changed content**: We modified the `ClientBase#initialize` function. If error occurs at any stage, this function will abort the initialization process and call `stop` to clean up.

**Reason for the change**: Previously we thought that if the `ClientBase#before_starting_server` stage fails, `ClientBase` doesn't need to call `stop` to clean up. But this is wrong, because `before_starting_server` may also perform some operations that need to be cleaned up.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/813)

### DIFF
--- a/mobly/snippet/client_base.py
+++ b/mobly/snippet/client_base.py
@@ -101,6 +101,9 @@ class ClientBase(abc.ABC):
       2. starting the snippet server on the remote device.
       3. making a connection to the snippet server.
 
+    If error occurs at any stage, this function will abort the initialization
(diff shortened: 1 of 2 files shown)


mhaoli:
Resolved.

xpconanfan (approved):

PR #815, opened 2022-05-11
Add a flag for whether to use Snippet Client V2

mhaoli:
Part of #793.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/815)

### DIFF
--- a/mobly/controllers/android_device_lib/services/snippet_management_service.py
+++ b/mobly/controllers/android_device_lib/services/snippet_management_service.py
@@ -42,7 +42,7 @@ class SnippetManagementService(base_service.BaseService):
 
   @property
   def is_using_client_v2(self):
-    """Does this service use snippet client V2."""
+    """Is this service using snippet client V2."""
     if self._is_using_client_v2 is None:
       device_dimensions = getattr(self._device, 'dimensions', {})
       self._is_using_client_v2 = (device_dimensions.get(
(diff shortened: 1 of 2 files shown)


mobly/controllers/android_device_lib/services/snippet_management_service.py
```diff
@@ -37,6 +38,16 @@ def __init__(self, device, configs=None):
     self._is_alive = False
     self._snippet_clients = {}
     super().__init__(device)
+    self._is_using_client_v2 = None
+
+  @property
+  def is_using_client_v2(self):
```
xpconanfan, line 44: can we name this the "private" way since we probably don't want users to call this.

mobly/controllers/android_device_lib/services/snippet_management_service.py
```diff
@@ -37,6 +38,16 @@ def __init__(self, device, configs=None):
...
+    """Is this service using snippet client V2."""
```
xpconanfan, line 45: make it clear that this is transient and will be removed after migration is complete.

xpconanfan (changes requested):

mobly/controllers/android_device_lib/services/snippet_management_service.py
mhaoli: Done.

mobly/controllers/android_device_lib/services/snippet_management_service.py
mhaoli: Done.

xpconanfan (approved):

xpconanfan:
Force merging since tests for 3.6 are removed.

PR #817, opened 2022-05-12
Create base suite class and add run_suite_class to suite runner.

booneng:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/817)

### DIFF
--- a/mobly/base_suite.py
+++ b/mobly/base_suite.py
@@ -1,4 +1,4 @@
-# Copyright 2017 Google Inc.
+# Copyright 2022 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/mobly/base_suite.py
+++ b/mobly/base_suite.py
@@ -21,6 +21,13 @@ class BaseSuite(abc.ABC):
 
   Use `BaseSuite.add_test_class` to specify which classes to run with which
   configs and test selectors.
+
+  After defining the sub class, the suite can be executed using
+  suite_runner.run_suite_class.
+
+  Users can use this class if they need to define their own setup and teardown
+  steps on the suite level. Otherwise, just use suite_runner.run_suite on the
+  list of test classes.
   """
 
   def __init__(self, runner, config):
--- a/mobly/base_suite.py
+++ b/mobly/base_suite.py
@@ -1,0 +1,66 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import abc
+
+class BaseSuite(abc.ABC):
+  """Class used to define a Mobly suite.
+
+  To create a suite, inherit from this class and implement setup_suite.
+
+  Use `BaseSuite.add_test_class` to specify which classes to run with which
+  configs and test selectors.
+  """
+
+  def __init__(self, runner, config):
+    self._runner = runner
+    self._config = config.copy()
+
+  @property
+  def user_params(self):
+    return self._config.user_params
+
+  def add_test_class(self, clazz, config=None, tests=None, name_suffix=None):
+    """Adds a test class to the suite.
+
+    Args:
+      clazz: class, a Mobly test class.
+      config: config_parser.TestRunConfig, the config to run the class with. If
+        not specified, the default config passed from google3 infra is used.
+      tests: list of strings, names of the tests to run in this test class, in
+        the execution order. If not specified, all tests in the class are
+        executed.
+      name_suffix: string, suffix to append to the class name for reporting.
+        This is used for differentiating the same class executed with different
+        parameters in a suite.
+    """
+    if not config:
+      config = self._config
+    self._runner.add_test_class(config, clazz, tests, name_suffix)
+
+  @abc.abstractmethod
+  def setup_suite(self, config):
+    """Function used to add test classes, has to be implemented by child class.
+
+    Args:
+      config: config_parser.TestRunConfig, the config provided by google3 infra.
+
+    Raises:
+      Error: when setup_suite is not implemented by child class.
+    """
+    pass
+
+  def teardown_suite(self):
+    """Function used to add post tests cleanup tasks (optional)."""
+    pass
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -140,7 +140,7 @@ def run_suite_class(argv=None):
   """
   if argv is None:
     argv = sys.argv
-  cli_args = _parse_cli_args(args)
+  cli_args = _parse_cli_args(argv)
   test_configs = config_parser.load_test_config_file(cli_args.config)
   config_count = len(test_configs)
   if config_count != 1:
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -132,41 +132,6 @@ def _find_suite_class():
   return test_suites[0]
 
 
-class _SuiteRunner(object):
-  """Class used to execute a Mobly suite in google3."""
-
-  def __init__(self, suite, args):
-    cli_args = _parse_cli_args(args)
-    test_configs = config_parser.load_test_config_file(cli_args.config)
-    config_count = len(test_configs)
-    if config_count != 1:
-      logging.error('Expect exactly one test config, found %d', config_count)
-    self._config = test_configs[0]
-    self._runner = test_runner.TestRunner(
-        log_dir=self._config.log_path, testbed_name=self._config.testbed_name)
-    self._suite = suite(self._runner, self._config)
-
-  def run(self):
-    """Runs the test suite.
-
-    This should not be called by users directly.
-    """
-    ok = False
-    with self._runner.mobly_logger():
-      try:
-        self._suite.setup_suite(self._config.copy())
-        try:
-          self._runner.run()
-          ok = self._runner.results.is_all_pass
-          print(ok)
-        except signals.TestAbortAll:
-          pass
-      finally:
-        self._suite.teardown_suite()
-    if not ok:
-      sys.exit(1)
-
-
 def run_suite_class(argv=None):
   """Executes tests in the test suite.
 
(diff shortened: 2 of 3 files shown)


mobly/base_suite.py
```diff
@@ -0,0 +1,66 @@
...
+      config: config_parser.TestRunConfig, the config provided by google3 infra.
+
+    Raises:
+      Error: when setup_suite is not implemented by child class.
+    """
+    pass
+
+  def teardown_suite(self):
+    """Function used to add post tests cleanup tasks (optional)."""
+    pass
```
xpconanfan, line 66: missing new line?

mobly/base_suite.py
```diff
@@ -0,0 +1,66 @@
+# Copyright 2017 Google Inc.
```
xpconanfan, line 1: why is this 2017? isn't this a new file in github?

mobly/suite_runner.py
```diff
@@ -13,8 +13,10 @@
 # limitations under the License.
 """Runner for Mobly test suites.
```
xpconanfan, line 14: I think we need to further clarify the purposes of this suite_runner and the base_suite
Both of the are meant to be boilerplate code or examples for users to construct their own suite code right?

What we provide works, but users should not feel they have to do things this way.
E.g. the argument format can be whatever users want to define.

mobly/base_suite.py
```diff
@@ -0,0 +1,66 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import abc
+
+class BaseSuite(abc.ABC):
```
xpconanfan, line 17: this is adding a new layer of abstraction between the test classes and the suite runner right?

We should promote the generated suite way as the default option to reduce boilerplate code for suites. So this class should only be used when configs need to be modified between classes in some way.

All these context should be clearly stated (maybe even repeatedly) to avoid confusion.

Do we need to introduce the generated suite feature in the suite runner?

xpconanfan (changes requested):

booneng (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/817)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @xpconanfan)

___
*[`mobly/base_suite.py` line 1 at r1](https://reviewable.io/reviews/google/mobly/817#-N1sZomVBDqwRT7VWDm3-r1-1:-N1vW2Fw4HaZoS9cUhrd:b-896fix) ([raw file](https://github.com/google/mobly/blob/8547041f866ecc8d27a54ddd27f5210785b311ef/mobly/base_suite.py#L1)):*

why is this 2017? isn't this a new file in github?
</blockquote></details>

Done.
___
*[`mobly/base_suite.py` line 17 at r1](https://reviewable.io/reviews/google/mobly/817#-N1sZomVBDqwRT7VWDm3-r1-17:-N1vYOvZ9Bd7VkeFBspq:b-63z9b1) ([raw file](https://github.com/google/mobly/blob/8547041f866ecc8d27a54ddd27f5210785b311ef/mobly/base_suite.py#L17)):*

this is adding a new layer of abstraction between the test classes and the suite runner right?

We should promote the generated suite way as the default option to reduce boilerplate code for suites. So this class should only be used when configs need to be modified between classes in some way.

All these context should be clearly stated (maybe even repeatedly) to avoid confusion.

Do we need to introduce the generated suite feature in the suite runner?

</blockquote></details>

The main purpose of using the BaseSuite class is for users who want to define setup and teardown steps on the suite level. Added this comment to docstring.

Hmm I think suite_runner.run_suite seems enough. for the internal generated suite feature, it is to allow users to define which test classes in the build rule. For external version, I guess this would be specifying the test classes in the config file instead. We could introduce it in the future but I think we should work on consolidating the suite libraries first.
___
*[`mobly/base_suite.py` line 66 at r1](https://reviewable.io/reviews/google/mobly/817#-N1sZomVBDqwRT7VWDm3-r1-66:-N1vW4PV13oyixf0FfhG:b-896fix) ([raw file](https://github.com/google/mobly/blob/8547041f866ecc8d27a54ddd27f5210785b311ef/mobly/base_suite.py#L66)):*

missing new line?
</blockquote></details>

Done.
___
*[`mobly/suite_runner.py` line 14 at r1](https://reviewable.io/reviews/google/mobly/817#-N1sZomVBDqwRT7VWDm4-r1-14:-N1vYDl49Rg_mAZom2G4:bxqmkng) ([raw file](https://github.com/google/mobly/blob/8547041f866ecc8d27a54ddd27f5210785b311ef/mobly/suite_runner.py#L14)):*

I think we need to further clarify the purposes of this suite_runner and the base_suite
Both of the are meant to be boilerplate code or examples for users to construct their own suite code right?

What we provide works, but users should not feel they have to do things this way.
E.g. the argument format can be whatever users want to define.
</blockquote></details>

Added comment to the docstring to clarify.

xpconanfan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/817)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @booneng and @xpconanfan)

___
*[`mobly/suite_runner.py` line 16 at r3](https://reviewable.io/reviews/google/mobly/817#-N1vrsB69UCTj6I3R5QL:-N1vrsB7Ezdu_6G2V-75:b3mohlb) ([raw file](https://github.com/google/mobly/blob/c05d8214058262fc096aada750984b4aec557f5c/mobly/suite_runner.py#L16)):*
> ```Python
> """Runner for Mobly test suites.
> 
> These are just utility functions to help users run a collection of Mobly test
> ```

"just utility functions" -> "example code" ?
can then say that users can either use them as is or customize per their own requirements
___
*[`mobly/suite_runner.py` line 34 at r3](https://reviewable.io/reviews/google/mobly/817#-N1vsABSDl05hJUuwSou:-N1vsABSDl05hJUuwSov:by5v4vu) ([raw file](https://github.com/google/mobly/blob/c05d8214058262fc096aada750984b4aec557f5c/mobly/suite_runner.py#L34)):*
> ```Python
> 
> 2. Create a subclass of base_suite.BaseSuite and add the individual test
> classes. Using the BaseSuite class allows users to define their own
> ```

also allows customize configs in code

booneng (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/817)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @booneng and @xpconanfan)

___
*[`mobly/suite_runner.py` line 16 at r3](https://reviewable.io/reviews/google/mobly/817#-N1vrsB69UCTj6I3R5QL:-N1wEpzxDFDgVf-gCoaB:b-896fix) ([raw file](https://github.com/google/mobly/blob/c05d8214058262fc096aada750984b4aec557f5c/mobly/suite_runner.py#L16)):*

"just utility functions" -> "example code" ?
can then say that users can either use them as is or customize per their own requirements
</blockquote></details>

Done.
___
*[`mobly/suite_runner.py` line 34 at r3](https://reviewable.io/reviews/google/mobly/817#-N1vsABSDl05hJUuwSou:-N1wEqy1EQt5Vh_OAcTI:b-896fix) ([raw file](https://github.com/google/mobly/blob/c05d8214058262fc096aada750984b4aec557f5c/mobly/suite_runner.py#L34)):*

also allows customize configs in code
</blockquote></details>

Done.

mobly/suite_runner.py
```diff
@@ -70,7 +107,89 @@ def run_suite(test_classes, argv=None):
...
+        test_suites.append(module_member)
+  if len(test_suites) != 1:
+    logging.error('Expected 1 test class per file, found %s.',
+                  [t.__name__ for t in test_suites])
+    sys.exit(1)
+  return test_suites[0]
+
+
+class _SuiteRunner(object):
+  """Class used to execute a Mobly suite in google3."""
```
xpconanfan, line 136: should not refer to google3 in any way

mobly/suite_runner.py
```diff
@@ -70,7 +107,89 @@ def run_suite(test_classes, argv=None):
...
+      if issubclass(module_member, base_suite.BaseSuite):
+        test_suites.append(module_member)
+  if len(test_suites) != 1:
+    logging.error('Expected 1 test class per file, found %s.',
+                  [t.__name__ for t in test_suites])
+    sys.exit(1)
+  return test_suites[0]
+
+
+class _SuiteRunner(object):
```
xpconanfan, line 135: Why is this named in the private way?

mobly/suite_runner.py
```diff
@@ -70,7 +107,89 @@ def run_suite(test_classes, argv=None):
...
+          print(ok)
+        except signals.TestAbortAll:
+          pass
+      finally:
+        self._suite.teardown_suite()
+    if not ok:
+      sys.exit(1)
+
+
+def run_suite_class(argv=None):
```
xpconanfan, line 170: let's define an order for members of a module?
usually functions come before classes, except error types

mobly/suite_runner.py
xpconanfan, line 136: I guess this is more like a reference implementation of what a suite runner could be like?

xpconanfan (changes requested):

booneng (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/817)* status: 0 of 3 files reviewed, 6 unresolved discussions (waiting on @booneng and @xpconanfan)

___
*[`mobly/suite_runner.py` line 135 at r4](https://reviewable.io/reviews/google/mobly/817#-N1sZomVBDqwRT7VWDm4-r4-135:-N2LVYhzD23zFIDozx1Q:bss4a2b) ([raw file](https://github.com/google/mobly/blob/bd5d7b4b77cb6a25480461528f2a178c834c2bce/mobly/suite_runner.py#L135)):*

Why is this named in the private way?
</blockquote></details>

Removed.
___
*[`mobly/suite_runner.py` line 136 at r4](https://reviewable.io/reviews/google/mobly/817#-N1sZomVBDqwRT7VWDm4-r4-136:-N2LVb_B41cbWAaq-qYE:b1b3alj) ([raw file](https://github.com/google/mobly/blob/bd5d7b4b77cb6a25480461528f2a178c834c2bce/mobly/suite_runner.py#L136)):*

I guess this is more like a reference implementation of what a suite runner could be like?
</blockquote></details>

Removed this class. I realised it is not really necessary. Not sure why it was there in the first place. All the logic can be moved to the run_suite_class function instead.
___
*[`mobly/suite_runner.py` line 170 at r4](https://reviewable.io/reviews/google/mobly/817#-N1sZomVBDqwRT7VWDm4-r4-170:-N2LO3EW2Rpyx8Fz3TuJ:bsscbr8) ([raw file](https://github.com/google/mobly/blob/bd5d7b4b77cb6a25480461528f2a178c834c2bce/mobly/suite_runner.py#L170)):*

let's define an order for members of a module?
usually functions come before classes, except error types
</blockquote></details>

Done. Moved to bottom.

xpconanfan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/mobly/817#-:-N2UgYYLDGpzGZ9SzrtW:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @booneng)

xpconanfan (approved):

PR #822, opened 2022-05-18
Event client clears host_port and device_port attribute

mhaoli:
Without cleaning host_port and device_port attribute of event client, there will be a caught exception during `teardown_class` process. Because the event client and snippet client will clean up the port forwarding configuration for the same port. 

This problem occurs whenever the user uses an async RPC. But this exception will be caught, so it won't cause the test to fail.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/822)

### DIFF
--- a/mobly/controllers/android_device_lib/snippet_client.py
+++ b/mobly/controllers/android_device_lib/snippet_client.py
@@ -262,6 +262,11 @@ class SnippetClient(jsonrpc_client_base.JsonRpcClientBase):
     """Releases all the resources acquired in `_start_event_client`."""
     if self._event_client:
       self._event_client.close_socket_connection()
+      # Without cleaning host_port of event_client, event client will try to
+      # stop the port forwarding when deconstructed, which should only be
+      # stopped by snippet client.
+      self._event_client.host_port = None
+      self._event_client.device_port = None
       self._event_client = None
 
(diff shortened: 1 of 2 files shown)


mobly/controllers/android_device_lib/snippet_client.py
```diff
@@ -265,6 +265,11 @@ def _stop_event_client(self):
     """Releases all the resources acquired in `_start_event_client`."""
     if self._event_client:
       self._event_client.close_socket_connection()
+      # Without cleaning host_port of event_client, the event client will try to
+      # stop the port forwarding when deconstructed, which should only be
+      # stopped by the corresponding snippet client.
+      self._event_client.host_port = None
```
xpconanfan, line 271: should we release the port here before setting the var to None?

xpconanfan (changes requested):

mobly/controllers/android_device_lib/snippet_client.py
mhaoli: The port forwarding should only be released by the corresponding snippet client. This is because that in `_start_event_client`, the event client directly uses the port of the snippet client without doing port forwarding.

xpconanfan (approved):

PR #823, opened 2022-05-18
Remove the dependency of client v2's unit test on subprocess

mhaoli:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/823)


mhaoli:
The failed kokoro-windows test on this page doesn't seem right. As it failed to pull code from github.

PR #825, opened 2022-05-23
Client V2 cleans event client when it stops

mhaoli:
Part of #793.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/825)

### DIFF
--- a/mobly/controllers/android_device_lib/snippet_client_v2.py
+++ b/mobly/controllers/android_device_lib/snippet_client_v2.py
@@ -621,6 +621,9 @@ class SnippetClientV2(client_base.ClientBase):
   def _destroy_event_client(self):
     """Releases all the resources acquired in `_create_event_client`."""
     if self._event_client:
+      # Without cleaning host_port of event_client first, the close_connection
+      # will try to stop the port forwarding, which should only be stopped by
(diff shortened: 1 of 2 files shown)


mobly/controllers/android_device_lib/snippet_client_v2.py
```diff
@@ -616,6 +618,14 @@ def _stop_server(self):
           self._device,
           f'Failed to stop existing apk. Unexpected output: {out}.')
 
+  def _destroy_event_client(self):
+    """Releases all the resources acquired in `_create_event_client`."""
+    if self._event_client:
+      self._event_client.host_port = None
+      self._event_client.device_port = None
+      self._event_client.close_connection()
```
xpconanfan, line 626: does the order of these statements matter?
Why not close the connect before setting everything to None?

xpconanfan (changes requested):

mobly/controllers/android_device_lib/snippet_client_v2.py
mhaoli: Yes, it matters. Added comments here for the reason.

xpconanfan (approved):

PR #826, opened 2022-05-25
Standardize Async RPC for snippet client v2

mhaoli:
A followup of #793.

Changes include:
* Splits the original callback handler into base class and GeneralCallbackHandler
* Snippet client v2 uses GeneralCallbackHandler.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/826)

### DIFF
--- a/mobly/controllers/android_device_lib/callback_handler_v2.py
+++ b/mobly/controllers/android_device_lib/callback_handler_v2.py
@@ -11,39 +11,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""A general handler class for Mobly Snippet Lib's callback events.
+"""The callback handler V2 module for Android Mobly Snippet Lib."""
 
-The protocol for pulling events expected by this module is:
-
-RPC 1: eventWaitAndGet: Waits and returns an event for the specified identifier
-                        with timeout.
-JSON RPC Request:
-{
-  'id': <RPC-id>,
-  'method': 'eventWaitAndGet',
-  'params': (<callback_id: string>, <event_name: string>, <timeout_ms: int>),
-}
-
-RPC 2: eventGetAll: Gets all existing events for the specified identifier
-                    without waiting.
-JSON RPC Request:
-{
-  'id': <RPC-id>,
-  'method': 'eventGetAll',
-  'params': (<callback_id: string>, <event_name: string>),
-}
-
-Timeout error message:
-In the response of eventWaitAndGet, this class checks for the existence of
-the substring "EventSnippetException: timeout." in the error message. If it
-exists, this class throws a timeout error from the original error.
-"""
 from mobly.snippet import callback_handler_base
 from mobly.snippet import errors
 
 
-class GeneralCallbackHandler(callback_handler_base.CallbackHandlerBase):
-  """A general handler class for Mobly Snippet Lib's callback events."""
+class CallbackHandlerV2(callback_handler_base.CallbackHandlerBase):
+  """The callback handler V2 class for Android Mobly Snippet Lib."""
 
   def __init__(self,
                callback_id,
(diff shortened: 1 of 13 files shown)


mobly/snippet/general_callback_handler.py
```diff
@@ -0,0 +1,115 @@
...
+Timeout error message:
+In the response of eventWaitAndGet, this class checks for the existence of
+the substring "EventSnippetException: timeout." in the error message. If it
+exists, this class throws a timeout error from the original error.
+"""
+from mobly.snippet import callback_handler_base
+from mobly.snippet import errors
+
+
+class GeneralCallbackHandler(callback_handler_base.CallbackHandlerBase):
```
xpconanfan, line 45: What makes this "general"?
This is general as opposed to what?
And why do we have both general handler and handler base...

This is quite confusing

xpconanfan (changes requested):

mobly/snippet/general_callback_handler.py
mhaoli: The Android and Windows clients use the general callback handler and the iOS client will use IosCallbackHandler. 

Use "General" here because we recommend future implementations of other clients to use this client when possible (this has restrictions on server-side RPC function names).

And the handler base is an abstract class which can not be used directly.

Does this make sense?

mhaoli: The iOS one is special because it has limitation on the function name of RPCs, which are for pulling events from server.

Through writing in this way, I'm trying to make this point clearer: if you need to implement a new callback handler, which methods should you override. These methods are marked as abstractmethod and you cannot instantiate you class without overriding them. This seems clearer than reminding the user via docstring which functions should be overridden.

So it seems that I'm making things complex and confusing?

xpconanfan:
This is more complicated than I expected...

Different platforms and programming languages will inherently have
different conventions for their snippet rpc impls. So we just need to
provide a simple abstract class for each client to implement the
corresponding rpc calls, which is just 2 or 3 methods.

mhaoli:
I think this PR provides the abstract class and one general handler class with 2 methods.

Is the reason it looks complicated because of the protocol it summarizes in the docstring?

Because the current implementations of Android and Windows callback handler are exactly the same, I think maybe other platforms can also implement this in future. Thus I summarize their protocol for interacting with the server side. Platforms that meet this protocol can reuse this class. Platforms that do not meet this protocol can implement callback handlers by themselves.

Or should I remove these descriptions of the protocol and the "general" word from the class name.

mobly/controllers/android_device_lib/callback_handler_v2.py
```diff
@@ -0,0 +1,90 @@
...
+"""The callback handler V2 module for Android Mobly Snippet Lib."""
+
+from mobly.snippet import callback_handler_base
+from mobly.snippet import errors
+
+
+class CallbackHandlerV2(callback_handler_base.CallbackHandlerBase):
+  """The callback handler V2 class for Android Mobly Snippet Lib."""
+
+  def __init__(self,
```
xpconanfan, line 23: can the constructor be inherited?

xpconanfan (changes requested):

mobly/controllers/android_device_lib/callback_handler_v2.py
```diff
@@ -0,0 +1,64 @@
...
+    Raises:
+      errors.CallbackHandlerTimeoutError: The expected event does not occur
+        within the time limit.
+    """
+    timeout_ms = int(timeout_sec * 1000)
+    try:
+      return self._event_client.eventWaitAndGet(callback_id, event_name,
+                                                timeout_ms)
+    except Exception as e:
+      if 'EventSnippetException: timeout.' in str(e):
```
xpconanfan, line 45: make this string a const?

also if all the server impl will have its own timeout msg, should this be a property for each child handler to override

xpconanfan (changes requested):

xpconanfan (approved):

mobly/controllers/android_device_lib/callback_handler_v2.py
mhaoli: Made it a constant.

I think we may not add it to a property for base class. As Subclasses currently do not have the constructor to override a property. So a constant string seems a good choice to make subclasses keep simple.

xpconanfan (approved):

PR #832, opened 2022-06-03
Alias error classes of callback handler v1 to the new ones

mhaoli:
Part of #793.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/832)

### DIFF
--- a/mobly/controllers/android_device_lib/callback_handler.py
+++ b/mobly/controllers/android_device_lib/callback_handler.py
@@ -14,8 +14,8 @@
 
 import time
 
-from mobly.controllers.android_device_lib import errors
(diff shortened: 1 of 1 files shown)


xpconanfan (approved):

mobly/controllers/android_device_lib/callback_handler.py
```diff
@@ -14,22 +14,18 @@
 
 import time
```
xpconanfan, line 16: should we also mark this and the other v1 classes "deprecated" in docstring?

mhaoli: Should we mark them as deprecated after setting v2 as default?

PR #834, opened 2022-06-08
ClientBase shouldn't call "stop" if "before_starting_server" failed

mhaoli:
Part of #793.

Reason for change:
* Keep the same logic as Android snippet client v1.
* Previously I thought it was ok to call "stop" to make sure everything is cleaned. But in fact this causes different behavior from client v1, which is wrong. e.g., when "before_starting_server" fails without installing the snippet apk, "stop" will fail and raise a new error.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/834)

### DIFF
--- a/mobly/snippet/client_base.py
+++ b/mobly/snippet/client_base.py
@@ -154,6 +154,10 @@ class ClientBase(abc.ABC):
     For example, subclass can check or modify the device settings at this
     stage.
 
+    NOTE: Do not acquire resources in this function, as this should only contain
+    preparation steps and any error at this stage will abort the initialization
+    without cleanup.
+
     Raises:
       errors.ServerStartPreCheckError: when prechecks for starting the server
         failed.
--- a/mobly/snippet/client_base.py
+++ b/mobly/snippet/client_base.py
@@ -97,13 +97,13 @@ class ClientBase(abc.ABC):
     """Initializes the snippet client to interact with the remote device.
 
     This function contains following stages:
-      1. before_starting_server: preparing to start the snippet server.
-      2. start_server: starting the snippet server on the remote device.
-      3. make_connection: making a connection to the snippet server.
+      1. before starting server: preparing to start the snippet server.
+      2. start server: starting the snippet server on the remote device.
+      3. make connection: making a connection to the snippet server.
 
     An error occurring at any stage will abort the initialization. Only errors
-    at the start_server and make_connection stages will trigger `stop` to clean
-    up.
+    at the `start_server` and `make_connection` stages will trigger `stop` to
+    clean up.
 
     Raises:
       errors.ProtocolError: something went wrong when exchanging data with the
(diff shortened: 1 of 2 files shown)


mobly/snippet/client_base.py
```diff
@@ -119,11 +119,10 @@ def initialize(self):
     self.log.info('Initializing the snippet package %s.', self.package)
     start_time = time.perf_counter()
 
-    try:
-      self.log.debug('Preparing to start the snippet server of %s.',
-                     self.package)
-      self.before_starting_server()
+    self.log.debug('Preparing to start the snippet server of %s.', self.package)
+    self.before_starting_server()
```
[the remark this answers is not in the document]
mhaoli: I think we'd better not to acquire resources in this function. Or this function should release the acquired resources if an error occurs. 

Added this to the docstring.

xpconanfan (approved):

mobly/snippet/client_base.py
```diff
@@ -101,8 +101,8 @@ def initialize(self):
       2. starting the snippet server on the remote device.
       3. making a connection to the snippet server.
 
-    If error occurs at any stage, this function will abort the initialization
-    process and call `stop` to clean up.
+    An error occurs at any stage will abort the initialization. Only errors in
```
xpconanfan, line 104: occurs -> occurring

mobly/snippet/client_base.py
```diff
@@ -101,8 +101,8 @@ def initialize(self):
...
+    the second and third stages will trigger `stop` to clean up.
```
xpconanfan, line 105: be clear about the stage names.
"second and third" is not specific enough

mobly/snippet/client_base.py
mhaoli: Done.

mobly/snippet/client_base.py
mhaoli: Done.

mobly/snippet/client_base.py
```diff
@@ -97,12 +97,13 @@ def initialize(self):
...
-      2. starting the snippet server on the remote device.
-      3. making a connection to the snippet server.
+      1. before_starting_server: preparing to start the snippet server.
+      2. start_server: starting the snippet server on the remote device.
+      3. make_connection: making a connection to the snippet server.
 
-    If error occurs at any stage, this function will abort the initialization
-    process and call `stop` to clean up.
+    An error occurring at any stage will abort the initialization. Only errors
+    at the start_server and make_connection stages will trigger `stop` to clean
```
xpconanfan, line 105: add "``" for function names

xpconanfan (changes requested):

mobly/snippet/client_base.py
mhaoli: Done.

xpconanfan (approved):

PR #839, opened 2022-06-29
Flip the default Android snippet client version to V2

mhaoli:
Part of #793.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/839)


tests/mobly/controllers/android_device_lib/services/snippet_management_service_test.py
```diff
@@ -27,137 +26,137 @@ class SnippetManagementServiceTest(unittest.TestCase):
 
   def test_empty_manager_start_stop(self):
     manager = snippet_management_service.SnippetManagementService(
-        mock.MagicMock())
+        mock.MagicMock(spec=['log']))
```
xpconanfan, line 29: why adding the `spec` now?

xpconanfan (changes requested):

tests/mobly/controllers/android_device_lib/services/snippet_management_service_test.py
mhaoli: Without `spec`, accessing any attribute of a mock object via `getattr` will return a mock object. This will cause the modified `_is_using_client_v2` method to always return False.

xpconanfan: This `spec` arg has no obvious meaning to readers though.
Can we not more explicitly set the value?

This mock is for the configs right?
Maybe you can set the config value so `_is_using_client_v2` is now always `True` for these unit tests?
It's better to set specific attr value to the mock than relying on implicit behaviors.

xpconanfan (changes requested):

tests/mobly/controllers/android_device_lib/services/snippet_management_service_test.py
mhaoli: Set the switch to True for most tests.

But still set `spec` in the tests that are for client v2 switch, as calculating the switch value is part of the test. I think it's okay to set `spec` in them because they will be deleted after migration.

xpconanfan (approved):

PR #852, opened 2022-10-27
Expose the method to set new log prefix in PrefixLoggerAdapter

mhaoli:
Why we need this? We need to modify the log prefix when changing the debug tag of device controllers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/852)

### DIFF
--- a/mobly/logger.py
+++ b/mobly/logger.py
@@ -427,4 +427,5 @@ class PrefixLoggerAdapter(logging.LoggerAdapter):
     Args:
       prefix: The new log prefix.
     """
+    self.debug('Setting the log prefix to "%s".', prefix)
     self.extra[PrefixLoggerAdapter.EXTRA_KEY_LOG_PREFIX] = prefix
--- a/mobly/logger.py
+++ b/mobly/logger.py
@@ -412,11 +412,19 @@ class PrefixLoggerAdapter(logging.LoggerAdapter):
     """Processes the logging call to insert contextual information.
 
     Args:
-      msg: the logging message
-      kwargs: keyword arguments passed in to a logging call
+      msg: The logging message.
+      kwargs: Keyword arguments passed in to a logging call.
 
     Returns:
-      the message and kwargs modified.
+      The message and kwargs modified.
     """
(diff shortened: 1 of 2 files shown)


xpconanfan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/852)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @mhaoli)

___
*[`mobly/logger.py` line 424 at r1](https://reviewable.io/reviews/google/mobly/852#-NFTzYiX0Herq2MiDuCR:-NFTzYiX0Herq2MiDuCS:b-b9h1vw) ([raw file](https://github.com/google/mobly/blob/5684b05879f2b38eb6dc4ee164c12781263142f3/mobly/logger.py#L424)):*
> ```Python
>     return (new_msg, kwargs)
> 
>   def set_log_prefix(self, prefix: str) -> None:
> ```

This implies users can call this method and change the prefix at any time, like in the middle of the test?
Just trying to clarify the behavior :)

mhaoli (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/852)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @mhaoli and @xpconanfan)

___
*[`mobly/logger.py` line 424 at r1](https://reviewable.io/reviews/google/mobly/852#-NFTzYiX0Herq2MiDuCR:-NFgDlUH5OcIGIUQdALJ:b-b80o8c) ([raw file](https://github.com/google/mobly/blob/5684b05879f2b38eb6dc4ee164c12781263142f3/mobly/logger.py#L424)):*

This implies users can call this method and change the prefix at any time, like in the middle of the test?
Just trying to clarify the behavior :)
</blockquote></details>

Yes, users can change the prefix at any time. 

`PrefixLoggerAdapter` doesn't prevent modification. If we need to prevent modification, it might make more sense to do it in the objects that hold logger adapters, like device controllers.

xpconanfan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/mobly/852#-:-NFvosHi1nTYDtJ9WkKp:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @mhaoli)

PR #856, opened 2022-12-05
Don't mutate BaseTestClass.record_data() argument.

wolverdude:
Mutating the argument can lead to weird results when mocking out the method for testing user code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/856)

### DIFF
--- a/tests/mobly/base_test_test.py
+++ b/tests/mobly/base_test_test.py
@@ -2198,6 +2198,7 @@ class BaseTestTest(unittest.TestCase):
 
   def test_write_user_data(self):
     content = {'a': 1}
+    original_content = content.copy()
 
     class MockBaseTest(base_test.BaseTestClass):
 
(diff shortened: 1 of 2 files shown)


xpconanfan (approved):

xpconanfan (changes requested):
Actually, can you add a unit test to validate this behavior?
Should be straightforward :)

tests/mobly/base_test_test.py
```diff
@@ -2215,7 +2216,9 @@ def test_something(self):
           continue
         hit = True
         self.assertEqual(c['a'], content['a'])
+        self.assertIn('timestamp', c)
         self.assertIsNotNone(c['timestamp'])
```
wolverdude, line 2220: This line would fail with an IndexError instead of an AssertionError if timestamp were missing, which is why I added the line above.

xpconanfan (approved):

PR #858, opened 2022-12-08
Handle SIGTERM in Mobly.

booneng:
Handle SIGTERM in Mobly to allow Mobly to complete writing out summary report.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/858)

### DIFF
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -76,13 +76,12 @@ def main(argv=None):
       runner.add_test_class(config, test_class, tests)
       try:
         runner.run()
+        ok = runner.results.is_all_pass and ok
       except signals.TestAbortAll:
         pass
       except Exception:
         logging.exception('Exception when executing %s.', config.testbed_name)
         ok = False
-      finally:
-        ok = runner.results.is_all_pass and ok
   if not ok:
     sys.exit(1)
 
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -406,9 +406,10 @@ class TestRunner:
 
     # When a SIGTERM is received during the execution of a test, the Mobly test
     # immediately terminates without executing any of the finally blocks. This
-    # signal handler handles this case by raising an Exception instead, so the
-    # SIGTERM is essentially "converted" to an Exception, which allows the
-    # finally blocks to be executed.
+    # handler converts the SIGTERM into a TestAbortAll signal so that the
+    # finally blocks will execute. We use TestAbortAll because other exceptions
+    # will be caught in the base test class and it will continue executing
+    # remaining tests.
     def sigterm_handler(*args):
       logging.warning('Test received a SIGTERM. Aborting all tests.')
       raise signals.TestAbortAll('Test received a SIGTERM.')
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -16,6 +16,7 @@ import argparse
 import contextlib
 import logging
 import os
+import signal
 import sys
 import time
 
@@ -66,6 +67,18 @@ def main(argv=None):
   if args.tests:
     tests = args.tests
   console_level = logging.DEBUG if args.verbose else logging.INFO
+
+  # When a SIGTERM is received during the execution of a test, the Mobly test
+  # immediately terminates without executing any of the finally blocks. This
+  # signal handler handles this case by raising an Exception instead, so the
+  # SIGTERM is essentially "converted" to an Exception, which allows the
+  # finally blocks to be executed.
+  def sigterm_handler(*args):
+    raise signals.TestAbortAll(
+        'Test was terminated. This could be due to a timeout.')
+
+  signal.signal(signal.SIGTERM, sigterm_handler)
+
   # Execute the test class with configs.
   ok = True
   for config in test_configs:
@@ -75,12 +88,13 @@ def main(argv=None):
       runner.add_test_class(config, test_class, tests)
       try:
         runner.run()
-        ok = runner.results.is_all_pass and ok
       except signals.TestAbortAll:
         pass
       except Exception:
         logging.exception('Exception when executing %s.', config.testbed_name)
         ok = False
+      finally:
+        ok = runner.results.is_all_pass and ok
   if not ok:
     sys.exit(1)
 
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -67,7 +67,6 @@ def main(argv=None):
   if args.tests:
     tests = args.tests
   console_level = logging.DEBUG if args.verbose else logging.INFO
-
   # Execute the test class with configs.
   ok = True
   for config in test_configs:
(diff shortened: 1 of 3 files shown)


mhaoli (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/858)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @booneng and @xpconanfan)

___
*[`mobly/test_runner.py` line 78 at r1](https://reviewable.io/reviews/google/mobly/858#-NIkt0Sd3eYP2CsZ2Ca7:-NIkt0Sd3eYP2CsZ2Ca8:b-4ohlk2) ([raw file](https://github.com/google/mobly/blob/759f2c62b1c53835eeb5f0f19f3bd6bfbe26b093/mobly/test_runner.py#L78)):*
> ```Python
>   def sigterm_handler(*args):
>     raise signals.TestAbortAll(
>         'Test was terminated. This could be due to a timeout.')
> ```

As this is added to the public Mobly repo, so it may receive SIGTERM for many reasons. Not sure if it's better that we don't guess the reason, but just say we get a SIGTERM signal? 



_Code quote:_
```Python
This could be due to a timeout.
```

booneng (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/858)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mhaoli and @xpconanfan)

___
*[`mobly/test_runner.py` line 78 at r1](https://reviewable.io/reviews/google/mobly/858#-NIkt0Sd3eYP2CsZ2Ca7:-NIkyNvV6ZDSTC5DGH7I:b-896fix) ([raw file](https://github.com/google/mobly/blob/759f2c62b1c53835eeb5f0f19f3bd6bfbe26b093/mobly/test_runner.py#L78)):*

As this is added to the public Mobly repo, so it may receive SIGTERM for many reasons. Not sure if it's better that we don't guess the reason, but just say we get a SIGTERM signal? 
</blockquote></details>

Done.

mhaoli (approved):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/858)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @booneng and @xpconanfan)

___
*[`mobly/test_runner.py` line 414 at r2](https://reviewable.io/reviews/google/mobly/858#-NIl-AsJ7v6Akt6_W55U:-NIl-AsJ7v6Akt6_W55V:b-h9rfuc) ([raw file](https://github.com/google/mobly/blob/60512bdd00b7e1a6fa09845aabcf525649e1f8d0/mobly/test_runner.py#L414)):*
> ```Python
>     # finally blocks to be executed.
>     def sigterm_handler(*args):
>       raise signals.TestAbortAll('Test received a SIGTERM.')
> ```

Nit: seems that we can print a log before raising, we can tell users we get the SIGTERM but we will do cleanup before exit. This might be helpful bc users may expect we exit right now, but we may need some time.



_Code quote:_
```Python
raise signals.TestAbortAll('Test received a SIGTERM.')
```

mobly/test_runner.py
```diff
@@ -75,12 +76,13 @@ def main(argv=None):
       runner.add_test_class(config, test_class, tests)
       try:
         runner.run()
-        ok = runner.results.is_all_pass and ok
```
xpconanfan, line 78: why remove this one?

xpconanfan (changes requested):

booneng (commented):
If we have a scenario where the first test fails and the next test receives TestAbortAll, then the overall result will still be passed because the TestAbortAll is ignored in this block. So I moved this line to finally, so the correct result will be set when TestAbortAll is received.

> *[Reviewable](https://reviewable.io/reviews/google/mobly/858#-:-NIp5ap59ezmgzRYab1M:b-rqyioe)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @booneng and @mhaoli)

booneng (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/858)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @booneng and @mhaoli)

___
*[`mobly/test_runner.py` line 414 at r2](https://reviewable.io/reviews/google/mobly/858#-NIl-AsJ7v6Akt6_W55U:-NIp7Qt4751pKLBlq4Xg:b-896fix) ([raw file](https://github.com/google/mobly/blob/60512bdd00b7e1a6fa09845aabcf525649e1f8d0/mobly/test_runner.py#L414)):*

Nit: seems that we can print a log before raising, we can tell users we get the SIGTERM but we will do cleanup before exit. This might be helpful bc users may expect we exit right now, but we may need some time.
</blockquote></details>

Done.

mobly/test_runner.py
xpconanfan: Please reply comments in their original thread.

It sounds like in that case the test should pass, so it's expected behavior, no?

In any case, that's a different change; should be a separate PR if needed

booneng (commented):
Ahh okay if that is the intended behaviour then that is fine.

I reverted this change.

> *[Reviewable](https://reviewable.io/reviews/google/mobly/858#-:-NIpH9be0Mw64g_rM0Yu:b-gl6vmr)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @booneng and @mhaoli)

mobly/test_runner.py
booneng: Reverted.

xpconanfan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/858)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @booneng and @mhaoli)

___
*[`mobly/test_runner.py` line 407 at r5](https://reviewable.io/reviews/google/mobly/858#-NJ3wspm3FL4ZZVNAWla:-NJ3wspm3FL4ZZVNAWlb:bmxta1g) ([raw file](https://github.com/google/mobly/blob/7e7cfdc6a5706b65a620c9bcc44a70b415737645/mobly/test_runner.py#L407)):*
> ```Python
>                      records.OUTPUT_FILE_SUMMARY))
> 
>     # When a SIGTERM is received during the execution of a test, the Mobly test
> ```

clarify the reasoning for the location of this and the signal being used

booneng (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/858)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @booneng, @mhaoli, and @xpconanfan)

___
*[`mobly/test_runner.py` line 407 at r5](https://reviewable.io/reviews/google/mobly/858#-NJ3wspm3FL4ZZVNAWla:-NJ4DtrD7PbxnfwOsVi9:b-896fix) ([raw file](https://github.com/google/mobly/blob/7e7cfdc6a5706b65a620c9bcc44a70b415737645/mobly/test_runner.py#L407)):*

clarify the reasoning for the location of this and the signal being used
</blockquote></details>

Done.

xpconanfan (approved):

PR #859, opened 2022-12-27
Fix list_fastboot_devices for other usb interfaces

c910335:
According to [fastboot/usb_linux.cpp](https://android.googlesource.com/platform/system/core/+/refs/heads/master/fastboot/usb_linux.cpp#272) and [fastboot/fastboot.cpp](https://android.googlesource.com/platform/system/core/+/refs/heads/master/fastboot/fastboot.cpp#290), the USB interface (the right part of `fastboot devices`) of a device is not always `fastboot`.
I have seen `Android Fastboot` and `fastbootd`, but I am not sure whether there are other cases.
Maybe we could just not check the interface?

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/859)


xpconanfan:
Thanks for the contribution!
This seems like a reasonable change. But it changes a public API so we need to put it in a major release instead of a minor.

The alternative of not checking sounds ok for the `fastboot devices` cmd, but we probably want to keep it for the `adb devices` cmd.

c910335:
I have updated it to not check only for `list_fastboot_devices`.

xpconanfan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/859)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @c910335)

___
*[`mobly/controllers/android_device.py` line 194 at r1](https://reviewable.io/reviews/google/mobly/859#-NKxzN1y9ELRCY8LAP0x:-NKxzN1y9ELRCY8LAP0y:b-vt32dy) ([raw file](https://github.com/google/mobly/blob/cdafc949d3423553fcdab670702a73f4dc08d4d1/mobly/controllers/android_device.py#L194)):*
> ```Python
>   Args:
>     device_list_str: Output of adb or fastboot.
>     key: The token that signifies a device in device_list_str.
> ```

can you update the docstring to call out the default behavior added for this arg?

c910335:
Done.

xpconanfan (approved):

PR #861, opened 2023-01-03
Add retry and repeat number to test result record

booneng:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/861)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -687,10 +687,15 @@ class BaseTestClass:
     if not should_retry(previous_record):
       return
 
+    previous_record.retry_number = 0
+
     for i in range(max_count - 1):
-      retry_name = f'{test_name}_retry_{i+1}'
+      retry_number = i + 1
+      retry_name = f'{test_name}_retry_{retry_number}'
       new_record = records.TestResultRecord(retry_name, self.TAG)
       new_record.retry_parent = previous_record
+      new_record.retry_number = retry_number
+
       previous_record = self.exec_one_test(retry_name, test_method, new_record)
       if not should_retry(previous_record):
         break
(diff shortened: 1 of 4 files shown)


xpconanfan:
I'm not sure why we need to add this now if everything worked without these before.

We need to be extra careful about adding more fields in the result record and making public API changes.

booneng (commented):
We have several users who generate tests and append digits to the test case name and have some monitoring setups that depend on it. In these cases, when parsing the test_summary.yaml, it is impossible to distinguish between user added digits or if it is a Mobly repeated case. This is blocking some of the new stuff.

> *[Reviewable](https://reviewable.io/reviews/google/mobly/861#-:-NKuZ9ke03e7W0Jn6fr_:b3kfuwy)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @xpconanfan)

booneng:
>

booneng:
> I'm not sure why we need to add this now if everything worked without these before.
> 
> We need to be extra careful about adding more fields in the result record and making public API changes.

We have several users who generate tests and append digits to the test case name and have some monitoring setups that depend on it. In these cases, when parsing the test_summary.yaml, it is impossible to distinguish between user added digits or if it is a Mobly repeated case. This is blocking some of the new stuff.

xpconanfan (commented):
The use case requirement sounds reasonable
but the solution is a little cumbersome.

Basically it's fair to express info specific to repeat/retry, and we already have `RECORD_RETRY_PARENT` which is in the same category.
But we can't keep adding new field for this area as we go. Let's think about how to handle this in a more efficient and proper design? I'm ok with deprecating `RECORD_RETRY_PARENT` if the new expression can capture all we need.

> *[Reviewable](https://reviewable.io/reviews/google/mobly/861#-:-NKxKI6-8e7tQWeK_m7v:b-cpip3m)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @booneng)

PR #865, opened 2023-01-14
Make the summary file path more accessible.

xpconanfan:
This file is the entry point of all result parsing. We make this path clear so the upper layer systems don't have to calculate it with assumptions that may change implicitly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/865)

### DIFF
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -252,6 +252,10 @@ class TestRunner:
                                            self._logger_start_time)
       return self.root_output_path
 
+    @property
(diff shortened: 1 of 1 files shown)


booneng (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/mobly/865)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @xpconanfan)

PR #872, opened 2023-02-28
record: emit explicit end when appending to summary 

uael:
While trying to handle test events in real-time, by reading additions the test summary file, I encountered incomplete `yaml` data.
The following change might fix that behavior.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/872)

### DIFF
--- a/mobly/records.py
+++ b/mobly/records.py
@@ -156,6 +156,7 @@ class TestSummaryWriter:
         yaml.safe_dump(new_content,
                        f,
                        explicit_start=True,
+                       explicit_end=True,
                        allow_unicode=True,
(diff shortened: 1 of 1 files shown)


xpconanfan:
Is it possible to create a "unit test" to demonstrate this bug and validate the fix at the same time?

xpconanfan (changes requested):
This PR is now very different from the first patch. Why?

uael:
I just updated the branch :+1:
I will open shortly a new PR with the other changes

xpconanfan (approved):

PR #873, opened 2023-03-02
record: add `TestClassList` to emitted records

uael:
When running multiple test classes in the test runner, it actually emit multiple `TestNameList` but one `Summary`. Which make impossible for a reader of the summary file to retrieve the complete list of tests before they had been executed.

To solve this a `TestClassList` record is now emitted before the run, which contains the full list of tests in the following format:
```yaml
Classes:
- Name: Foo
  Tests:
  - test_a
  - test_b
- Name: Bar
  Tests:
  - test_c
  - test_d
```

It was chosen to add a new record and to to not change the actual `TestNameList` record to not break the actual behavior.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/873)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -200,6 +200,7 @@ class BaseTestClass:
     self.user_params = configs.user_params
     self.results = records.TestResult()
     self.summary_writer = configs.summary_writer
+    self._pre_run_result = None
     self._generated_test_table = collections.OrderedDict()
     self._controller_manager = controller_manager.ControllerManager(
         class_name=self.TAG, controller_configs=configs.controller_configs)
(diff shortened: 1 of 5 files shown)


xpconanfan:
This is a more involved change, which would have worked better if a design discussion were held beforehand

mobly/base_test.py
```diff
@@ -348,6 +349,10 @@ def _pre_run(self):
     Returns:
       True if setup is successful, False otherwise.
     """
+    # Make sure `_pre_run` is executed only once, since it might
+    # be called by `get_test_names` before `run`.
```
xpconanfan, line 353: this is actually intentional

xpconanfan (changes requested):

uael:
> This is a more involved change, which would have worked better if a design discussion were held beforehand

Sounds good to me :+1:
I was not very sure how to proceed on this one, I will be happy to follow the right process.

uael:
Closing this one in favor of #875.

PR #875, opened 2023-03-15
suite-runner: try to mimic test-runner cli

uael:
As of the test runner:
* Parse only known arguments.
* Handle -l, --list_tests option.
* Handle -tb, --test_bed option.
* Handle -v, --verbose option.

Also, do not force test name to only contains one dot `.` by passing `maxsplit=1` to `test_name.split`. This allow for more complex generated test names.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/875)

### DIFF
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -92,22 +92,38 @@ def _parse_cli_args(argv):
     Namespace containing the parsed args.
   """
   parser = argparse.ArgumentParser(description='Mobly Suite Executable.')
-  parser.add_argument('-c',
-                      '--config',
+  group = parser.add_mutually_exclusive_group(required=True)
+  group.add_argument('-c',
+                     '--config',
+                     type=str,
+                     metavar='<PATH>',
+                     help='Path to the test configuration file.')
+  group.add_argument(
+      '-l',
+      '--list_tests',
+      action='store_true',
+      help='Print the names of the tests defined in a script without '
+      'executing them.')
+  parser.add_argument('--tests',
+                      '--test_case',
+                      nargs='+',
                       type=str,
-                      required=True,
-                      metavar='<PATH>',
-                      help='Path to the test configuration file.')
-  parser.add_argument(
-      '--tests',
-      '--test_case',
-      nargs='+',
-      type=str,
-      metavar='[ClassA[.test_a] ClassB[.test_b] ...]',
-      help='A list of test classes and optional tests to execute.')
+                      metavar='[ClassA[.test_a] ClassB[.test_b] ...]',
+                      help='A list of test classes and optional tests to execute.')
+  parser.add_argument('-tb',
+                      '--test_bed',
+                      nargs='+',
+                      type=str,
+                      metavar='[<TEST BED NAME1> <TEST BED NAME2> ...]',
+                      help='Specify which test beds to run tests on.')
+
+  parser.add_argument('-v',
+                      '--verbose',
+                      action='store_true',
+                      help='Set console logger level to DEBUG')
   if not argv:
     argv = sys.argv[1:]
-  return parser.parse_args(argv)
+  return parser.parse_known_args(argv)[0]
 
 
 def _find_suite_class():
@@ -132,6 +148,33 @@ def _find_suite_class():
   return test_suites[0]
 
 
+def _print_test_names(test_classes):
+  """Prints the names of all the tests in all test classes.
+  Args:
+    test_classes: classes, the test classes to print names from.
+  """
+  for test_class in test_classes:
+    cls = test_class(config_parser.TestRunConfig())
+    test_names = []
+    try:
+      # Executes pre-setup procedures, this is required since it might
+      # generate test methods that we want to return as well.
+      cls._pre_run()
+      if cls.tests:
+        # Specified by run list in class.
+        test_names = list(cls.tests)
+      else:
+        # No test method specified by user, list all in test class.
+        test_names = cls.get_existing_test_names()
+    except Exception:
+      logging.exception('Failed to retrieve generated tests.')
+    finally:
+      cls._clean_up()
+    print('==========> %s <==========' % cls.TAG)
+    for name in test_names:
+      print(f"{cls.TAG}.{name}")
+
+
 def run_suite_class(argv=None):
   """Executes tests in the test suite.
 
@@ -139,17 +182,22 @@ def run_suite_class(argv=None):
     argv: A list that is then parsed as CLI args. If None, defaults to sys.argv.
   """
   cli_args = _parse_cli_args(argv)
-  test_configs = config_parser.load_test_config_file(cli_args.config)
+  suite_class = _find_suite_class()
+  if cli_args.list_tests:
+    _print_test_names([suite_class])
+    sys.exit(0)
+  test_configs = config_parser.load_test_config_file(cli_args.config,
+                                                     cli_args.test_bed)
   config_count = len(test_configs)
   if config_count != 1:
     logging.error('Expect exactly one test config, found %d', config_count)
   config = test_configs[0]
   runner = test_runner.TestRunner(
       log_dir=config.log_path, testbed_name=config.testbed_name)
-  suite_class = _find_suite_class()
   suite = suite_class(runner, config)
+  console_level = logging.DEBUG if cli_args.verbose else logging.INFO
   ok = False
-  with runner.mobly_logger():
+  with runner.mobly_logger(console_level=console_level):
     try:
       suite.setup_suite(config.copy())
       try:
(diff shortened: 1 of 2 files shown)


xpconanfan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/875)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @uael)

___
*[`mobly/suite_runner.py` line 98 at r1](https://reviewable.io/reviews/google/mobly/875#-NQalzJI5ShQZ7pRt5n8:-NQalzJJ37FadsGWn9vb:b18peiz) ([raw file](https://github.com/google/mobly/blob/63699380829c94f636e2bd27680e761a37768628/mobly/suite_runner.py#L98)):*
> ```Python
>                       '--config',
>                       type=str,
>                       required=True,
> ```

why are we removing this?
___
*[`mobly/suite_runner.py` line 162 at r1](https://reviewable.io/reviews/google/mobly/875#-NQam6te2yyrCMm7CZbU:-NQam6te2yyrCMm7CZbV:b-a6ob8a) ([raw file](https://github.com/google/mobly/blob/3f59fb850d52c7b23eee60703639d2a02f9627ec/mobly/suite_runner.py#L162)):*
> ```Python
>     test_names = []
>     try:
>       cls.setup_generated_tests()
> ```

I think we should use `pre_run` instead of `setup_generated_tests` which is deprecated?
___
*[`mobly/suite_runner.py` line 162 at r1](https://reviewable.io/reviews/google/mobly/875#-NQamEQRDTvP_cTtm7nz:-NQamEQRDTvP_cTtm7o-:b-8ragni) ([raw file](https://github.com/google/mobly/blob/3f59fb850d52c7b23eee60703639d2a02f9627ec/mobly/suite_runner.py#L162)):*
> ```Python
>     test_names = []
>     try:
>       cls.setup_generated_tests()
> ```

since this might instantiate controller objects, should we call `clean_up` at the end?

uael (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/875)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @xpconanfan)

___
*[`mobly/suite_runner.py` line 98 at r1](https://reviewable.io/reviews/google/mobly/875#-NQalzJI5ShQZ7pRt5n8:-NQapMzx-Aayqr2k1M27:b-doeqpv) ([raw file](https://github.com/google/mobly/blob/63699380829c94f636e2bd27680e761a37768628/mobly/suite_runner.py#L98)):*

why are we removing this?
</blockquote></details>

Oh it's not, the diff is weird 
___
*[`mobly/suite_runner.py` line 162 at r1](https://reviewable.io/reviews/google/mobly/875#-NQam6te2yyrCMm7CZbU:-NQapWR78kkmCLOvMopG:bavzk50) ([raw file](https://github.com/google/mobly/blob/3f59fb850d52c7b23eee60703639d2a02f9627ec/mobly/suite_runner.py#L162)):*

I think we should use `pre_run` instead of `setup_generated_tests` which is deprecated?
</blockquote></details>

I just copy pasted all the missing logic from the test runner, I prefer this rather than having two separate behavior in runners, ok for you ?
___
*[`mobly/suite_runner.py` line 162 at r1](https://reviewable.io/reviews/google/mobly/875#-NQamEQRDTvP_cTtm7nz:-NQapYXgASsJJIgT50yC:b-ya81ng) ([raw file](https://github.com/google/mobly/blob/3f59fb850d52c7b23eee60703639d2a02f9627ec/mobly/suite_runner.py#L162)):*

since this might instantiate controller objects, should we call `clean_up` at the end?
</blockquote></details>

Same as above

xpconanfan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/875)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @uael)

___
*[`mobly/suite_runner.py` line 162 at r1](https://reviewable.io/reviews/google/mobly/875#-NQam6te2yyrCMm7CZbU:-NQapviYBpKauh0MgeLO:bm1hht7) ([raw file](https://github.com/google/mobly/blob/3f59fb850d52c7b23eee60703639d2a02f9627ec/mobly/suite_runner.py#L162)):*
<details><summary><i>Previously, uael (Lucas Abel) wrote…</i></summary><blockquote>

I just copy pasted all the missing logic from the test runner, I prefer this rather than having two separate behavior in runners, ok for you ?
</blockquote></details>

We should do the right thing to avoid more technical debt

uael (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/875)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @xpconanfan)

___
*[`mobly/suite_runner.py` line 162 at r1](https://reviewable.io/reviews/google/mobly/875#-NQam6te2yyrCMm7CZbU:-NQaq5Qr8rGPK1lnQNg6:b-svyigi) ([raw file](https://github.com/google/mobly/blob/3f59fb850d52c7b23eee60703639d2a02f9627ec/mobly/suite_runner.py#L162)):*

We should do the right thing to avoid more technical debt
</blockquote></details>

Do I take the responsibility to change both runner ?

uael:
@xpconanfan 
I just added two commits before the suite runner one in order to support using `_pre_run` instead of `setup_generated_tests` when retrieving test names :+1:

xpconanfan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/875)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @uael)

___
*[`mobly/base_test.py` line 353 at r4](https://reviewable.io/reviews/google/mobly/875#-NQb_MoYCVhz4GvB69CZ:-NQb_MoYCVhz4GvB69C_:b-4rt2gg) ([raw file](https://github.com/google/mobly/blob/82a41ba81e9d30e64e83ac930a80144241b2b6c2/mobly/base_test.py#L353)):*
> ```Python
>     """
>     # Make sure `_pre_run` is executed only once, since it might
>     # be called by `get_test_names` before `run`.
> ```

`list test name` cmd is a command that is executed and exit immediately.
Wouldn't it make more sense to make sure the suite calls cleanup as part of the `list test name`?

uael:
Added `_clean_up` :+1:

uael (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/875)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @xpconanfan)

___
*[`mobly/base_test.py` line 353 at r4](https://reviewable.io/reviews/google/mobly/875#-NQb_MoYCVhz4GvB69CZ:-NQbiFDF6feJn1PvIDl-:b-896fix) ([raw file](https://github.com/google/mobly/blob/82a41ba81e9d30e64e83ac930a80144241b2b6c2/mobly/base_test.py#L353)):*

`list test name` cmd is a command that is executed and exit immediately.
Wouldn't it make more sense to make sure the suite calls cleanup as part of the `list test name`?
</blockquote></details>

Done.

xpconanfan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/875)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @uael)

___
*[`mobly/suite_runner.py` line 98 at r1](https://reviewable.io/reviews/google/mobly/875#-NQalzJI5ShQZ7pRt5n8:-NQd0feyBxJNoEP4zW4E:bwrpuf6) ([raw file](https://github.com/google/mobly/blob/63699380829c94f636e2bd27680e761a37768628/mobly/suite_runner.py#L98)):*
<details><summary><i>Previously, uael (Lucas Abel) wrote…</i></summary><blockquote>

Oh it's not, the diff is weird 
</blockquote></details>

You are adding a group

I think list tests can also need configs sometimes.
E.g. some team may generate tests based on the device's info.

So let's not use `add_mutually_exclusive_group`?
___
*[`mobly/suite_runner.py` line 237 at r6](https://reviewable.io/reviews/google/mobly/875#-NQd1q_79Ym8t743UMdi:-NQd1q_79Ym8t743UMdj:bz7lv62) ([raw file](https://github.com/google/mobly/blob/fda7b4d4908b6c37815e033723cb08ee98ea4b6a/mobly/suite_runner.py#L237)):*
> ```Python
> 
>   if args.list_tests:
>     _print_test_names(test_classes)
> ```

why are there two blocks of these?
I'd expect only one place needs to trigger `_print_test_names`



_Code quote:_
```Python
  if args.list_tests:
    _print_test_names(test_classes)
    sys.exit(0)
```

uael:
I copy/pasted the parser and group from `teet_runner.py`, I would prefer to not diverge both implementation even if that make sense. Also, in avatar we use a `parameterized` decorator, is this something we can consider in future for generated tests instead of the logic in the base test class? It could reduce complexity a bit.

About the two `_print_test_names` calls, the two functions `run_suite_class` and `run_suite` are independents from each other and both parse arguments/load config and run tests so it make sense to me for them to behave the same way.

xpconanfan (approved):

PR #877, opened 2023-03-31
Fix unused 'name' argument in AttenuatorPath constructor

antofara:
store the path name for future usage

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/877)


antofara:
Gentle ping for reviewing this small change.
Some more context about the rationale of it: We want to retrieve at run time the attenuators paths to verify that the topology matches the expected one, and raise an error if that's not the case

xpconanfan (approved):

PR #886, opened 2023-05-05
Support am instrument options in snippet client

mhaoli:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/886)

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -901,7 +901,7 @@ class AndroidDevice:
     # So we need to wait for the device to come back before proceeding.
     self.adb.wait_for_device(timeout=DEFAULT_TIMEOUT_BOOT_COMPLETION_SECOND)
 
-  def load_snippet(self, name, package, snippet_client_configs=None):
+  def load_snippet(self, name, package, snippet_configs=None):
     """Starts the snippet apk with the given package name and connects.
 
     Examples:
@@ -917,9 +917,9 @@ class AndroidDevice:
         client. E.g. `name='maps'` attaches the snippet client to
         `ad.maps`.
       package: string, the package name of the snippet apk to connect to.
-      snippet_client_configs: snippet_client_v2.Config, the configuration
-        object for controlling the snippet client behaviors. See the docstring
-        of the `Config` class for supported configurations.
+      snippet_configs: snippet_client_v2.Config, the configuration object for
+        controlling the snippet behaviors. See the docstring of the `Config`
+        class for supported configurations.
 
     Raises:
       SnippetError: Illegal load operations are attempted.
@@ -930,7 +930,7 @@ class AndroidDevice:
           self,
           'Attribute "%s" already exists, please use a different name.' % name)
     self.services.snippets.add_snippet_client(
-        name, package, snippet_client_configs=snippet_client_configs
+        name, package, snippet_configs=snippet_configs
     )
 
   def unload_snippet(self, name):
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -901,7 +901,7 @@ class AndroidDevice:
     # So we need to wait for the device to come back before proceeding.
     self.adb.wait_for_device(timeout=DEFAULT_TIMEOUT_BOOT_COMPLETION_SECOND)
 
-  def load_snippet(self, name, package, instrument_options=None):
+  def load_snippet(self, name, package, snippet_client_configs=None):
     """Starts the snippet apk with the given package name and connects.
 
     Examples:
@@ -917,12 +917,9 @@ class AndroidDevice:
         client. E.g. `name='maps'` attaches the snippet client to
         `ad.maps`.
       package: string, the package name of the snippet apk to connect to.
-      instrument_options: dict[str, str], the Android am instrument options used
-        for controlling the `onCreate` process of the app under test. Note that
-        this should only be used for controlling the app launch process, options
-        for other purposes may not take effect and you should use snippet RPCs.
-        This is because Mobly snippet runner changes the subsequent
-        instrumentation process.
+      snippet_client_configs: snippet_client_v2.Config, the configuration
+        object for controlling the snippet client behaviors. See the docstring
+        of the `Config` class for supported configurations.
 
     Raises:
       SnippetError: Illegal load operations are attempted.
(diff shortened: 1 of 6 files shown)


mobly/controllers/android_device.py
```diff
@@ -917,6 +917,12 @@ def load_snippet(self, name, package):
         client. E.g. `name='maps'` attaches the snippet client to
         `ad.maps`.
       package: string, the package name of the snippet apk to connect to.
+      instrument_options: dict[str, str], the Android am instrument options used
+        for controlling the `onCreate` process of the app under test. Note that
```
xpconanfan, line 921: is there a specific set of options that's valid here?

mobly/controllers/android_device.py
```diff
@@ -901,7 +901,7 @@ def root_adb(self):
     # So we need to wait for the device to come back before proceeding.
     self.adb.wait_for_device(timeout=DEFAULT_TIMEOUT_BOOT_COMPLETION_SECOND)
 
-  def load_snippet(self, name, package):
+  def load_snippet(self, name, package, instrument_options=None):
```
xpconanfan, line 904: is there any other params that might make sense to pass in here?

Bigger question is, should we add instrumentation options, or a config type?

xpconanfan (changes requested):

mobly/controllers/android_device.py
mhaoli: I didn't find it and I think it is not documented in official doc.

mobly/controllers/android_device.py
mhaoli: Good point. I think using a config type would be better. It's possible that other params will be needed in future.

Working on adding it.

mobly/controllers/android_device_lib/snippet_client_v2.py
```diff
@@ -83,15 +83,16 @@ class Config:
   """A configuration class for configuring the snippet client.
 
   Attributes:
-    am_instrument_options: dict[str, str], the Android am instrument options
-      used for controlling the `onCreate` process of the app under test. Note
-      that this should only be used for controlling the app launch process,
-      options for other purposes may not take effect and you should use snippet
-      RPCs. This is because Mobly snippet runner changes the subsequent
-      instrumentation process.
+    am_instrument_options: The Android am instrument options used for
```
xpconanfan, line 86: why remove the type indication?

mobly/controllers/android_device.py
xpconanfan: Well, to be fair, it took many years for us to need this one..
Maybe check what other things have to be passed in from cli and fit that into the config type as well?

mhaoli: I took a look on it but not sure whether other params are needed. So I think we can add params when we really need them.

mobly/controllers/android_device_lib/snippet_client_v2.py
mhaoli: This is because I turned it into a dataclass attribute which requires type in code, so I deleted the type indication in docstring.

mobly/controllers/android_device.py
```diff
@@ -917,6 +917,9 @@ def load_snippet(self, name, package):
         client. E.g. `name='maps'` attaches the snippet client to
         `ad.maps`.
       package: string, the package name of the snippet apk to connect to.
+      snippet_client_configs: snippet_client_v2.Config, the configuration
```
xpconanfan, line 920: simply `config` or `client_config` should be fine?
It's pretty clear that we are dealing with snippet here..

also the args are eventually passed to the snippet server affecting the launch stage. So it's not exactly "client config"

mobly/controllers/android_device_lib/snippet_client_v2.py
```diff
@@ -76,6 +78,23 @@
 _CALLBACK_DEFAULT_TIMEOUT_SEC = 60 * 2
 
 
+@dataclasses.dataclass
+class Config:
+  """A configuration class for configuring the snippet client.
```
xpconanfan, line 83: the am instrumentation options are passed to the server side right?
so this is not exactly "configuring the snippet client"?

Sometimes less is more :P

xpconanfan (changes requested):

mobly/controllers/android_device.py
mhaoli: How about `snippet_configs`?

I have thought about this problem, but the name `config` is a bit confusing, because in this module there is another kind of config, i.e. controller configs. So I want to make it more clear. Also, you're right `client_config` is also not accurate, so I changed it `snippet_configs`.

mobly/controllers/android_device_lib/snippet_client_v2.py
mhaoli: Thanks. Removed it.

mobly/controllers/android_device.py
xpconanfan: Between the type hints and docstring, it's obvious which config this is right?

The method's name literally includes "snippet" in it..

mhaoli: Ok, make sense. Changed `snippet_configs` -> `configs`.

mobly/controllers/android_device_lib/snippet_client_v2.py
```diff
@@ -109,12 +128,14 @@ class SnippetClientV2(client_base.ClientBase):
...
   """
 
-  def __init__(self, package, ad):
+  def __init__(self, package, ad, configs=None):
     """Initializes the instance of Snippet Client V2.
 
     Args:
       package: str, see base class.
       ad: AndroidDevice, the android device object associated with this client.
+      configs: Config, the configuration object. See the docstring of the
```
xpconanfan, line 137: is this `configs` or `config`?
Seems like it's singular?

I see we already have `def load_config(self, config):`

Same for all the other places

xpconanfan (changes requested):

mobly/controllers/android_device_lib/snippet_client_v2.py
mhaoli: I used `configs` because the config arg of all services are named in plural form, i.e. BaseService's constructor is `__init__(..., configs)`.

So looks like there's a conflct in existing config arg names. Based on this, what name would you suggest to use here? We can insist on

xpconanfan: I took a look in the codebase.
Let's use singular here.

mhaoli: Done.

xpconanfan (approved):

PR #891, opened 2023-09-01
Fix whitespace stripping in _collect_process_tree

puradox:
Fix UtilsTest.test_stop_standing_subproc_and_descendants failure for versions of ps that do not include a space after each newline.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/891)


xpconanfan (changes requested):
Add unit tests for the cases you are trying to cover with this change?

xpconanfan:
Looked into the related code more.

Looks like the current logic for stopping all children of a pid is a bit cumbersome.
We should clean it up soon.

So merging this in as is since it'll be cleaned up next.

xpconanfan (approved):

PR #893, opened 2023-09-01
Remove unused variables

puradox:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/893)

### DIFF
--- a/tests/mobly/controllers/android_device_lib/adb_test.py
+++ b/tests/mobly/controllers/android_device_lib/adb_test.py
@@ -168,7 +168,7 @@ class AdbTest(unittest.TestCase):
     mock_popen.return_value.stdout.readline.side_effect = ['1', '2', '']
     mock_handler = mock.MagicMock()
 
-    err = adb.AdbProxy()._execute_and_process_stdout(['fake_cmd'],
+    _ = adb.AdbProxy()._execute_and_process_stdout(['fake_cmd'],
                                                      shell=False,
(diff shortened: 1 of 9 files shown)


tests/mobly/controllers/android_device_lib/adb_test.py
```diff
@@ -285,7 +285,7 @@ def test_execute_and_process_stdout_when_handler_crash(self, mock_popen):
     mock_handler.side_effect = ['', TypeError('fake crash'), '', '']
 
     with self.assertRaisesRegex(TypeError, 'fake crash'):
-      err = adb.AdbProxy()._execute_and_process_stdout(['fake_cmd'],
+      _ = adb.AdbProxy()._execute_and_process_stdout(['fake_cmd'],
```
xpconanfan, line 288: why keep the assignment at all?
same for elsewhere

xpconanfan (changes requested):

tests/mobly/controllers/android_device_lib/adb_test.py
puradox: This is to inform readers these return values are intentionally dropped, rather than erronously forgotten about.

xpconanfan: I don't think it's needed since we are not doing anything with the return values.
Could you remove all?

xpconanfan:
Closing due to lack of responses

PR #894, opened 2023-09-13
Run Gitlab CI daily

ko1in1u:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/894)


ko1in1u:
No need to run unittest daily, run when PR is created is enough.

PR #899, opened 2023-10-23
Allow configuring snippet user.

nkprasad12:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/899)

### DIFF
--- a/mobly/controllers/android_device_lib/snippet_client_v2.py
+++ b/mobly/controllers/android_device_lib/snippet_client_v2.py
@@ -18,7 +18,7 @@ import enum
 import json
 import re
 import socket
-from typing import Dict
+from typing import Dict, Union
 
 from mobly import utils
 from mobly.controllers.android_device_lib import adb
@@ -89,11 +89,12 @@ class Config:
       other purposes may not take effect and you should use snippet RPCs. This
       is because Mobly snippet runner changes the subsequent instrumentation
       process.
+    user_id: The user id under which to launch the snippet process.
   """
 
   am_instrument_options: Dict[str, str] = dataclasses.field(
       default_factory=dict)
-
+  user_id: Union[int, None] = None
 
 class ConnectionHandshakeCommand(enum.Enum):
   """Commands to send to the server when sending the handshake request.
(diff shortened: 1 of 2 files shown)


mhaoli (approved):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/899)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @nkprasad12 and @xpconanfan)

___
*[`mobly/controllers/android_device_lib/snippet_client_v2.py` line 194 at r1](https://reviewable.io/reviews/google/mobly/899#-NhQCr5w609km_gpU7Bg:-NhQCr5w609km_gpU7Bh:b-us9jn4) ([raw file](https://github.com/google/mobly/blob/c17a74ddf1ab6d8a6ce6394a2a84c2a07ea44f25/mobly/controllers/android_device_lib/snippet_client_v2.py#L194)):*
> ```Python
>     self._disable_hidden_api_blocklist()
> 
>   def _validate_snippet_app_on_device(self):
> ```

This method checks that the snippet app is installed under the user_id. Better to add a test case to verify that this method is using the user_id from the config?



_Code quote:_
```Python
_validate_snippet_app_on_device
```

mobly/controllers/android_device_lib/snippet_client_v2.py
```diff
@@ -89,11 +89,12 @@ class Config:
       other purposes may not take effect and you should use snippet RPCs. This
       is because Mobly snippet runner changes the subsequent instrumentation
       process.
+    user_id: The user id under which to launch the snippet process.
   """
 
   am_instrument_options: Dict[str, str] = dataclasses.field(
       default_factory=dict)
-
+  user_id: Union[int, None] = None
```
ko1in1u, line 97: Should we just use `Optional`?

nkprasad12: > Should we just use `Optional`?

No, because that is new in 3.10 but we officially support 3.7 / 3.8 / 3.9 too in I.

mhaoli (approved):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/899)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @ko1in1u and @xpconanfan)

PR #904, opened 2023-12-14
Using tcp:0 in adb forward to pick available host port in snippet client

mhaoli:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/904)

### DIFF
--- a/mobly/controllers/android_device_lib/snippet_client_v2.py
+++ b/mobly/controllers/android_device_lib/snippet_client_v2.py
@@ -408,8 +408,8 @@ class SnippetClientV2(client_base.ClientBase):
       if self.host_port in adb.list_occupied_adb_ports():
         raise errors.Error(
             self._device,
-            f'Cannot forward to host port {self.host_port} because adb has forwarded'
-            ' another device port to it.'
+            f'Cannot forward to host port {self.host_port} because adb has'
+            ' forwarded another device port to it.',
         )
       host_port = self.host_port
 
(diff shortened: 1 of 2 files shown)


mobly/controllers/android_device_lib/snippet_client_v2.py
```diff
@@ -403,9 +403,10 @@ def make_connection(self):
 
   def _forward_device_port(self):
     """Forwards the device port to a host port."""
-    if not self.host_port:
-      self.host_port = utils.get_available_host_port()
-    self._adb.forward([f'tcp:{self.host_port}', f'tcp:{self.device_port}'])
+    host_port = self.host_port or 0
+    stdout = self._adb.forward([f'tcp:{host_port}', f'tcp:{self.device_port}'])
```
xpconanfan, line 407: if the host port passed in is already occupied, would the cmd fail and raise an error?

xpconanfan:
Should probably mark `utils.get_available_host_port` as deprecated
add a warning log and update the docstring to reflect.
can be in a separate CL

mobly/controllers/android_device_lib/snippet_client_v2.py
```diff
@@ -403,9 +403,10 @@ def make_connection(self):
...
+    # Example stdout: '12345\n'
```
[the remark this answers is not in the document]
mhaoli: Yes, it will print the forwarded port, so it's safe to overide the var `self.host_port`.

mobly/controllers/android_device_lib/snippet_client_v2.py
mhaoli: Thanks for noticing a bug here. Fixed.

For most situations, it will raise an error. But if the port is occupied by adb forwarding, then adb will not raise an error but silently override the original binding. Added a check and raise an error for this situation.

mhaoli:
> Should probably mark `utils.get_available_host_port` as deprecated add a warning log and update the docstring to reflect. can be in a separate CL

Sg. And yeah, let's submit this CL and make sure this doesn't break users first :)

xpconanfan (approved):

PR #907, opened 2024-01-11
Deprecate the util method get_available_host_port

mhaoli:
Deprecate `utils.get_available_host_port`. After #904 it is no longer needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/907)


mobly/utils.py
```diff
@@ -603,13 +603,22 @@ def wait_for_standing_subprocess(proc, timeout=None):
...
   Returns:
     An integer representing a port number on the host available for adb
     forward.
 
   Raises:
     Error: when no port is found after MAX_PORT_ALLOCATION_RETRY times.
   """
+  logging.warning(
+      'The method mobly.utils.get_available_host_port is deprecated because it '
+      'is inherently racy. Instead you should pass "tcp:0" to adb and let adb '
```
xpconanfan, line 618: just say this method is unreliable.

no need to say "racy"

xpconanfan (changes requested):

mobly/utils.py
mhaoli: Modified

xpconanfan (approved):

PR #908, opened 2024-01-19
Adding aeroflex attenuator support to mobly

Anand-G-Choudhari:
Adding aeroflex attenuator support to mobly for dependency project.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/908)


xpconanfan:
We will not check this in as it's a copy of code from AOSP

You should NOT be doing this sort of things.

Also you don't have to check the controller code into mobly in order to use it as Mobly's controller model is flexible

PR #917, opened 2024-04-11
Log the command execution result in utils.run_command

mhaoli:
We can log the command execution result in each call of utils.run_command.

With this, we no longer need to print them in adb module.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/917)


mhaoli:
Fixing format problem in #918

xpconanfan (approved):

xpconanfan:
CI failing?

mhaoli:
CI failing is caused by pyink format problem. Fixing it in #918, let's submit it first.

PR #923, opened 2024-06-01
Use self.is_adb_detectable() to prevent build_info from None

minghsikang:
The logic now is trying to use build_info when it is not in bootloader, but if the device isn't detectable in neither adb nor fastboot, the exception occurs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/923)

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -844,7 +844,7 @@ class AndroidDevice:
 
   @property
   def is_rootable(self):
-    return not self.is_bootloader and self.build_info['debuggable'] == '1'
+    return self.is_adb_detectable() and self.build_info['debuggable'] == '1'
 
   @functools.cached_property
   def model(self):
--- a/tests/lib/mock_android_device.py
+++ b/tests/lib/mock_android_device.py
@@ -87,6 +87,7 @@ class MockAdbProxy:
       mock_properties=None,
       installed_packages=None,
       instrumented_packages=None,
+      adb_detectable=True,
   ):
     self.serial = serial
     self.fail_br = fail_br
@@ -101,6 +102,7 @@ class MockAdbProxy:
     self.installed_packages = installed_packages
     if instrumented_packages is None:
       instrumented_packages = []
+    self.adb_detectable = adb_detectable
     self.installed_packages = installed_packages
     self.instrumented_packages = instrumented_packages
 
@@ -153,6 +155,12 @@ class MockAdbProxy:
     )
     if expected not in args:
       raise Error('"Expected "%s", got "%s"' % (expected, args))
+
+  def devices(self):
+    out = b'xxxx\tdevice\nyyyy\tdevice'
+    if self.adb_detectable:
+      out += f'\n{self.serial}\tdevice'.encode()
+    return out
 
   def __getattr__(self, name):
     """All calls to the none-existent functions in adb proxy would
--- a/tests/mobly/controllers/android_device_lib/services/logcat_test.py
+++ b/tests/mobly/controllers/android_device_lib/services/logcat_test.py
@@ -149,7 +149,7 @@ class LogcatTest(unittest.TestCase):
 
   @mock.patch(
       'mobly.controllers.android_device_lib.adb.AdbProxy',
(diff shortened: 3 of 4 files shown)


xpconanfan:
The logic is still trying to access `build_info` right? What's the difference?

minghsikang:
> The logic is still trying to access `build_info` right? What's the difference?

The difference is that we can only use build_info when the device is detectable. Otherwise, the build_info is None.

Here is the context: 
when I tried to init android device, is_rootable could raise exception if the device is not found. 
> return not self.is_bootloader and self.build_info['debuggable'] == '1'

That's why it is not in bootloader but still trying to access build_info

Hence, I come up with two solutions
1. change `not is_bootloader` to `is_adb_detectable` to make sure we won't access build_info when it is None. (current change)
2.  add a parameter to not use root_adb in init method.

WDYT?

mhaoli:
> when I tried to init android device, is_rootable could raise exception if the device is not found.

Are you meaning you are initializing an `AndroidDevice` object while there is no corresponding Android device? 

We assume each `AndroidDevice` object corresponds to one Android device. What are you going to do if there's no Android device?

minghsikang:
> Are you meaning you are initializing an `AndroidDevice` object while there is no corresponding Android device?

Partially correct. There is the android device, but it is not detectable by adb or fastboot (even not in rom recovery mode and need ftdi commands to bring it back).

> We assume each `AndroidDevice` object corresponds to one Android device. What are you going to do if there's no Android device?

We will use some other recover scripts/tools to bring it back to fastboot or android and then use Android device.

mhaoli (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/923)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @minghsikang)

___
*[`tests/lib/mock_android_device.py` line 90 at r1](https://reviewable.io/reviews/google/mobly/923#-NzS-YMDE6YQO2dxk_4Z:-NzS-YMDE6YQO2dxk_4_:b-uxdd4n) ([raw file](https://github.com/google/mobly/blob/70c068607e900aa4fa80c2f4b8d23f81b9d2fb77/tests/lib/mock_android_device.py#L90)):*
> ```Python
>       installed_packages=None,
>       instrumented_packages=None,
>       detectable=True,
> ```

Renaming this to `adb_detectable` will be more accurate?

mhaoli:
> > Are you meaning you are initializing an `AndroidDevice` object while there is no corresponding Android device?
> 
> Partially correct. There is the android device, but it is not detectable by adb or fastboot (even not in rom recovery mode and need ftdi commands to bring it back).
> 
> > We assume each `AndroidDevice` object corresponds to one Android device. What are you going to do if there's no Android device?
> 
> We will use some other recover scripts/tools to bring it back to fastboot or android and then use Android device.

I see.

I think in Mobly repo this change makes sense to prevent using a None as a dict.

But for your usage, sounds like it would make more sense to first recover the device and then initialize `AndroidDevice` object?

minghsikang:
> But for your usage, sounds like it would make more sense to first recover the device and then initialize `AndroidDevice` object?

Totally agree.

mhaoli (approved):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/923)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @minghsikang)

minghsikang (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/923)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @mhaoli and @minghsikang)

___
*[`tests/lib/mock_android_device.py` line 90 at r1](https://reviewable.io/reviews/google/mobly/923#-NzS-YMDE6YQO2dxk_4Z:-NzSiOk1AzoY7aanI5Dl:b-896fix) ([raw file](https://github.com/google/mobly/blob/58b58b0fff1bc14fe70d78bf01dbb1d4a006f36a/tests/lib/mock_android_device.py#L90)):*

Renaming this to `adb_detectable` will be more accurate?
</blockquote></details>

Done.

minghsikang:
It seems need your help to resolve the comment.

PR #924, opened 2024-06-07
[WIP] Make AndroidDevice#build_info strictly returns a dict

mhaoli:
Now AndroidDevice#build_info returns a None when in fastboot and a dict otherwise.

I'm not quite sure if we were intentionally returning a None when the device is in fastboot. But I think making it strictly return a dict looks more reliable. So users won't get errors like 

"TypeError: 'NoneType' object is not subscriptable"

WDYT @xpconanfan ?

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/924)

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -815,7 +815,7 @@ class AndroidDevice:
     """
     if self.is_bootloader:
       self.log.error('Device is in fastboot mode, could not get build info.')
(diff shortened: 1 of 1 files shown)


mhaoli:
I can update docstring and other things if we agree this change makes sense.

xpconanfan (approved):

PR #934, opened 2024-08-20
Support passing stdout and stderr in start_standing_subprocess

mhaoli:
`start_standing_subprocess` will pass `stdout` and `stderr` to `Popen` as is.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/934)

### DIFF
--- a/mobly/utils.py
+++ b/mobly/utils.py
@@ -507,7 +507,13 @@ def run_command(
   return process.returncode, out, err
 
 
-def start_standing_subprocess(cmd, shell=False, env=None):
+def start_standing_subprocess(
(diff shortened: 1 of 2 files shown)


xpconanfan:
What's the use case for passing in these two?

mhaoli:
> What's the use case for passing in these two?

I'm using this to run a long running process (i.e., dongle sniffer) and stream the process stdout and stderr to log files. 

Without this PR, we can only get stdout and stderr by calling `process.communicate()`.

xpconanfan (approved):

PR #938, opened 2024-08-30
Support selecting test cases in a class with regex

### LINKED ISSUE #887: Allow regex matching for specifying test cases
## Summary

The `--test_case` CLI option currently only supports specifying full test names. This makes it tedious to run only a portion of a large test class (100+ tests) where there are many similar test names, such as when test cases are generated with `BaseTestClass.generate_tests`.

This issue discusses the design of adding regex matching capability for selecting which tests to run. Is this functionality something we want to include in `BaseTestClass` for every Mobly user? Or should the user implement this for their own base test class? What should this look like? I propose we add a new CLI option which allows for test case regex matching:

```
python FooBarTest.py --test_case_regex 'test_(foo|bar)'
```

## Proposal

Add a _new_ CLI option named `--test_case_regex` (also aliased to `--tests_regex`, akin to `--tests`) that performs a _full match_ on test names using regex syntax defined by Python Standard Library [`re`](https://docs.python.org/3/library/re.html).

- Using a new CLI option benefits the user by being backwards compatible with `--test_case` and allowing for using both options in conjunction.
- Performing only [full matches](https://docs.python.org/3/library/re.html#re.Pattern.fullmatch) enforces full specification of test names and behaves more predictably since partial matches are not possible (regex `test_can` doesn't match `test_cannot`). In addition, this also removes the need to specify anchors (e.g. `^`, `$`), making regexes more readable (`^test_foo_.+$` vs `test_foo_.+`).
- Using Python regex syntax should be familiar to most users as it is similar to Perl regex. It implements all common regex and is easy to add to Mobly :)

`--test_case_regex` can be repeated to allow for matching on multiple patterns. This option can also work in conjunction with the `--list_tests` option; tests matched the regex will be listed rather than ran.

## Alternatives explored

Modification of the `--test_case` option was considered instead of adding a new option. However, there is no guarantee that test names contain regex special characters. If the existing option was to be changed, it would be a breaking change.

I've considered using [`fnmatch`](https://docs.python.org/3/library/fnmatch.html) as the pattern matching syntax, similar to ACTS' [`test_runner`](https://android.googlesource.com/platform/tools/test/connectivity/+/master/acts/framework/acts/test_runner.py#209), but I've found this syntax confusing since test names are being matched against, not file names. For example `test_*` works as expected but `test_(a|b)` doesn't. 

## Feedback

Let me know what you think!

xpconanfan:
Now users can enter regex strings with prefix `re:` in the `test_cases` field to select with regex. Only full matches apply.

Fixes #887

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/938)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -1044,10 +1044,10 @@ class BaseTestClass:
   ):
     matching_name_tuples = []
     for name in current_valid_names:
-      if re.match(test_name_regex, name):
+      if re.fullmatch(test_name_regex, name):
         matching_name_tuples.append((name, getattr(self, name)))
     for name in self._generated_test_table.keys():
-      if re.match(test_name_regex, name):
+      if re.fullmatch(test_name_regex, name):
         self._assert_valid_test_name(name)
         matching_name_tuples.append((name, self._generated_test_table[name]))
     if not matching_name_tuples:
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -19,6 +19,7 @@ import functools
 import inspect
 import logging
 import os
+import re
 import sys
 
 from mobly import controller_manager
@@ -31,6 +32,7 @@ from mobly import utils
 # Macro strings for test result reporting.
 TEST_CASE_TOKEN = '[Test]'
 RESULT_LINE_TEMPLATE = TEST_CASE_TOKEN + ' %s %s'
+TEST_SELECTOR_REGEX_PREFIX = 're:'
 
 TEST_STAGE_BEGIN_LOG_TEMPLATE = '[{parent_token}]#{child_token} >>> BEGIN >>>'
 TEST_STAGE_END_LOG_TEMPLATE = '[{parent_token}]#{child_token} <<< END <<<'
@@ -1003,7 +1005,8 @@ class BaseTestClass:
     """Resolves test method names to bound test methods.
 
     Args:
-      test_names: A list of strings, each string is a test method name.
+      test_names: A list of strings, each string is a test method name or a
+        regex for matching test names.
 
     Returns:
       A list of tuples of (string, function). String is the test method
@@ -1014,20 +1017,52 @@ class BaseTestClass:
         This can only be caused by user input.
     """
     test_methods = []
+    # Call once and reuse the list through this function for efficiency.
+    current_valid_names = self.get_existing_test_names()
+    # Process the test name selector one by one.
     for test_name in test_names:
-      if not test_name.startswith('test_'):
-        raise Error(
-            'Test method name %s does not follow naming '
-            'convention test_*, abort.' % test_name
+      if test_name.startswith(TEST_SELECTOR_REGEX_PREFIX):
+        # process the selector as a regex.
+        regex_matching_methods = self._get_regex_matching_test_methods(
+            test_name[len(TEST_SELECTOR_REGEX_PREFIX) :], current_valid_names
         )
+        test_methods += regex_matching_methods
+        continue
+      # process the selector as a regular test name string.
+      self._assert_valid_test_name(test_name)
+      if test_name not in current_valid_names:
+        raise Error('%s does not have test method %s.' % (self.TAG, test_name))
       if hasattr(self, test_name):
         test_method = getattr(self, test_name)
       elif test_name in self._generated_test_table:
         test_method = self._generated_test_table[test_name]
-      else:
-        raise Error('%s does not have test method %s.' % (self.TAG, test_name))
       test_methods.append((test_name, test_method))
     return test_methods
+
+  def _get_regex_matching_test_methods(
+      self, test_name_regex, current_valid_names
+  ):
+    matching_name_tuples = []
+    for name in current_valid_names:
+      if re.match(test_name_regex, name):
+        matching_name_tuples.append((name, getattr(self, name)))
+    for name in self._generated_test_table.keys():
+      if re.match(test_name_regex, name):
+        self._assert_valid_test_name(name)
+        matching_name_tuples.append((name, self._generated_test_table[name]))
+    if not matching_name_tuples:
+      raise Error(
+          f'{test_name_regex} does not match with any valid test case '
+          f'in {self.TAG}, abort!'
+      )
+    return matching_name_tuples
+
+  def _assert_valid_test_name(self, test_name):
+    if not test_name.startswith('test_'):
+      raise Error(
+          'Test method name %s does not follow naming '
+          'convention test_*, abort.' % test_name
+      )
 
   def _skip_remaining_tests(self, exception):
     """Marks any requested test that has not been executed in a class as
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -1017,21 +1017,19 @@ class BaseTestClass:
         This can only be caused by user input.
     """
     test_methods = []
-    # Call once and reuse the list through this function for efficiency.
-    current_valid_names = self.get_existing_test_names()
     # Process the test name selector one by one.
     for test_name in test_names:
       if test_name.startswith(TEST_SELECTOR_REGEX_PREFIX):
         # process the selector as a regex.
         regex_matching_methods = self._get_regex_matching_test_methods(
-            test_name[len(TEST_SELECTOR_REGEX_PREFIX) :], current_valid_names
+            test_name.removeprefix(TEST_SELECTOR_REGEX_PREFIX)
         )
         test_methods += regex_matching_methods
         continue
       # process the selector as a regular test name string.
       self._assert_valid_test_name(test_name)
-      if test_name not in current_valid_names:
-        raise Error('%s does not have test method %s.' % (self.TAG, test_name))
+      if test_name not in self.get_existing_test_names():
+        raise Error(f'{self.TAG} does not have test method {test_name}.')
       if hasattr(self, test_name):
         test_method = getattr(self, test_name)
       elif test_name in self._generated_test_table:
@@ -1039,17 +1037,18 @@ class BaseTestClass:
       test_methods.append((test_name, test_method))
     return test_methods
 
-  def _get_regex_matching_test_methods(
-      self, test_name_regex, current_valid_names
-  ):
+  def _get_regex_matching_test_methods(self, test_name_regex):
     matching_name_tuples = []
-    for name in current_valid_names:
-      if re.fullmatch(test_name_regex, name):
-        matching_name_tuples.append((name, getattr(self, name)))
-    for name in self._generated_test_table.keys():
-      if re.fullmatch(test_name_regex, name):
+    for name, method in inspect.getmembers(self, callable):
+      if (
+          name.startswith('test_')
+          and re.fullmatch(test_name_regex, name) is not None
+      ):
+        matching_name_tuples.append((name, method))
+    for name, method in self._generated_test_table.items():
+      if re.fullmatch(test_name_regex, name) is not None:
         self._assert_valid_test_name(name)
-        matching_name_tuples.append((name, self._generated_test_table[name]))
+        matching_name_tuples.append((name, method))
     if not matching_name_tuples:
       raise Error(
           f'{test_name_regex} does not match with any valid test case '
--- a/mobly/base_suite.py
+++ b/mobly/base_suite.py
@@ -47,8 +47,10 @@ class BaseSuite(abc.ABC):
       config: config_parser.TestRunConfig, the config to run the class with. If
         not specified, the default config passed from google3 infra is used.
       tests: list of strings, names of the tests to run in this test class, in
-        the execution order. If not specified, all tests in the class are
-        executed.
+        the execution order. Or a string with prefix `re:` for full regex match
+        of test cases; all matched test cases will be executed; an error is
+        raised if no match is found.
+        If not specified, all tests in the class are executed.
       name_suffix: string, suffix to append to the class name for reporting.
         This is used for differentiating the same class executed with different
         parameters in a suite.
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -126,8 +126,12 @@ def parse_mobly_cli_args(argv):
       '--test_case',
       nargs='+',
       type=str,
-      metavar='[test_a test_b...]',
-      help='A list of tests in the test class to execute.',
+      metavar='[test_a test_b re:test_(c|d)...]',
+      help=(
+          'A list of tests in the test class to execute. Each value can be a '
+          'test name string or a `re:` prefixed string for full regex match of'
+          ' test names.'
+      ),
   )
   parser.add_argument(
       '-tb',
--- a/tests/mobly/base_test_test.py
+++ b/tests/mobly/base_test_test.py
@@ -201,6 +201,52 @@ class BaseTestTest(unittest.TestCase):
     actual_record = bt_cls.results.passed[0]
     self.assertEqual(actual_record.test_name, 'test_something')
 
+  def test_cli_test_selection_with_regex(self):
+    class MockBaseTest(base_test.BaseTestClass):
+
+      def __init__(self, controllers):
+        super().__init__(controllers)
+        self.tests = ('test_never',)
+
+      def test_something_1(self):
+        pass
+
+      def test_something_2(self):
+        pass
+
+      def test_something_3(self):
+        pass
+
+      def test_never(self):
+        # This should not execute it's not selected by cmd line input.
+        never_call()
+
+    bt_cls = MockBaseTest(self.mock_test_cls_configs)
+    bt_cls.run(test_names=['re:test_something*'])
+    self.assertEqual(len(bt_cls.results.passed), 3)
+    self.assertEqual(bt_cls.results.passed[0].test_name, 'test_something_1')
+    self.assertEqual(bt_cls.results.passed[1].test_name, 'test_something_2')
+    self.assertEqual(bt_cls.results.passed[2].test_name, 'test_something_3')
+
+  def test_cli_test_selection_with_regex_fail_by_convention(self):
+    class MockBaseTest(base_test.BaseTestClass):
+
+      def __init__(self, controllers):
+        super().__init__(controllers)
+        self.tests = ('test_never',)
+
+      def test_something(self):
+        pass
+
+    bt_cls = MockBaseTest(self.mock_test_cls_configs)
+    expected_msg = (
+        r'not_a_test_something does not match with any valid test case in '
+        r'MockBaseTest, abort!'
+    )
+    with self.assertRaisesRegex(base_test.Error, expected_msg):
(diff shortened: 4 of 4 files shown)


mobly/base_test.py
```diff
@@ -1014,21 +1017,53 @@ def _get_test_methods(self, test_names):
...
         test_method = self._generated_test_table[test_name]
-      else:
-        raise Error('%s does not have test method %s.' % (self.TAG, test_name))
       test_methods.append((test_name, test_method))
     return test_methods
 
+  def _get_regex_matching_test_methods(
+      self, test_name_regex, current_valid_names
+  ):
+    matching_name_tuples = []
```
ko1in1u, line 1045: The constant name seems not accurately reflect its type. Should we rename it to `matching_name_lists`?

ko1in1u (approved):

mobly/base_test.py
xpconanfan: sorry, can you clarify?
this is a list of tuples, no?

ko1in1u: Ah, got it. At first glance, it looks like a tuple type from its name.

xianyuanjia (changes requested):
> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/mobly/938)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @xpconanfan)

___
*[`mobly/base_test.py` line 1039 at r1](https://reviewable.io/reviews/google/mobly/938#-O5Vk3928RnFAYwupKAt:-O5Vk3928RnFAYwupKAu:b-b1ecrn) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1039)):*
> ```Python
>       elif test_name in self._generated_test_table:
>         test_method = self._generated_test_table[test_name]
>       test_methods.append((test_name, test_method))
> ```

What happens when a test method is matched by a regex but also specified by an exact string? It seems that the test would be duplicated.



_Code quote:_
```Python
      test_methods.append((test_name, test_method))
```
___
*[`mobly/base_test.py` line 1052 at r1](https://reviewable.io/reviews/google/mobly/938#-O5VpOSB3sPGdc9xE0Vk:-O5VpOSB3sPGdc9xE0Vl:b4hmcy5) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1052)):*
> ```Python
>       if re.fullmatch(test_name_regex, name):
>         self._assert_valid_test_name(name)
>         matching_name_tuples.append((name, self._generated_test_table[name]))
> ```

`current_valid_names` contains both the predefined and generated test names.

So if you loop through `current_valid_names`, you can get generated tests, which would fail on `getattr`.

This is probably something that should also be verified in a unit test.

xpconanfan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/938)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @xianyuanjia)

___
*[`mobly/base_test.py` line 1039 at r1](https://reviewable.io/reviews/google/mobly/938#-O5Vk3928RnFAYwupKAt:-O5VtODhBhJE7zN6OS17:b-hfhbtp) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1039)):*

What happens when a test method is matched by a regex but also specified by an exact string? It seems that the test would be duplicated.
</blockquote></details>

We have always allowed the same test case in a class to be selected multiple times actually.

eric100lin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/938)* status: 1 of 4 files reviewed, 9 unresolved discussions (waiting on @xianyuanjia and @xpconanfan)

___
*[`mobly/base_test.py` line 1027 at r1](https://reviewable.io/reviews/google/mobly/938#-O5W23t7B41sMGFi49ZG:-O5W23t7B41sMGFi49ZH:b-xu63rn) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1027)):*
> ```Python
>         # process the selector as a regex.
>         regex_matching_methods = self._get_regex_matching_test_methods(
>             test_name[len(TEST_SELECTOR_REGEX_PREFIX) :], current_valid_names
> ```

nit: Consider removeprefix (added in version 3.9) and pass as keyword arguments .

```
        regex_matching_methods = self._get_regex_matching_test_methods(
            test_name_regex=test_name.removeprefix(TEST_SELECTOR_REGEX_PREFIX),
            current_valid_names=current_valid_names,
        )
```



_Code quote:_
```Python
test_name[len(TEST_SELECTOR_REGEX_PREFIX) :], current_valid_names
```
___
*[`mobly/base_test.py` line 1034 at r1](https://reviewable.io/reviews/google/mobly/938#-O5W4JcY8nGGVR33t88K:-O5W4JcY8nGGVR33t88L:b49u8lh) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1034)):*
> ```Python
>       self._assert_valid_test_name(test_name)
>       if test_name not in current_valid_names:
>         raise Error('%s does not have test method %s.' % (self.TAG, test_name))
> ```

nit: consider f-string



_Code quote:_
```Python
'%s does not have test method %s.'
```
___
*[`mobly/base_test.py` line 1047 at r1](https://reviewable.io/reviews/google/mobly/938#-O5W4sQE9rat6wJeA8kL:-O5W4sQE9rat6wJeA8kM:bki30g7) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1047)):*
> ```Python
>     matching_name_tuples = []
>     for name in current_valid_names:
>       if re.fullmatch(test_name_regex, name):
> ```

https://google.github.io/styleguide/pyguide.html#2144-decision
Always use `if foo is None:` (or `is not None`) to check for a `None` value:
```
if re.fullmatch(test_name_regex, name) is not None:
```



_Code quote:_
```Python
if re.fullmatch(test_name_regex, name)
```
___
*[`mobly/base_test.py` line 1050 at r1](https://reviewable.io/reviews/google/mobly/938#-O5W5iY00aUgQtkus9US:-O5W5iY18onk3rvRcnUR:b-7ropln) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1050)):*
> ```Python
>         matching_name_tuples.append((name, getattr(self, name)))
>     for name in self._generated_test_table.keys():
>       if re.fullmatch(test_name_regex, name):
> ```

ditto



_Code quote:_
```Python
if re.fullmatch(test_name_regex, name)
```
___
*[`mobly/base_test.py` line 1052 at r1](https://reviewable.io/reviews/google/mobly/938#-O5W6phFC0zWOUEO59v8:-O5W6phFC0zWOUEO59v9:bj9eega) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1052)):*
> ```Python
>       if re.fullmatch(test_name_regex, name):
>         self._assert_valid_test_name(name)
>         matching_name_tuples.append((name, self._generated_test_table[name]))
> ```

nit: consider taking both key and value
```
    for name, method in self._generated_test_table.items():
      if re.fullmatch(test_name_regex, name):
        self._assert_valid_test_name(name)
        matching_name_tuples.append((name, method))
```



_Code quote:_
```Python
    for name in self._generated_test_table.keys():
      if re.fullmatch(test_name_regex, name):
        self._assert_valid_test_name(name)
        matching_name_tuples.append((name, self._generated_test_table[name]))
```
___
*[`tests/mobly/base_test_test.py` line 205 at r1](https://reviewable.io/reviews/google/mobly/938#-O5W9Tl8Bud328YpS15a:-O5W9Tl8Bud328YpS15b:b-qy6c6c) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/tests/mobly/base_test_test.py#L205)):*
> ```Python
> 
>   def test_cli_test_selection_with_regex(self):
>     class MockBaseTest(base_test.BaseTestClass):
> ```

Please cover `generate_tests` case, I found the following error:
```
  File "mobly/mobly/base_test.py", line 1026, in _get_test_methods
    regex_matching_methods = self._get_regex_matching_test_methods(
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "mobly/mobly/base_test.py", line 1049, in _get_regex_matching_test_methods
    matching_name_tuples.append((name, getattr(self, name)))
                                       ^^^^^^^^^^^^^^^^^^^
AttributeError: 'MockBaseTest' object has no attribute 'test_something_4'
```
when I add codes like:
```
      def pre_run(self):
        self.generate_tests(
            test_logic=self.logic,
            name_func=self.name_gen,
            arg_sets=[('something', 4)],
        )

      def name_gen(self, a, b):
        return 'test_%s_%s' % (a, b)

      def logic(self, a, b):
        pass
```

You might need to check `hasattr(self, name)` in the first loop.



_Code quote:_
```Python
MockBaseTest
```

eric100lin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/938)* status: 1 of 4 files reviewed, 10 unresolved discussions (waiting on @xianyuanjia and @xpconanfan)

___
*[`mobly/base_test.py` line 1048 at r1](https://reviewable.io/reviews/google/mobly/938#-O5WHc136tQsM0PBQjnK:-O5WHc136tQsM0PBQjnL:b8cz664) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1048)):*
> ```Python
>     for name in current_valid_names:
>       if re.fullmatch(test_name_regex, name):
>         matching_name_tuples.append((name, getattr(self, name)))
> ```

I think we should also check `self._assert_valid_test_name(name)` here,
otherwise I can inject invalid tests like this :)
```
    class MockBaseTest(base_test.BaseTestClass):

        # ...

      def get_existing_test_names(self):
        return [
            'logic2',
        ]

      def logic2(self):
        pass

    bt_cls = MockBaseTest(self.mock_test_cls_configs)
    bt_cls.run(test_names=['re:logic2'])
```



_Code quote:_
```Python
matching_name_tuples.append((name, getattr(self, name)))
```

xpconanfan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/938)* status: 1 of 4 files reviewed, 8 unresolved discussions (waiting on @eric100lin, @ko1in1u, and @xianyuanjia)

___
*[`mobly/base_test.py` line 1045 at r1](https://reviewable.io/reviews/google/mobly/938#gh-1737543457:-O5WsThn6OrnbrkETm1L:b-896fix) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1045)):*
<details><summary><i>Previously, ko1in1u (Kolin Lu) wrote…</i></summary><blockquote>

Ah, got it. At first glance, it looks like a tuple type from its name.
</blockquote></details>

Done.
___
*[`mobly/base_test.py` line 1047 at r1](https://reviewable.io/reviews/google/mobly/938#-O5W4sQE9rat6wJeA8kL:-O5Wsw9m3ZuhG-cdkctS:b-896fix) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1047)):*
<details><summary><i>Previously, eric100lin (Eric Lin (Tzu Hsiang Lin)) wrote…</i></summary><blockquote>

https://google.github.io/styleguide/pyguide.html#2144-decision
Always use `if foo is None:` (or `is not None`) to check for a `None` value:
```
if re.fullmatch(test_name_regex, name) is not None:
```
</blockquote></details>

Done.

xpconanfan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/938)* status: 1 of 4 files reviewed, 7 unresolved discussions (waiting on @eric100lin, @ko1in1u, and @xianyuanjia)

___
*[`mobly/base_test.py` line 1048 at r1](https://reviewable.io/reviews/google/mobly/938#-O5WHc136tQsM0PBQjnK:-O5WtMlx084qLZRdYOzu:b-7todte) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1048)):*
<details><summary><i>Previously, eric100lin (Eric Lin (Tzu Hsiang Lin)) wrote…</i></summary><blockquote>

I think we should also check `self._assert_valid_test_name(name)` here,
otherwise I can inject invalid tests like this :)
```
    class MockBaseTest(base_test.BaseTestClass):

        # ...

      def get_existing_test_names(self):
        return [
            'logic2',
        ]

      def logic2(self):
        pass

    bt_cls = MockBaseTest(self.mock_test_cls_configs)
    bt_cls.run(test_names=['re:logic2'])
```
</blockquote></details>

well, if you consider the case of overriding `get_existing_test_names`, then you can say the `_assert_valid_test_name` can also be overriden... and we have to draw the line somewhere right..
___
*[`mobly/base_test.py` line 1050 at r1](https://reviewable.io/reviews/google/mobly/938#-O5W5iY00aUgQtkus9US:-O5WtD2w0Hu4-hsw-gkH:b-896fix) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1050)):*
<details><summary><i>Previously, eric100lin (Eric Lin (Tzu Hsiang Lin)) wrote…</i></summary><blockquote>

ditto
</blockquote></details>

Done.
___
*[`tests/mobly/base_test_test.py` line 205 at r1](https://reviewable.io/reviews/google/mobly/938#-O5W9Tl8Bud328YpS15a:-O5WwRtC2AOHRV7zxmJD:b-896fix) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/tests/mobly/base_test_test.py#L205)):*
<details><summary><i>Previously, eric100lin (Eric Lin (Tzu Hsiang Lin)) wrote…</i></summary><blockquote>

Please cover `generate_tests` case, I found the following error:
```
  File "mobly/mobly/base_test.py", line 1026, in _get_test_methods
    regex_matching_methods = self._get_regex_matching_test_methods(
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "mobly/mobly/base_test.py", line 1049, in _get_regex_matching_test_methods
    matching_name_tuples.append((name, getattr(self, name)))
                                       ^^^^^^^^^^^^^^^^^^^
AttributeError: 'MockBaseTest' object has no attribute 'test_something_4'
```
when I add codes like:
```
      def pre_run(self):
        self.generate_tests(
            test_logic=self.logic,
            name_func=self.name_gen,
            arg_sets=[('something', 4)],
        )

      def name_gen(self, a, b):
        return 'test_%s_%s' % (a, b)

      def logic(self, a, b):
        pass
```

You might need to check `hasattr(self, name)` in the first loop.
</blockquote></details>

Done.

eric100lin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/938)* status: 1 of 4 files reviewed, 7 unresolved discussions (waiting on @ko1in1u, @xianyuanjia, and @xpconanfan)

___
*[`mobly/base_test.py` line 1048 at r1](https://reviewable.io/reviews/google/mobly/938#-O5WHc136tQsM0PBQjnK:-O5WzYc06AfEN4DJQNwO:b-q8t7t3) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1048)):*

well, if you consider the case of overriding `get_existing_test_names`, then you can say the `_assert_valid_test_name` can also be overriden... and we have to draw the line somewhere right..
</blockquote></details>

But `get_existing_test_names` is not start with _ prefix, it means open for users to overwrite it.
It's nice to have that check.

eric100lin (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/938)* status: 1 of 4 files reviewed, 6 unresolved discussions (waiting on @ko1in1u, @xianyuanjia, and @xpconanfan)

___
*[`mobly/base_test.py` line 1048 at r1](https://reviewable.io/reviews/google/mobly/938#-O5WHc136tQsM0PBQjnK:-O5X0Jqz8AnYBWKbVVBY:bv41y2e) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1048)):*
<details><summary><i>Previously, eric100lin (Eric Lin (Tzu Hsiang Lin)) wrote…</i></summary><blockquote>

But `get_existing_test_names` is not start with _ prefix, it means open for users to overwrite it.
It's nice to have that check.
</blockquote></details>

I see you update the logic for this path, marked as resolved.

eric100lin (approved):
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/mobly/938)* status: all files reviewed, 3 unresolved discussions (waiting on @ko1in1u, @xianyuanjia, and @xpconanfan)

xpconanfan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/938)* status: all files reviewed, 3 unresolved discussions (waiting on @ko1in1u and @xianyuanjia)

___
*[`mobly/base_test.py` line 1052 at r1](https://reviewable.io/reviews/google/mobly/938#-O5VpOSB3sPGdc9xE0Vk:-O5X2USG1YR1l8GhsltR:b-896fix) ([raw file](https://github.com/google/mobly/blob/bd0d282ff20742920cf1784f051bbd4084d33c1d/mobly/base_test.py#L1052)):*

`current_valid_names` contains both the predefined and generated test names.

So if you loop through `current_valid_names`, you can get generated tests, which would fail on `getattr`.

This is probably something that should also be verified in a unit test.
</blockquote></details>

Done.

xianyuanjia (approved):
:lgtm:

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/mobly/938#-:-O5_HU_C3VfpsoCRdnDS:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @ko1in1u)

xianyuanjia (approved):
> Dismissed @ko1in1u from a discussion.
> *[Reviewable](https://reviewable.io/reviews/google/mobly/938)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @xpconanfan)

PR #947, opened 2024-11-27
Support test case selection when using suite class.

mhaoli:
This CL adds a public method `set_test_selecotr` to `base_suite.BaseSuite`. 

This method is for passing test selector from CLI args to suite class.

### DIFF
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -389,4 +389,3 @@ def _parse_raw_test_selector(selected_tests):
       test_class_name_to_tests[test_name] = None
 
   return test_class_name_to_tests
-
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -210,6 +210,8 @@ def run_suite_class(argv=None):
       log_dir=config.log_path, testbed_name=config.testbed_name
   )
   suite = suite_class(runner, config)
+  test_selector = _parse_raw_test_selector(cli_args.tests)
+  suite.set_test_selector(test_selector)
   console_level = logging.DEBUG if cli_args.verbose else logging.INFO
   ok = False
   with runner.mobly_logger(console_level=console_level):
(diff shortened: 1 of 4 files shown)


mobly/suite_runner.py
```diff
@@ -326,6 +328,50 @@ def compute_selected_tests(test_classes, selected_tests):
   # The user is selecting some tests to run. Parse the selectors.
   # Dict from test_name class name to list of tests to execute (or None for all
   # tests).
+  test_class_name_to_tests = _parse_raw_test_selector(selected_tests)
+
+  # Now transform class names to class objects.
```
xianyuanjia, line 333: This doesn't match what is described in the function docstring, where `test_classes` is a list of strings. Looking deeper into the code, the `test_classes` param is indeed a list of class objects, so the docstring should be fixed.

mobly/suite_runner.py
```diff
@@ -326,6 +328,50 @@ def compute_selected_tests(test_classes, selected_tests):
...
+  # Dict from test_name class name to instance.
+  class_name_to_class = {cls.__name__: cls for cls in test_classes}
+  for test_class_name, tests in test_class_name_to_tests.items():
+    test_class = class_name_to_class.get(test_class_name)
+    if not test_class:
+      raise Error('Unknown test_name class %s' % test_class_name)
```
xianyuanjia, line 339: Should be "Unknown test_class name %s"

mobly/suite_runner.py
```diff
@@ -326,6 +328,50 @@ def compute_selected_tests(test_classes, selected_tests):
...
+    class. E.g. the example in `selected_tests` would translate to:
+
+    .. code-block:: python
+      {
+        'FooTest': None,
+        'BarTest': None,
+        'BazTest': ['test_method_a', 'test_method_b'],
+      }
+
+    This returns None if `elected_tests` is None.
```
xianyuanjia, line 371: `selected_tests`

mobly/suite_runner.py
mhaoli: Good catch. DONE.

mobly/suite_runner.py
mhaoli: DONE

xianyuanjia (approved):
:lgtm:

> Reviewed 3 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/mobly/947#-:-OCqb-Qc8cw2JbeEhWZa:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @xpconanfan)

PR #949, opened 2024-12-02
Record suite info in test summary yaml.

mhaoli:
Record the suite class name in summary yaml. No information will be added if the test is triggered by `suite_runner.run_suite`.

To record suite info, this PR adds a new test summary entry type `SuiteInfo` and record class `SuiteInfoRecord`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/949)

### DIFF
--- a/mobly/base_suite.py
+++ b/mobly/base_suite.py
@@ -38,7 +38,7 @@ class BaseSuite(abc.ABC):
     self._config = config.copy()
     self._test_selector = None
     self._suite_run_display_name = self.__class__.__name__
-    self._suite_info = None
+    self._suite_info = {}
 
   @property
   def user_params(self):
@@ -107,7 +107,8 @@ class BaseSuite(abc.ABC):
   def set_suite_run_display_name(self, suite_run_display_name):
     """Interface for sub-classes to set a customized display name.
 
-    If not set, suite class name will be used by default.
+    This name provides run-specific context intended for display. Default to
+    suite class name. Set this in sub-classes to include run-specific context.
 
     Args:
       suite_run_display_name: str, the display name to set.
--- a/mobly/base_suite.py
+++ b/mobly/base_suite.py
@@ -37,6 +37,8 @@ class BaseSuite(abc.ABC):
     self._runner = runner
     self._config = config.copy()
     self._test_selector = None
+    self._suite_run_display_name = self.__class__.__name__
+    self._suite_info = None
 
   @property
   def user_params(self):
@@ -96,37 +98,32 @@ class BaseSuite(abc.ABC):
     """Function used to add post tests cleanup tasks (optional)."""
     pass
 
-  # Optional interfaces that users can override to record customized suite
-  # information to test summary.
+  # Methods for sub-classes to record customized suite information to
+  # test summary.
 
-  def get_suite_name(self):
-    """Override to return a customized suite name (optional).
+  def set_suite_run_display_name(self, suite_run_display_name):
+    """Interface for sub-classes to set a customized display name.
 
-    Use suite class name by default.
+    If not set, suite class name will be used by default.
 
-    Returns:
-      A string that indicates the suite name.
+    Args:
+      suite_run_display_name: str, the display name to set.
     """
-    return self.__class__.__name__
+    self._suite_run_display_name = suite_run_display_name
 
-  def get_run_identifier(self):
-    """Override to record identifier describing the key run context (optional).
+  def get_suite_run_display_name(self):
+    """Returns the suite run display name."""
+    return self._suite_run_display_name
 
-    Users can include test runtime info as this method will be called after all
-    test classes are executed.
+  def set_suite_info(self, suite_info=None):
+    """Interface for sub-classes to set user defined extra info to test summary.
 
-    Returns:
-      A string that indicates key run context information.
+    Args:
+      suite_info: dict, A dict of suite information. Keys and values must be
+        serializable.
     """
-    return None
+    self._suite_info = suite_info or {}
 
   def get_suite_info(self):
-    """Override to record user defined extra info to test summary (optional).
-
-    Users can include test runtime info as this method will be called after all
-    test classes are executed.
-
-    Returns:
-      A dict of suite information. Keys and values must be serializable.
-    """
-    return {}
+    """Returns suite information."""
+    return self._suite_info
--- a/mobly/base_suite.py
+++ b/mobly/base_suite.py
@@ -50,6 +50,18 @@ class BaseSuite(abc.ABC):
     """
     self._test_selector = test_selector
 
+  def get_suite_name(self):
+    """The name of this suite.
+
+    By default, use the name of the test class. User can overwrite to return
+    a customized suite name. User can include test runtime info as this will be
+    collected after all test classes are executed.
+
+    Returns:
+      A string of suite name.
+    """
+    return self.__class__.__name__
+
   def get_suite_info(self):
     """User defined extra suite information to be recorded in test summary.
 
--- a/mobly/base_suite.py
+++ b/mobly/base_suite.py
@@ -29,7 +29,13 @@ class BaseSuite(abc.ABC):
   Users can use this class if they need to define their own setup and teardown
   steps on the suite level. Otherwise, just use suite_runner.run_suite on the
   list of test classes.
+
+  Attributes:
+    VERSION: str, the test suite version string. Default to None. Subclasses may
+      overwrite this attribute to record suite version in test summary.
   """
+
+  VERSION = None
 
   def __init__(self, runner, config):
     self._runner = runner
--- a/mobly/base_suite.py
+++ b/mobly/base_suite.py
@@ -96,7 +96,8 @@ class BaseSuite(abc.ABC):
     """Function used to add post tests cleanup tasks (optional)."""
     pass
 
-  # Optional interfaces to record user defined suite information to test summary
+  # Optional interfaces that users can override to record customized suite
+  # information to test summary.
 
   def get_suite_name(self):
     """Override to return a customized suite name (optional).
--- a/mobly/base_suite.py
+++ b/mobly/base_suite.py
@@ -36,13 +36,18 @@ class BaseSuite(abc.ABC):
     self._config = config.copy()
 
   @property
-  def suite_info(self):
-    """Returns user defined suite info that will be recorded to test summary."""
-    return {}
-
-  @property
   def user_params(self):
     return self._config.user_params
+
+  def get_suite_info(self):
+    """User defined extra suite information to be recorded in test summary.
+
+    This method will be called after all test classes are executed.
+
+    Returns:
+      A dict of suite information. Keys and values must be serializable.
+    """
+    return {}
 
   def add_test_class(self, clazz, config=None, tests=None, name_suffix=None):
     """Adds a test class to the suite.
--- a/mobly/base_suite.py
+++ b/mobly/base_suite.py
@@ -71,12 +71,15 @@ class BaseSuite(abc.ABC):
     """
     if self._test_selector:
       cls_name = clazz.__name__
-      if cls_name not in self._test_selector:
+      if (cls_name, name_suffix) in self._test_selector:
+        tests = self._test_selector[(cls_name, name_suffix)]
+      elif cls_name in self._test_selector:
+        tests = self._test_selector[cls_name]
+      else:
         logging.info(
             'Skipping test class %s due to CLI argument `tests`.', cls_name
         )
         return
-      tests = self._test_selector[cls_name]
 
     if not config:
       config = self._config
--- a/mobly/base_suite.py
+++ b/mobly/base_suite.py
@@ -14,6 +14,8 @@
 
 import abc
 
+import logging
+
 
 class BaseSuite(abc.ABC):
   """Class used to define a Mobly suite.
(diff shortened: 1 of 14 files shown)


xianyuanjia (changes requested):
How would this impact any downstream parsing of the results? Would the new entries be ignored by default?

> *[Reviewable](https://reviewable.io/reviews/google/mobly/949#-:-OD8_UVj92geTydgz9bZ:brfgpl3)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @mhaoli and @xpconanfan)

___
*[`mobly/records.py` line 734 at r1](https://reviewable.io/reviews/google/mobly/949#-OD8agfA6sybs3z8Gdsy:-OD8agfA6sybs3z8Gdsz:b-huru15) ([raw file](https://github.com/google/mobly/blob/fd0c097a297bf132bae107a2dc8f6a925cfbb9ae/mobly/records.py#L734)):*
> ```Python
>   """A record representing the suite info in test summary."""
> 
>   KEY_SUITE_CLASS_NAME = 'Suite Class Name'
> ```

Could simplify to Suite Name. 



_Code quote:_
```Python
KEY_SUITE_CLASS_NAME
```
___
*[`mobly/records.py` line 739 at r1](https://reviewable.io/reviews/google/mobly/949#-OD8_tl52yaAzLVu57Hy:-OD8_tl52yaAzLVu57Hz:b-vpg9ca) ([raw file](https://github.com/google/mobly/blob/fd0c097a297bf132bae107a2dc8f6a925cfbb9ae/mobly/records.py#L739)):*
> ```Python
>   def __init__(self, suite_class_name):
>     self.suite_class_name = suite_class_name
>     self.timestamp = time.time()
> ```

Would it be better to record a begin and end time instead? So we can report the total execution time of a suite.



_Code quote:_
```Python
    self.timestamp = time.time()
```

mobly/records.py
```diff
@@ -729,19 +729,34 @@ def summary_dict(self):
...
+  Attributes:
+    suite_class: str, the class name of the test suite class.
+    begin_time: int, epoch timestamp of when the suite started.
+    end_time: int, epoch timestamp of when the suite ended.
+  """
 
-  def __init__(self, suite_class_name):
-    self.suite_class_name = suite_class_name
-    self.timestamp = time.time()
+  KEY_SUITE_CLASS = 'Suite Class'
```
xpconanfan, line 740: what's "suite class"?

mhaoli (commented):
I did some verification and think this CL should be safe. I'd love to know if you suggest other things to be checked.

I ran a mobly_mh_test successfully, consulted Yao about the result parsing logic, did code search to find possible breakage. I only found that a warning will be printed on MH side, I'll ask MH team about how to fix it. 

> *[Reviewable](https://reviewable.io/reviews/google/mobly/949#-:-ODBXfCb2znMCjpyKHZH:b-e692uj)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @xianyuanjia and @xpconanfan)

___
*[`mobly/records.py` line 734 at r1](https://reviewable.io/reviews/google/mobly/949#-OD8agfA6sybs3z8Gdsy:-ODBXQBZBXC80S4XgOa-:bte0oz4) ([raw file](https://github.com/google/mobly/blob/fd0c097a297bf132bae107a2dc8f6a925cfbb9ae/mobly/records.py#L734)):*

Could simplify to Suite Name.
</blockquote></details>

I'm considering suite name may contain more info than only suite class name. Pinged you to discuss offline.
___
*[`mobly/records.py` line 739 at r1](https://reviewable.io/reviews/google/mobly/949#-OD8_tl52yaAzLVu57Hy:-ODBXNnD2GjPTyXCpkto:bcuycx1) ([raw file](https://github.com/google/mobly/blob/fd0c097a297bf132bae107a2dc8f6a925cfbb9ae/mobly/records.py#L739)):*

Would it be better to record a begin and end time instead? So we can report the total execution time of a suite.
</blockquote></details>

Good idea. DONE.

xpconanfan (changes requested):

mobly/records.py
mhaoli: The test suite class that inherits from `base_suite.BaseSuite`. 

If "Suite Class" is not clear enough, I modified it to "Test Suite Class".

xpconanfan:
Do you really need a new class for a string and two timestamps?

Also, why adding a field to the test report when most situations don't
involve suite?

Could you do this without modifying the records module?

xpconanfan:
Since the suite info is only created by suite runner, would it make sense
to have the suite runner to simply append a section in the results yaml at
the end?

Also, would it make sense to summarize other info in the suite info? Like
suite version, total pass/fail

mobly/suite_runner.py
```diff
@@ -77,12 +77,60 @@ def setup_suite(self, config):
...
+  """A record representing the test suite info in test summary.
+
+  Attributes:
+    test_suite_class: str, the class name of the test suite class.
+    test_suite_version: str, the test suite version.
+    begin_time: int, epoch timestamp of when the suite started.
+    end_time: int, epoch timestamp of when the suite ended.
+  """
+
+  KEY_TEST_SUITE_CLASS = 'Test Suite Class'
```
xpconanfan, line 97: still not sure what this means

mobly/suite_runner.py
```diff
@@ -77,12 +77,60 @@ def setup_suite(self, config):
...
+  KEY_TEST_SUITE_VERSION = 'Test Suite Version'
```
xpconanfan, line 98: Instead of hard coding fields one by one, should we provide a mechanism for users to dump their own info as needed.
Like in the test case record, the essential fields are hard coded, then there's the user defined info field

xpconanfan (changes requested):

mobly/suite_runner.py
mhaoli: Good point. Added a property `BaseSuite.suite_info` for user defined suite info.

mobly/suite_runner.py
mhaoli: It's the class name of test suite class. Do you mean that `Test Suite Class` is confusing and I'd better use `Test Suite Class Name`?

I didn't use `Test Suite Class Name` just because `records.py` is using `Test Class` for test class name.

mhaoli (commented):
Resolved comments in this thread.

> *[Reviewable](https://reviewable.io/reviews/google/mobly/949#-:-OEZUpocAdgey3qzrO4-:b-ytuuto)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @xianyuanjia and @xpconanfan)

___
*[`mobly/records.py` line 734 at r1](https://reviewable.io/reviews/google/mobly/949#-OD8agfA6sybs3z8Gdsy:-OEZUIaMB6G7BrnoTai4:b-6wzzth) ([raw file](https://github.com/google/mobly/blob/fd0c097a297bf132bae107a2dc8f6a925cfbb9ae/mobly/records.py#L734)):*

I'm considering suite name may contain more info than only suite class name. Pinged you to discuss offline.
</blockquote></details>

Added a field 'Suite Name' which is either the suite class name or a user defined suite name. But I didn't remove field `Test Suite Class` as I think we need to keep this info.

mobly/suite_runner.py
xianyuanjia: My preference is still either Suite Class or Suite Class Name. Having `Test` here doesn't really make it clearer, as "suite" in Mobly already unambiguously means "test suite". If the clarification needed is to show that this is the literal Python class name of the suite in question, then maybe Suite Python Class?

xianyuanjia (changes requested):
> Reviewed 1 of 5 files at r4, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/mobly/949)* status: 4 of 5 files reviewed, 7 unresolved discussions (waiting on @mhaoli and @xpconanfan)

___
*[`mobly/base_suite.py` line 56 at r6](https://reviewable.io/reviews/google/mobly/949#-OEoW5Iq8IJysXzgVGsp:-OEoW5Iq8IJysXzgVGsq:b-ughu30) ([raw file](https://github.com/google/mobly/blob/929cfe808b48aca4ebacad86b4cae8d9323a4338/mobly/base_suite.py#L56)):*
> ```Python
>     """The name of this suite.
> 
>     By default, use the name of the test class. User can overwrite to return
> ```

suite class?



_Code quote:_
```Python
test class
```
___
*[`mobly/base_suite.py` line 73 at r6](https://reviewable.io/reviews/google/mobly/949#-OEoWBka4tSm_wvIhJTh:-OEoWBka4tSm_wvIhJTi:baz0yxj) ([raw file](https://github.com/google/mobly/blob/929cfe808b48aca4ebacad86b4cae8d9323a4338/mobly/base_suite.py#L73)):*
> ```Python
>       A dict of suite information. Keys and values must be serializable.
>     """
>     return {}
> ```

There is no implementation here. Is this intentional? If so, please leave a TODO.



_Code quote:_
```Python
return {}
```
___
*[`mobly/suite_runner.py` line 114 at r6](https://reviewable.io/reviews/google/mobly/949#-OEp4A5J-B5WisWJ_Dp4:-OEp4A5K9uz7PWiaaT57:b-axsu9c) ([raw file](https://github.com/google/mobly/blob/929cfe808b48aca4ebacad86b4cae8d9323a4338/mobly/suite_runner.py#L114)):*
> ```Python
>   _end_time: int | None
> 
>   def __init__(self, test_suite_class):
> ```

Is there anywhere that indicates clearly that this record only works for suite classes and not suites directly assembled via `run_suite`?

mobly/suite_runner.py
mhaoli: Thanks, I'd prefer "Suite Class Name".

mhaoli (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/949)* status: 1 of 5 files reviewed, 7 unresolved discussions (waiting on @xianyuanjia and @xpconanfan)

___
*[`mobly/base_suite.py` line 56 at r6](https://reviewable.io/reviews/google/mobly/949#-OEoW5Iq8IJysXzgVGsp:-OErVOXx8shYoiPps1XG:boigme5) ([raw file](https://github.com/google/mobly/blob/929cfe808b48aca4ebacad86b4cae8d9323a4338/mobly/base_suite.py#L56)):*

suite class?
</blockquote></details>

Good catch. DONE.
___
*[`mobly/base_suite.py` line 73 at r6](https://reviewable.io/reviews/google/mobly/949#-OEoWBka4tSm_wvIhJTh:-OErVTPqDVMMLTfxq8V-:b7d7g2q) ([raw file](https://github.com/google/mobly/blob/929cfe808b48aca4ebacad86b4cae8d9323a4338/mobly/base_suite.py#L73)):*

There is no implementation here. Is this intentional? If so, please leave a TODO.
</blockquote></details>

This is intentional. This method is for users to override to record customized info to test summary, e.g. suite version. By default, it just returns an empty dict.

Modified docstring to clarify this.
___
*[`mobly/suite_runner.py` line 114 at r6](https://reviewable.io/reviews/google/mobly/949#-OEp4A5J-B5WisWJ_Dp4:-OErZl1dEsX18YWRP1zl:b10k04) ([raw file](https://github.com/google/mobly/blob/929cfe808b48aca4ebacad86b4cae8d9323a4338/mobly/suite_runner.py#L114)):*

Is there anywhere that indicates clearly that this record only works for suite classes and not suites directly assembled via `run_suite`?
</blockquote></details>

Good point, clarified this point in docstring of this class.

xianyuanjia (approved):
:lgtm:

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/mobly/949#-:-OF35lLXEaCQaPrSj1sB:bnfp4nl)* status: 1 of 5 files reviewed, 4 unresolved discussions (waiting on @xpconanfan)

mobly/base_suite.py
```diff
@@ -95,3 +95,38 @@ def setup_suite(self, config):
   def teardown_suite(self):
     """Function used to add post tests cleanup tasks (optional)."""
     pass
+
+  # Optional interfaces that users can override to record customized suite
+  # information to test summary.
+
+  def get_suite_name(self):
```
xpconanfan, line 102: shouldn't this just be part of the suite info?
why should it be a separate getter method?

mobly/base_suite.py
```diff
@@ -95,3 +95,38 @@ def setup_suite(self, config):
...
+    """Override to return a customized suite name (optional).
+
+    Use suite class name by default.
+
+    Returns:
+      A string that indicates the suite name.
+    """
+    return self.__class__.__name__
+
+  def get_run_identifier(self):
```
xpconanfan, line 112: what is "run identifier"?

mobly/base_suite.py
mhaoli: Test suites can return any custom values in `suite info`, our infra won't parse it and just write it to summary file.

So I created separate getter methods for fields that our infra needs to parse.

mobly/base_suite.py
mhaoli: This is for key run context that will be displayed as part of the title in our result viewer.

Through this we can include info like the phone brand and phone model.

mhaoli:
**IIRC, Ang's main concerns previously were**:

1. The approach of having subclasses override `get_xxx` methods inherited from `BaseSuite` is very confusing. We should provide `set_xxx` methods in BaseSuite for subclasses to call.
2. The name `run_identifier` is confusing. It's unclear what it's for. Can the field `suite_name` contain all information?

**So the latest commit does the following changes**:

1. For point 1, fixed.
2. For point 2, I merged the `suite_name` and `run_identifier` fields, but the new field has been renamed to `suite_run_display_name`. It was renamed because I don't think `suite_name` field should contain runtime info.

xpconanfan (approved):

PR #952, opened 2025-01-16
Add timeout support for FastbootProxy

otakusaikou:
TEST=unit tests with tox

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/952)

### DIFF
--- a/mobly/controllers/android_device_lib/fastboot.py
+++ b/mobly/controllers/android_device_lib/fastboot.py
@@ -13,40 +13,9 @@
 # limitations under the License.
 
 import logging
-from subprocess import Popen, PIPE
+from subprocess import PIPE
 
 from mobly import utils
-
-
-def exe_cmd(*cmds):
-  """Executes commands in a new shell. Directing stderr to PIPE.
-
-  This is fastboot's own exe_cmd because of its peculiar way of writing
-  non-error info to stderr.
-
-  Args:
-    cmds: A sequence of commands and arguments.
-
-  Returns:
-    The output of the command run.
-
-  Raises:
-    Exception: An error occurred during the command execution.
-  """
-  cmd = ' '.join(cmds)
-  proc = Popen(cmd, stdout=PIPE, stderr=PIPE, shell=True)
-  (out, err) = proc.communicate()
-  ret = proc.returncode
-  logging.debug(
-      'cmd: %s, stdout: %s, stderr: %s, ret: %s',
-      utils.cli_cmd_to_string(cmds),
-      out,
-      err,
-      ret,
-  )
-  if not err:
-    return out
-  return err
 
 
 class FastbootProxy:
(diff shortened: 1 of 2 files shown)


otakusaikou:
Hello @xpconanfan, would you please help review this PR? Thanks.

xpconanfan:
Please resolve the conflict if you still wish to proceed with this PR.
Otherwise please close this

xpconanfan:
Please do not rename or move the function into the class.
Let's just focus on adding a timeout option

otakusaikou:
conflict resolved and keep the original exec function name

mobly/controllers/android_device_lib/fastboot.py
```diff
@@ -61,19 +67,24 @@ class FastbootProxy:
   >> fb.devices() # will return the console output of "fastboot devices".
   """
 
-  def __init__(self, serial=''):
+  def __init__(self, serial='', timeout=1200):
```
xpconanfan, line 70: `timeout` doesn't need to be a constructor value.

See how `adb` handles this and follow the same pattern.

otakusaikou: updated by following the recommendation

mobly/controllers/android_device_lib/fastboot.py
```diff
@@ -13,33 +13,42 @@
 # limitations under the License.
 
 import logging
-from subprocess import Popen, PIPE
+from subprocess import PIPE
 
 from mobly import utils
 
+# The default fastboot command timeout settings.
+DEFAULT_TIMEOUT_SEC = 1200
```
xpconanfan, line 21: Why does the default need to be so long?

otakusaikou: From my experience, as there wasn't any "wait for fastboot" function, so we used to just run fastboot command, including "waiting device back from other state", such as:
- adb reboot bootloader -> run fastboot command -> 20-30s -> command run
- device bootloader crash and bootloader ab fallback triggered -> 3-6 mins (boot and retry) -> fallback to other slot -> command run

Hence I would suggest not to set this too short (as it was previously "infinite"), if set this to sth like 10s, I think we may break many existing test scripts. (10 ~ 20mins looks suitable, 20 mins was the one that I've been using internally and didn't see issues)

xpconanfan: But most fastboot cmds are short?
We should set a value reasonable for most cmds so we don't waste time when the operation is stuck for real.
20min is way too long.

For cmds that require more time, the caller should explicitly extend the timeout. We could do a large scale change for internal scripts.

otakusaikou: updated to 3 mins

xpconanfan:
Please rebase the branch

otakusaikou (commented):
Branch updated

xpconanfan (approved):

xpconanfan:
CI is failing, pls fix

xpconanfan (changes requested):
Fix CI

otakusaikou:
Fixed the format check issue by running pyink tool locally to reformat fastboot_test.py.

xpconanfan (approved):

PR #956, opened 2025-03-12
Fix suite_runner.run_suite example in docs

sbrocket:
suite_runner.run_suite expects a list in the first argument, not a variable number of arguments.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/956)


mhaoli (approved):

mhaoli:
Thanks for the fix!

PR #957, opened 2025-03-17
Check whether adb forwarding takes effect when hit socket connection failure

mhaoli:
Check whether adb forwarding takes effect when hit socket connection failure.

Tested E2E locally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/957)

### DIFF
--- a/mobly/controllers/android_device_lib/snippet_client_v2.py
+++ b/mobly/controllers/android_device_lib/snippet_client_v2.py
@@ -457,10 +457,16 @@ class SnippetClientV2(client_base.ClientBase):
         if not stdout:
           raise errors.Error(
               self._device,
-              'The Adb forward command execution does not take effect. Please'
-              ' check if there are other processes occupying adb forward on the'
+              'The Adb forward command execution did not take effect. Please'
+              ' check if there are other processes affecting adb forward on the'
               ' host.',
           ) from err2
+
+        raise errors.Error(
+            self._device,
+            'Failed to establish socket connection from host to snippet server'
+            ' running on Android device.'
+        ) from err2
 
     self._conn.settimeout(_SOCKET_READ_TIMEOUT)
     self._client = self._conn.makefile(mode='brw')
--- a/mobly/controllers/android_device_lib/snippet_client_v2.py
+++ b/mobly/controllers/android_device_lib/snippet_client_v2.py
@@ -446,9 +446,21 @@ class SnippetClientV2(client_base.ClientBase):
       self.log.debug(
           'Failed to connect to localhost, trying 127.0.0.1: %s', str(err)
       )
-      self._conn = socket.create_connection(
-          ('127.0.0.1', self.host_port), _SOCKET_CONNECTION_TIMEOUT
-      )
+      try:
+        self._conn = socket.create_connection(
+            ('127.0.0.1', self.host_port), _SOCKET_CONNECTION_TIMEOUT
+        )
+      except ConnectionRefusedError as err2:
+        _, stdout, _ = utils.run_command(
+            f'netstat tulpn | grep ":{self.host_port}"', shell=True
+        )
+        if not stdout:
+          raise errors.Error(
+              self._device,
+              'The Adb forward command execution does not take effect. Please'
+              ' check if there are other processes occupying adb forward on the'
+              ' host.'
+          ) from err2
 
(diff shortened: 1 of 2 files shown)


xianyuanjia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/957)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @xpconanfan)

___
*[`mobly/controllers/android_device_lib/snippet_client_v2.py` line 457 at r1](https://reviewable.io/reviews/google/mobly/957#-OL_ZXqe2VS22eP4sdu0:-OL_ZXqe2VS22eP4sdu1:bxozujq) ([raw file](https://github.com/google/mobly/blob/727670944995b6fabfeebe61c05c2ce027ca2ed3/mobly/controllers/android_device_lib/snippet_client_v2.py#L457)):*
> ```Python
>             f'netstat tulpn | grep ":{self.host_port}"', shell=True
>         )
>         if not stdout:
> ```

What happens when there is a stdout? Seems like it would just fail silently and run into attribute error in the next line.
___
*[`mobly/controllers/android_device_lib/snippet_client_v2.py` line 460 at r1](https://reviewable.io/reviews/google/mobly/957#-OL_ZtkC43dqyPmHPRrC:-OL_ZtkC43dqyPmHPRrD:b2ni15m) ([raw file](https://github.com/google/mobly/blob/727670944995b6fabfeebe61c05c2ce027ca2ed3/mobly/controllers/android_device_lib/snippet_client_v2.py#L460)):*
> ```Python
>           raise errors.Error(
>               self._device,
>               'The Adb forward command execution does not take effect. Please'
> ```

did



_Code quote:_
```Python
does
```

mhaoli (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/957)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @xianyuanjia and @xpconanfan)

___
*[`mobly/controllers/android_device_lib/snippet_client_v2.py` line 457 at r1](https://reviewable.io/reviews/google/mobly/957#-OL_ZXqe2VS22eP4sdu0:-OLdc9nk1kHhyfvZ7POR:b-v0s0uu) ([raw file](https://github.com/google/mobly/blob/727670944995b6fabfeebe61c05c2ce027ca2ed3/mobly/controllers/android_device_lib/snippet_client_v2.py#L457)):*

What happens when there is a stdout? Seems like it would just fail silently and run into attribute error in the next line.
</blockquote></details>

Opps, good catch. Added the logic to raise an exception with default error message and unit test for it.
___
*[`mobly/controllers/android_device_lib/snippet_client_v2.py` line 460 at r1](https://reviewable.io/reviews/google/mobly/957#-OL_ZtkC43dqyPmHPRrC:-OLdc9MeC1vApjIXDEwy:b-896fix) ([raw file](https://github.com/google/mobly/blob/727670944995b6fabfeebe61c05c2ce027ca2ed3/mobly/controllers/android_device_lib/snippet_client_v2.py#L460)):*

did
</blockquote></details>

Done.

mobly/controllers/android_device_lib/snippet_client_v2.py
```diff
@@ -446,9 +446,27 @@ def create_socket_connection(self):
...
-      )
+      try:
+        self._conn = socket.create_connection(
+            ('127.0.0.1', self.host_port), _SOCKET_CONNECTION_TIMEOUT
+        )
+      except ConnectionRefusedError as err2:
+        _, stdout, _ = utils.run_command(
+            f'netstat tulpn | grep ":{self.host_port}"', shell=True
+        )
+        if not stdout:
```
xpconanfan, line 457: why is the check `not stdout` instead of checking `stderr`?

mhaoli: `stderr` is not empty when the execution succeds. The is because it is not running with `sudo` so `stderr` contains permission warning.

Thus I'm checking `stdout`.

mhaoli: Right, using the ret code would be slightly better. Modified.

xianyuanjia (approved):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/957)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @mhaoli and @xpconanfan)

PR #970, opened 2025-05-22
android_device: update take_screenshot to support multiple displays

bbrother:
new all_displays arg to take_screenshot allows capturing a screenshot on all connected displays including internal

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/970)

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -1053,22 +1053,52 @@ class AndroidDevice:
     self.log.debug('Bugreport taken at %s.', full_out_path)
     return full_out_path
 
-  def take_screenshot(self, destination, prefix='screenshot'):
+  def take_screenshot(
+      self, destination, prefix='screenshot', all_displays=False
+  ):
     """Takes a screenshot of the device.
 
     Args:
       destination: string, full path to the directory to save in.
       prefix: string, prefix file name of the screenshot.
+      all_displays: bool, if true will take a screenshot on all connnected
+        displays, if false will take a screenshot on the default display.
 
     Returns:
-      string, full path to the screenshot file on the host.
+      string, full path to the screenshot file on the host, or
+      list[str], when all_displays is True, the full paths to the screenshot
+        files on the host.
     """
     filename = self.generate_filename(prefix, extension_name='png')
+    filename_no_extension, _ = os.path.splitext(filename)
     device_path = os.path.join('/storage/emulated/0/', filename)
     self.adb.shell(
-        ['screencap', '-p', device_path], timeout=TAKE_SCREENSHOT_TIMEOUT_SECOND
+        ['screencap', '-p', '-a' if all_displays else '', device_path],
+        timeout=TAKE_SCREENSHOT_TIMEOUT_SECOND,
     )
     utils.create_dir(destination)
+    if all_displays:
+      pic_paths = []
+      png_files = [device_path]
+      # iterate over all files that match the filename, if all_displays is true
+      # then filename will get a suffix of display number eg filenmame.png ->
+      # filename_0.png, filename_1.png
+      png_files = (
+          self.adb.shell('ls /storage/emulated/0/*.png')
+          .decode('utf-8')
+          .split('\n')
+      )
+      for device_path in png_files:
+        if device_path.find(filename_no_extension) < 0:
+          continue
+        self.adb.pull([device_path, destination])
+        pic_paths.append(
+            os.path.join(destination, os.path.basename(device_path))
+        )
+        self.log.debug('Screenshot taken, saved on the host: %s', pic_paths[-1])
+        self.adb.shell(['rm', device_path])
+      return pic_paths
+    # handle single screenshot when all_displays=False
     self.adb.pull([device_path, destination])
     pic_path = os.path.join(destination, filename)
     self.log.debug('Screenshot taken, saved on the host: %s', pic_path)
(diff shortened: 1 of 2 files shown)


mobly/controllers/android_device.py
```diff
@@ -1053,26 +1053,39 @@ def take_bug_report(
     self.log.debug('Bugreport taken at %s.', full_out_path)
     return full_out_path
 
-  def take_screenshot(self, destination, prefix='screenshot'):
+  def take_screenshot(self, destination, prefix='screenshot', all_displays=False):
```
xpconanfan, line 1056: should we make it take all displays by default?
since most devices have one screen only, this wouldn't affect the current usage?

xpconanfan (changes requested):
Need to add unit tests to cover the new scenarios

mobly/controllers/android_device.py
bbrother: I think its best to leave it as is.  In adding tests, I had it return all the images generated when all_displays=True for consistency but that would break existing uses.

ko1in1u (approved):

mobly/controllers/android_device.py
```diff
@@ -1053,27 +1053,44 @@ def take_bug_report(
...
-    self.log.debug('Screenshot taken, saved on the host: %s', pic_path)
-    self.adb.shell(['rm', device_path])
-    return pic_path
+    pic_paths = []
+    png_files = [device_path]
+    # iterate over all files that match the filename, if all_displays is true
+    # then filename will get a suffix of display number eg filenmame.png ->
+    # filename_0.png, filename_1.png
+    if all_displays:
+      png_files = self.adb.shell('ls /storage/emulated/0/*.png').decode('utf-8').split('\n')
```
xpconanfan, line 1083: if someone else called the cmd earlier and didn't remove the files, you may pick up irrelevant pics?

xpconanfan (changes requested):

ko1in1u (approved):

tests/mobly/controllers/android_device_test.py
```diff
@@ -37,6 +37,26 @@
 MockSnippetClient.package = MOCK_SNIPPET_PACKAGE_NAME
 
 
+class _MockAdbProxy(mock_android_device.MockAdbProxy):
```
xpconanfan, line 40: if this is specific to one test, name it accordingly

tests/mobly/controllers/android_device_test.py
```diff
@@ -37,6 +37,26 @@
...
+    """Initializes the instance of _MockAdbProxy."""
+    super().__init__(*args, **kwargs)
+
+  def shell(self, *args, **kwargs):
+    """Mock `shell` of mobly.controllers.android_device_lib.adb.AdbProxy."""
+    # Record all the call args
+    self.mock_shell_func(*args, **kwargs)
+
+    # Handle the server stop command properly
+    if f'ls /storage/emulated/0/*.png' in args:
```
xpconanfan, line 53: this causes the data expected in a unit test to be split from the test itself

it's confusing for others to read and debug, and it's hard to scale to more cases, e.g. we should have a case where the dir contains more leftover screenshots from other calls.

Restructure so the data sits inside the test that uses it.

xpconanfan (changes requested):

tests/mobly/controllers/android_device_test.py
bbrother: Updated with cleaner implementation that is in the function

tests/mobly/controllers/android_device_test.py
bbrother: Done, with additional test case suggested here.

ko1in1u (dismissed):

mobly/controllers/android_device.py
```diff
@@ -1053,27 +1053,50 @@ def take_bug_report(
...
 
     Args:
       destination: string, full path to the directory to save in.
       prefix: string, prefix file name of the screenshot.
+      all_displays: bool, if true will take a screenshot on all connnected
+        displays, if false will take a screenshot on the default display.
 
     Returns:
-      string, full path to the screenshot file on the host.
+      string, full path to the screenshot file on the host, or list of full
```
xpconanfan, line 1068: update the type hint in the beginning to avoid confusion.

now it can return a string or a list of strings.

maybe make a new line that starts with "list of strings"?

mobly/controllers/android_device.py
```diff
@@ -1053,27 +1053,50 @@ def take_bug_report(
...
+          .split('\n')
+      )
+    for device_path in png_files:
+      if device_path.find(filename_no_extension) < 0:
+        continue
+      self.adb.pull([device_path, destination])
+      pic_paths.append(os.path.join(destination, os.path.basename(device_path)))
+      self.log.debug('Screenshot taken, saved on the host: %s', pic_paths[-1])
+      self.adb.shell(['rm', device_path])
+    if all_displays:
```
xpconanfan, line 1097: can we consolidate the branch of `all_displays` into one `if` to simplify the logic?

can keep the old logic branch as is, then have a new branch behind `if` for the new logic.

easier to debug this way

xpconanfan (changes requested):

mobly/controllers/android_device.py
bbrother: Done

xpconanfan (changes requested):

tests/mobly/controllers/android_device_test.py
```diff
@@ -1159,6 +1159,104 @@ def test_AndroidDevice_take_screenshot(
         ),
     )
```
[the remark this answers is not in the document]
bbrother: Done

PR #974, opened 2025-06-02
Introduce Async functions to mobly.

HowardChung1209:
The goal is to facilitate input event tracking, using getevent. It needs to be done asynchronously to enable long running waits. This includes a new function `run_command_async`, that uses asyncio to execute the command, and a new monitor class for Input Events that makes use of it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/974)

### DIFF
--- a/mobly/controllers/android_device.py
+++ b/mobly/controllers/android_device.py
@@ -513,6 +513,8 @@ class AndroidDevice:
     adb_logcat_file_path: A string that's the full path to the adb logcat
       file collected, if any.
     adb: An AdbProxy object used for interacting with the device via adb.
+    adb_async: An AdbAsyncProxy object used for interacting with the device via
+      adb asynchronously.
     fastboot: A FastbootProxy object used for interacting with the device
       via fastboot.
     services: ServiceManager, the manager of long-running services on the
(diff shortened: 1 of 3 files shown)


HowardChung1209:
@xpconanfan could you please review this PR? Thanks!

xpconanfan:
We do not want two ways of doing the same thing which causes confusions and is error prone.

Need to discuss this carefully beforehand.

HowardChung1209:
Hi xpconanfan,

Thanks for the feedback. I understand the concern about having two similar approaches. My initial thought for the separate adb proxy was to handle all adb commands (e.g., shell, instruments, am) in an asynchronous manner. However, our primary need is specifically for running adb shell commands asynchronously. This is particularly important for commands like getevent, which run forever and allow callers to continuously monitor their output.

Would you be open to integrating an async_shell method into the existing AdbProxy class to address this? Or do you think a design document would be a better first step for discussing this change more thoroughly?

Thanks,
Howard

HowardChung1209:
If no other team/users would like this features, we are OK to move this logic to our own test base.

PR #976, opened 2025-06-18
Improve controller error message when DUT isn't found

mark-yacoub:
Improve the error messaging to indicate that an expected DUT might not be connected via ADB.

Test: atest UCD with ADB disconnected

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/976)

### DIFF
--- a/mobly/controller_manager.py
+++ b/mobly/controller_manager.py
@@ -141,7 +141,8 @@ class ControllerManager:
     if actual_number < min_number:
       module.destroy(objects)
       raise signals.ControllerError(
-          'Expected to get at least %d controller objects, got %d.'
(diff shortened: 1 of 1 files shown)


mobly/controller_manager.py
```diff
@@ -141,7 +141,8 @@ def register_controller(self, module, required=True, min_number=1):
     if actual_number < min_number:
       module.destroy(objects)
       raise signals.ControllerError(
-          'Expected to get at least %d controller objects, got %d.'
+          'Expected to get at least %d controller objects, got %d. '
+          'Verify that the DUT is connected via ADB'
```
xpconanfan, line 145: How about "is listed in `adb devices` as `device`"

PR #981, opened 2025-07-04
Add iteration tracking to test execution

shashankdevaraj:
Introduces `_iteration` counter to track the current iteration of a test case, especially useful for repeated or retried tests.
* The `_iteration` variable is incremented before each test execution.
* The `current_test_info` now includes the current iteration number by introducing an `iteration` property to the `RuntimeTestInfo` class.
* Add relevant unit tests for the change.

This allows clients to easily access the current iteration without needing to manage a separate counter.

[unit_test_log.log](https://github.com/user-attachments/files/21051952/unit_test_log.log)

### DIFF
--- a/mobly/base_test.py
+++ b/mobly/base_test.py
@@ -209,6 +209,7 @@ class BaseTestClass:
     self.testbed_name = configs.testbed_name
     self.user_params = configs.user_params
     self.results = records.TestResult()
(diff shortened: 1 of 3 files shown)


xpconanfan:
This breaks encapsulation. Rejecting

PR #982, opened 2025-07-16
AndroidDevice controller should use a timeout for `adb root`

abalawender:
When using a remote proxied device, `adb root` can destabilize the connection and cause ADB to hang. Retrying the command usually works. We should use a timeout instead of allowing the call to hang indefinitely with no feedback to the user.

Test: mobly test using a remote proxied device no longer hangs indefinitely

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly/982)

### DIFF
--- a/mobly/controllers/android_device_lib/adb.py
+++ b/mobly/controllers/android_device_lib/adb.py
@@ -556,7 +556,7 @@ class AdbProxy:
               'Retry the command "%s" since %s.'
               % (
                   utils.cli_cmd_to_string(e.cmd),
-                  ('Error %s occurred' % e.stderr.decode('utf-8').strip())
+                  ('Error "%s" occurred' % e.stderr.decode('utf-8').strip())
                   if isinstance(e, AdbError)
                   else ('it timed out after %d seconds' % e.timeout),
               )
--- a/mobly/controllers/android_device_lib/adb.py
+++ b/mobly/controllers/android_device_lib/adb.py
@@ -31,7 +31,7 @@ ADB_PORT_LOCK = threading.Lock()
 # the timeout of this command.
 ADB_ROOT_RETRY_ATTEMPTS = 3
 ADB_ROOT_RETRY_ATTEMPT_INTERVAL_SEC = 10
-ADB_ROOT_ATTEMPT_TIMEOUT_SEC = 10
+ADB_ROOT_ATTEMPT_TIMEOUT_SEC = 5
 
 # Qualified class name of the default instrumentation test runner.
 DEFAULT_INSTRUMENTATION_RUNNER = (
--- a/mobly/controllers/android_device_lib/adb.py
+++ b/mobly/controllers/android_device_lib/adb.py
@@ -552,13 +552,15 @@ class AdbProxy:
         )
       except (AdbError, AdbTimeoutError) as e:
         if attempt + 1 < ADB_ROOT_RETRY_ATTEMPTS:
+          retry_reason = (
+              f'Error "{e.stderr.decode("utf-8").strip()}" occurred'
+              if isinstance(e, AdbError)
+              else f'it timed out after {e.timeout} seconds'
+          )
           logging.debug(
-              f'Retry the command "{utils.cli_cmd_to_string(e.cmd)}" since '
-              + (
-                  f'Error "{e.stderr.decode("utf-8").strip()}" occurred.'
-                  if isinstance(e, AdbError)
-                  else f'it timed out after {e.timeout} seconds.'
-              )
+              'Retry the command "%s" since %s.',
+              utils.cli_cmd_to_string(e.cmd),
+              retry_reason,
           )
 
           # Buffer between "adb root" commands.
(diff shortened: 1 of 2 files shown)


mobly/controllers/android_device_lib/adb.py
```diff
@@ -537,22 +538,21 @@ def root(self) -> bytes:
...
+      except (AdbError, AdbTimeoutError) as e:
         if attempt + 1 < ADB_ROOT_RETRY_ATTEMPTS:
-          logging.debug(
-              'Retry the command "%s" since Error "%s" occurred.'
-              % (
-                  utils.cli_cmd_to_string(e.cmd),
-                  e.stderr.decode('utf-8').strip(),
-              )
-          )
+          logging.debug('Retry root() since Error "%s" occurred.', e)
```
xpconanfan, line 555: do not change the log msg here

mobly/controllers/android_device_lib/adb.py
```diff
@@ -27,10 +27,11 @@
 # do with port forwarding must happen under this lock.
 ADB_PORT_LOCK = threading.Lock()
 
-# Number of attempts to execute "adb root", and seconds for interval time of
-# this commands.
+# Number of attempts to execute "adb root", and seconds for interval time and
+# the timeout of this command.
 ADB_ROOT_RETRY_ATTEMPTS = 3
 ADB_ROOT_RETRY_ATTEMPT_INTERVAL_SEC = 10
+ADB_ROOT_ATTEMPT_TIMEOUT_SEC = 10
```
xpconanfan, line 34: Is 10s too long?

xpconanfan (changes requested):

abalawender (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/982)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @abalawender and @xpconanfan)

___
*[`mobly/controllers/android_device_lib/adb.py` line 564 at r2](https://reviewable.io/reviews/google/mobly/982#-OVjgqisAJbaE3PiacCr:-OVjh_Zm1LB_PhsyhDXx:b-zdrj2m) ([raw file](https://github.com/google/mobly/blob/a6247d9aa3c13ebb8e13da991b3923f230e268ed/mobly/controllers/android_device_lib/adb.py#L564)):*

I meant the original message was fully capable of capturing the context for any cmd, so you don't need a dedicated message just for reboot?

Please don't add another `except` block
</blockquote></details>

I'm not sure I follow. The original message uses `e.stderr` which exists in `AdbError` but not in `AdbTimeoutError`. I've proposed a minimal change that used the `__str__` method of the two exception classes but you've asked to keep the original message. Would you like to go back to my original approach?

mobly/controllers/android_device_lib/adb.py
abalawender: Thank you for being specific. I take it to understand that you do not wish to have a test for the log format.

The debug logs look like these now:
```
DEBUG    root:adb.py:555 Retry the command "adb root" since it timed out after 5 seconds.
DEBUG    root:adb.py:555 Retry the command "adb root" since Error "adb: unable to connect for root: closed" occurred.
```

The first is identical to the log produced by HEAD:
```
DEBUG    root:adb.py:549 Retry the command "adb root" since Error "adb: unable to connect for root: closed" occurred.
```

abalawender (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly/982)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @abalawender and @xpconanfan)

mobly/controllers/android_device_lib/adb.py
```diff
@@ -537,22 +538,30 @@ def root(self) -> bytes:
...
+            shell=False,
+            timeout=ADB_ROOT_ATTEMPT_TIMEOUT_SEC,
+            stderr=None,
         )
-      except AdbError as e:
+      except (AdbError, AdbTimeoutError) as e:
         if attempt + 1 < ADB_ROOT_RETRY_ATTEMPTS:
           logging.debug(
-              'Retry the command "%s" since Error "%s" occurred.'
+              'Retry the command "%s" since %s.'
```
xpconanfan, line 556: When something don't fit in one line, it's better to use multiple lines for better readability

xpconanfan (changes requested):

tests/mobly/controllers/android_device_lib/adb_test.py
```diff
@@ -853,17 +853,25 @@ def test_root_success_with_retry(self, mock_exec_cmd, mock_sleep):
         adb.AdbError('adb root', '', MOCK_ROOT_ERROR_OUTPUT, 1),
         MOCK_ROOT_SUCCESS_OUTPUT,
     ]
-    output = adb.AdbProxy().root()
+    with self.assertLogs(level='DEBUG') as logs:
+      output = adb.AdbProxy().root()
+    self.assertEqual(
+        logs.output,
+        [
```
xpconanfan, line 860: `[]` is not needed
make the string fit within line lenght

abalawender: [] is needed, logs.output is a list. 

Addressed line length.

xpconanfan (approved):

PR #987, opened 2025-08-20
Refactor: Improve subprocess timeout handling

ko1in1u:
This refactors `run_command` to use `subprocess.Popen.communicate(timeout=...)` for a more robust timeout mechanism.

The old method, relying on a custom `threading.Timer`, could fail to terminate stubborn processes like `fastboot oem dmesg` which ignore `SIGTERM`. The new approach uses `communicate()`, which sends a forceful `SIGKILL` on timeout, guaranteeing the process is terminated.

This change also updates the code to use the modern `text` argument introduced in Python 3.7 to replace `universal_newlines`.

### DIFF
--- a/mobly/controllers/android_device_lib/services/snippet_management_service.py
+++ b/mobly/controllers/android_device_lib/services/snippet_management_service.py
@@ -69,31 +69,36 @@ class SnippetManagementService(base_service.BaseService):
         class for supported configurations.
 
     Raises:
-      Error, if a duplicated name or package is passed in.
+      Error: if a duplicated name is passed in, or the same package has
+        already been registered under the same Android user ID.
     """
     # Should not load snippet with the same name more than once.
     if name in self._snippet_clients:
       raise Error(
           self,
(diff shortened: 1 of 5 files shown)


xianyuanjia (approved):

mobly/utils.py
```diff
@@ -474,36 +474,23 @@ def run_command(
       shell=shell,
       cwd=cwd,
       env=env,
-      universal_newlines=universal_newlines,
+      text=universal_newlines,  # "text" is introdcued in Python 3.7.
```
xpconanfan, line 477: is `text` functionally identical to `universal_newline`?

ko1in1u: Yes, `text` is the modern and preferred keyword for handling text mode. `universal_newlines` is the older and deprecated equivalent.

mhaoli (dismissed):

ko1in1u:
Sync to the head and pass internal TAP train, please help to review again.

mhaoli (approved):

PR #994, opened 2025-12-08
Allow loading snippets with the same package but different user IDs

mhaoli:
This is to relax the check in snippet_management_service to allow loading same snippet package with different user IDs.

With this CL, I verified that the same snippet app installed to different users can be loaded and used at the same time.

### DIFF
--- a/mobly/controllers/android_device_lib/services/snippet_management_service.py
+++ b/mobly/controllers/android_device_lib/services/snippet_management_service.py
@@ -87,7 +87,7 @@ class SnippetManagementService(base_service.BaseService):
       if package == client.package and new_snippet_user_id == client.user_id:
         raise Error(
             self,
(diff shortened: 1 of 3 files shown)


mobly/controllers/android_device_lib/services/snippet_management_service.py
```diff
@@ -69,7 +69,7 @@ def add_snippet_client(self, name, package, config=None):
         class for supported configurations.
 
     Raises:
-      Error, if a duplicated name or package is passed in.
+      Error: if a duplicated name or package is passed in.
```
xpconanfan: the error message here should be updated as package name alone is no longer a differentiating factor?

mhaoli:
I've created PR #996 to support the same feature by adding a client identifier property. So closing this.

PR #995, opened 2025-12-09
Add `enable_runtime_perms` flag to APK `install` helper function

eado:
Allows installing an APK without runtime permissions automatically granted in order to test permissions flows accurately.

Includes various formatting + readability fixes for `apk_utils.py`

### DIFF
--- a/tests/mobly/controllers/android_device_lib/apk_utils_test.py
+++ b/tests/mobly/controllers/android_device_lib/apk_utils_test.py
@@ -86,6 +86,16 @@ class ApkUtilsTest(unittest.TestCase):
         stderr=mock.ANY,
     )
 
+  def test_install_no_enable_perms(self):
+    self.mock_device.build_info = {'build_version_sdk': 23}
+    self.mock_device.adb.getprop.return_value = 'none'
+    apk_utils.install(self.mock_device, APK_PATH, enable_runtime_perms=False)
+    self.mock_device.adb.install.assert_called_with(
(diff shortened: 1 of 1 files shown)


mobly/controllers/android_device_lib/apk_utils.py
```diff
@@ -80,8 +80,9 @@ def install(
     device: AndroidDevice,
     apk_path: str,
     timeout: int = DEFAULT_TIMEOUT_INSTALL_APK_SEC,
-    user_id: Optional[int] = None,
-    params: Optional[Iterable[str]] = None,
+    user_id: int | None = None,
```
xpconanfan: the `Optional` is removed here and below?

mobly/controllers/android_device_lib/apk_utils.py
```diff
@@ -97,6 +98,7 @@ def install(
         install for the current user by default. Android's multi-user support
         did not realistically work until SDK 24.
     params: string list, additional parameters included in the adb install cmd.
+    enable_runtime_perms: Set the -g flag, which allows all runtime permissions.
```
xpconanfan: wrap "-g" with `` quote

xpconanfan (changes requested):

mobly/controllers/android_device_lib/apk_utils.py
eado: Gemini suggested that these are "best practices." If you disagree I can revert it.

PR #997, opened 2025-12-24
fix: ensure non-zero exit code on TestAbortAll (#960)

### LINKED ISSUE #960: Mobly exits with status code 0 after TestAbortAll
If a test executed by `test_runner` raises `signals.TestAbortAll` before any failures or errors are received, the test will exit with status code `0` instead of what I expected to be `1`. This is due to not executing `ok = False`

https://github.com/google/mobly/blob/991ecb2d9f7f859f8d86ab771d991473c30f80bc/mobly/test_runner.py#L71-L87

To make this more confusing, if a test failure or error is experienced before `TestAbortAll` is raised, the test will exit with status code `1`.

I believe both of these behaviors oppose most users' expectations of the implication of aborting a test. I would expect the test to always return status code `1` whenever an abort occurs. WDYT about adding `ok = False` to the `except signals.TestAbortAll` statement?

At the very least, some doc comments above each signal type would be valuable for setting expectations.

amytsengtw:
Ensures that Mobly returns a non-zero exit code when `TestAbortAll` is raised, 
preventing CI/CD systems from incorrectly reporting success.

Fixes #960

### DIFF
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -79,9 +79,10 @@ def main(argv=None):
         runner.run()
         ok = runner.results.is_all_pass and ok
       except signals.TestAbortAll:
-        pass
+        ok = False
+        logging.error('Test run aborted by TestAbortAll.')
       except Exception:
-        logging.exception('Exception when executing %s.', config.testbed_name)
+        logging.exception('Exception when executing %s.', config.test_bed_name)
         ok = False
   if not ok:
     sys.exit(1)
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -79,8 +79,8 @@ def main(argv=None):
         runner.run()
         ok = runner.results.is_all_pass and ok
       except signals.TestAbortAll:
-        ok = False
         logging.error('Test run aborted by TestAbortAll.')
+        ok = False
       except Exception:
         logging.exception('Exception when executing %s.', config.testbed_name)
         ok = False
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -82,7 +82,7 @@ def main(argv=None):
         ok = False
         logging.error('Test run aborted by TestAbortAll.')
       except Exception:
-        logging.exception('Exception when executing %s.', config.test_bed_name)
+        logging.exception('Exception when executing %s.', config.testbed_name)
         ok = False
   if not ok:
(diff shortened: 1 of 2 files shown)


xpconanfan (changes requested):
please add a corresponding unit test

mobly/test_runner.py
```diff
@@ -79,9 +79,10 @@ def main(argv=None):
         runner.run()
         ok = runner.results.is_all_pass and ok
       except signals.TestAbortAll:
-        pass
+        ok = False
+        logging.error('Test run aborted by TestAbortAll.')
       except Exception:
-        logging.exception('Exception when executing %s.', config.testbed_name)
+        logging.exception('Exception when executing %s.', config.test_bed_name)
```
xpconanfan: I don't think `test_bed_name` exists? why are you changing it?

amytsengtw: Thanks for catching that! It was a typo. I've reverted test_bed_name back to testbed_name in the latest commit

mobly/test_runner.py
```diff
@@ -79,7 +79,8 @@ def main(argv=None):
         runner.run()
         ok = runner.results.is_all_pass and ok
       except signals.TestAbortAll:
-        pass
+        ok = False
```
xpconanfan: let's stay consistent with the next block in terms of the order between the two lines

xpconanfan (changes requested):
Please add a test for this scenario
There are other tests in `test_runner_test` already

amytsengtw:
Fixed the testbed_name typo and added a self-contained unit test test_main_exits_with_error_on_test_abort_all. The test now creates its own temporary config to verify the exit code when signals.TestAbortAll is raised. Ready for another review

tests/mobly/test_runner_test.py
```diff
@@ -65,6 +65,22 @@ def _assertControllerInfoEqual(self, info, expected_info_dict):
         expected_info_dict['Controller Info'], info.controller_info
     )
 
+  def test_main_exits_with_error_on_test_abort_all(self):
```
xpconanfan, line 68: move this test next to the other `test_main_*` tests and try to maintain a reasonable order

xpconanfan:
Closing per #1006

PR #998, opened 2025-12-26
Improve controller initialization error messages with actionable hints

amytsengtw:
**Test Configuration:**
Tested with a malformed AndroidDevice configuration in YAML.
**Expected Output:**
Successfully verified that the hint "Ensure 'adb devices' shows your device" is displayed instead of a raw traceback.

### DIFF
--- a/mobly/controller_manager.py
+++ b/mobly/controller_manager.py
@@ -125,12 +125,15 @@ class ControllerManager:
       original_config = self.controller_configs[module_config_name]
       controller_config = copy.deepcopy(original_config)
       objects = module.create(controller_config)
-    except Exception:
-      logging.exception(
-          'Failed to initialize objects for controller %s, abort!',
-          module_config_name,
+    except Exception as e:
+      msg = (
+          f'Failed to initialize objects for controller "{module_config_name}". '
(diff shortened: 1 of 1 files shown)


mobly/controller_manager.py
```diff
@@ -125,12 +125,15 @@ def register_controller(self, module, required=True, min_number=1):
...
       controller_config = copy.deepcopy(original_config)
       objects = module.create(controller_config)
-    except Exception:
-      logging.exception(
-          'Failed to initialize objects for controller %s, abort!',
-          module_config_name,
+    except Exception as e:
+      msg = (
+          f'Failed to initialize objects for controller "{module_config_name}". '
+          'Check if the device is connected and the config is correct.'
```
xpconanfan, line 131: a controller can be anything, so we intentionally didn't say "device" in the error message.
and we should not have Android specific logic in the generic controller manager module.

I'm rejecting this PR

xpconanfan:
A controller can be anything, so we intentionally didn't say "device" in the error message.
And we should not have Android specific logic in the generic controller manager module.

Hence I'm rejecting this PR

PR #999, opened 2026-01-13
Fix incorrect timing of the "RPC request sent" log.

mhaoli:
Before this PR, the log message was printed after receiving the RPC response, which is incorrect.

### DIFF
--- a/mobly/controllers/android_device_lib/snippet_client_v2.py
+++ b/mobly/controllers/android_device_lib/snippet_client_v2.py
@@ -588,6 +588,7 @@ class SnippetClientV2(client_base.ClientBase):
     try:
       self._client.write(f'{message}\n'.encode('utf8'))
       self._client.flush()
+      self.log.debug('RPC request sent.')
     except socket.error as e:
       raise errors.Error(
           self._device,
--- a/mobly/snippet/client_base.py
+++ b/mobly/snippet/client_base.py
@@ -289,7 +289,6 @@ class ClientBase(abc.ABC):
 
       self.log.debug('Sending RPC request %s.', request)
       response = self.send_rpc_request(request)
-      self.log.debug('RPC request sent.')
 
       if self.verbose_logging or _MAX_RPC_RESP_LOGGING_LENGTH >= len(response):
         self.log.debug('Snippet received: %s', response)


ko1in1u (approved):

mobly/controllers/android_device_lib/snippet_client_v2.py
```diff
@@ -588,6 +588,7 @@ def _client_send(self, message):
     try:
       self._client.write(f'{message}\n'.encode('utf8'))
       self._client.flush()
+      self.log.debug('RPC request sent.')
```
xpconanfan, line 591: can we make this log more useful by adding relevant info instead of just being a static marker?

mobly/snippet/client_base.py
```diff
@@ -289,7 +289,6 @@ def _rpc(self, rpc_func_name, *args, **kwargs):
 
       self.log.debug('Sending RPC request %s.', request)
       response = self.send_rpc_request(request)
-      self.log.debug('RPC request sent.')
```
xpconanfan, line 292: is it useful to keep this log point with an accurate message?

mhaoli: Line 293-300 prints the content of the received RPC response, so I think we can simply delete line 292?

mobly/controllers/android_device_lib/snippet_client_v2.py
mhaoli: We've already printed the details before sending the request, so I think this should be enough.

I.e., here are the logs for one RPC request: 

```
01-13 23:04:16.662 DEBUG [AndroidDevice|xxx] Sending RPC request {"id": 0, "method": "isTechnologySupported", "params": [2]}.
01-13 23:04:16.669 DEBUG [AndroidDevice|xxx] RPC request sent.
01-13 23:04:16.689 DEBUG [AndroidDevice|xxx] Snippet received: {"id":0,"result":false,"callback":null,"error":null}
```

xpconanfan: maybe include the id?
remember the log will be mixed with many other logs. so a log line without an identifier is a log line that will be a source of ambiguity in debugging

mhaoli: Agree it's better to include the id. But the solution to support this is not that straightforward since the public API `send_rpc_request` does not contain an id today. 

How about adding the id in a separate PR? I could create it immediately after submitting this one.

xpconanfan: Ah, ok. Nvm then. No need to make that change

xpconanfan (approved):

PR #1001, opened 2026-01-13
Capture logcat messages while logcat collection was suspended

### LINKED ISSUE #1000: adb logcat collection skips over log messages while suspended
When you use the logcat collection feature and need to suspend the logcat collection, since the USB connection gets disabled, all logcat messages between the time the logcat collection was suspended until it gets resumed will be dropped, since we only look back one logcat message when starting the logcat collection: https://github.com/google/mobly/blob/master/mobly/controllers/android_device_lib/services/logcat.py#L250

```
    cmd = ' "%s" -s %s logcat -v threadtime -T 1 %s >> "%s" ' % (
        adb.ADB,
        self._ad.serial,
        self._config.logcat_params,
        self.adb_logcat_file_path,
    )
```

logcat also supports that a timestamp can be provided to the `-T` option, so that we can properly resume the logcat collection if we store the current system time as part of the suspend operation, and provide it to the command above when we resume the logcat collection.

xr-gh:
Prior to this change the logcat collection would be missing logcat messages that got created while the logcat collection was suspended if the logcat collection got subsequently resumed.

This change stores the local time when the adb connection got suspended, so that we can use it to specify from which point in time we want to collect logcat messages again. This ensures that the final log file contains the log messages while the logcat collection was suspended once the logcat collection got resumed.

Fixes #1000

### DIFF
--- a/mobly/controllers/android_device_lib/services/logcat.py
+++ b/mobly/controllers/android_device_lib/services/logcat.py
@@ -63,6 +63,7 @@ class Logcat(base_service.BaseService):
     self._adb_logcat_process = None
     self._adb_logcat_file_obj = None
     self.adb_logcat_file_path = None
+    self._last_connection_time = None
     # Logcat service uses a single config obj, using singular internal
     # name: `_config`.
     self._config = configs if configs else Config()
@@ -247,9 +248,14 @@ class Logcat(base_service.BaseService):
     # In debugging mode of IntelijIDEA, "patch_args" remove
     # double quotes in args if starting and ending with it.
     # Add spaces at beginning and at last to fix this issue.
-    cmd = ' "%s" -s %s logcat -v threadtime -T 1 %s >> "%s" ' % (
+    if self._last_connection_time is not None:
+      t_argument_value = self._last_connection_time
+    else:
+      t_argument_value = '1'
+    cmd = ' "%s" -s %s logcat -v threadtime -T "%s" %s >> "%s" ' % (
         adb.ADB,
         self._ad.serial,
+        t_argument_value,
         self._config.logcat_params,
         self.adb_logcat_file_path,
     )
@@ -270,6 +276,7 @@ class Logcat(base_service.BaseService):
     except Exception:
       self._ad.log.exception('Failed to stop adb logcat.')
     self._adb_logcat_process = None
+    self._last_connection_time = None
 
   def pause(self):
     """Pauses logcat.
@@ -279,6 +286,12 @@ class Logcat(base_service.BaseService):
     some logs would be lost.
     """
     self._stop()
+    try:
+      response = self._ad.adb.shell(['date', '-Is', r'+%Y-%m-%d\ %H:%M:%S.%N'])
+      if response:
+        self._last_connection_time = response.decode('utf-8').strip()
+    except adb.AdbError as e:
+      self._ad.log.exception('Failed to get host time.')
 
   def resume(self):
     """Resumes a paused logcat service."""
--- a/tests/mobly/controllers/android_device_lib/services/logcat_test.py
+++ b/tests/mobly/controllers/android_device_lib/services/logcat_test.py
@@ -67,6 +67,8 @@ MOCK_LOGPERSIST_START_MISSING_ADB_ERROR = adb.AdbError(
     0,
 )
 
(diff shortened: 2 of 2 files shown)


tests/mobly/controllers/android_device_lib/services/logcat_test.py
```diff
@@ -264,8 +265,101 @@ def test_update_config_while_running(
...
+    adb_cmd = ' "adb" -s %s logcat -v threadtime -T "1"  >> %s'
+    start_proc_mock.assert_called_with(
+        adb_cmd % (ad.serial, '"%s" ' % expected_log_path), shell=True
+    )
+
+    clear_adb_mock.assert_called_once_with()
+    self.assertTrue(logcat_service.is_alive)
+
+    def pause_resume(date):
+      nonlocal date_result
```
xpconanfan: can we avoid using `nonlocal` ?

tests/mobly/controllers/android_device_lib/services/logcat_test.py
```diff
@@ -264,8 +265,101 @@ def test_update_config_while_running(
...
+      get_timestamp_mock,
+      clear_adb_mock,
+      open_logcat_mock,
+      stop_proc_mock,
+      start_proc_mock,
+      create_dir_mock,
+      FastbootProxy,
+      MockAdbProxy,
+  ):
+    DATE_COMMAND = ['date', '-Is', r'+%Y-%m-%d\ %H:%M:%S.%N']
```
xpconanfan: I didn't expect to see so much change in the existing test.
is it possible to simplify the logic here?

tests/mobly/controllers/android_device_lib/services/logcat_test.py
```diff
@@ -264,8 +265,101 @@ def test_update_config_while_running(
...
+    )
+    adb_cmd = ' "adb" -s %s logcat -v threadtime -T "1"  >> %s'
+    start_proc_mock.assert_called_with(
+        adb_cmd % (ad.serial, '"%s" ' % expected_log_path), shell=True
+    )
+
+    clear_adb_mock.assert_called_once_with()
+    self.assertTrue(logcat_service.is_alive)
+
+    def pause_resume(date):
```
xpconanfan: It's difficult to understand the purpose of this from the name.
what are we doing here? could we make the test more readable and easier to understand? E.g. simplify the structure, change the name, add inline comments.

xpconanfan (changes requested):

tests/mobly/controllers/android_device_lib/services/logcat_test.py
xr-gh: I added comments around the code, but I feel this is as simple as it gets, not sure how to simplify this any further.

tests/mobly/controllers/android_device_lib/services/logcat_test.py
xr-gh: I have renamed the function to `pause_logcat_service_resume_logcat_service` and added comments into the code to explain what we are doing here. I hope that helps.

tests/mobly/controllers/android_device_lib/services/logcat_test.py
xpconanfan: could nested functions be un-nested?

xr-gh: Done, the nested functions got factored out

xpconanfan (approved):

mobly/controllers/android_device_lib/services/logcat.py
```diff
@@ -279,6 +286,12 @@ def pause(self):
     some logs would be lost.
     """
     self._stop()
+    try:
+      response = self._ad.adb.shell(['date', '-Is', r'+%Y-%m-%d\ %H:%M:%S.%N'])
```
mhaoli, line 290: With the current impl, I think we might still be losing a small number of logcat messages. More specifically, this could be due to the timestamp difference between the time the logcat subprocess is killed and the time the `date` command is executed.

A better approach would seem to be to calculate the timestamp by adding 0.001 to the timestamp of the last logcat message in file `self.adb_logcat_file_path`?

xpconanfan: I don't think we need to do that at this point. It's hard to estimate the time.

Issue #950, opened 2025-01-08
Refactor run_suite_class


jrotkiewicz:
```
def run_suite_class(argv=None):
  """Executes tests in the test suite.

  Args:
    argv: A list that is then parsed as CLI args. If None, defaults to sys.argv.
  """
  cli_args = _parse_cli_args(argv)
  suite_class = _find_suite_class()
  if cli_args.list_tests:
    _print_test_names([suite_class])
    sys.exit(0)
  test_configs = config_parser.load_test_config_file(
      cli_args.config, cli_args.test_bed
  )
  config_count = len(test_configs)
  if config_count != 1:
    logging.error('Expect exactly one test config, found %d', config_count)
  config = test_configs[0]
  runner = test_runner.TestRunner(
      log_dir=config.log_path, testbed_name=config.testbed_name
  )
  suite = suite_class(runner, config)
  console_level = logging.DEBUG if cli_args.verbose else logging.INFO
  ok = False
  with runner.mobly_logger(console_level=console_level):
    try:
      suite.setup_suite(config.copy())
      try:
        runner.run()
        ok = runner.results.is_all_pass
        print(ok)
      except signals.TestAbortAll:
        pass
    finally:
      suite.teardown_suite()
  if not ok:
    sys.exit(1)
'''

There are two issues with this function:
1. it doesn't support --list_tests - it's currently broken
2. it doesn't work as proposed here:
```
2. Create a subclass of base_suite.BaseSuite and add the individual test
classes. Using the BaseSuite class allows users to define their own setup
and teardown steps on the suite level as well as custom config for each test
class.

.. code-block:: python

  from mobly import base_suite
  from mobly import suite_runner

  from my.path import MyFooTest
  from my.path import MyBarTest

  class MySuite(base_suite.BaseSuite):

    def setup_suite(self, config):
      # Add a class with default config.
      self.add_test_class(MyFooTest)
      # Add a class with test selection.
      self.add_test_class(MyBarTest,
                          tests=['test_a', 'test_b'])
      # Add the same class again with a custom config and suffix.
      my_config = some_config_logic(config)
      self.add_test_class(MyBarTest,
                          config=my_config,
                          name_suffix='WithCustomConfig')

  if __name__ == '__main__':
    suite_runner.run_suite_class()
"""
```

[...]

PR #1006, opened 2026-04-30
fix: set ok=False when TestAbortAll is raised in run_suite_class

devimano2011:
Related to #950 and #960

## Problem

In `run_suite_class()` in `suite_runner.py`, when `signals.TestAbortAll` is raised, the `except` block only has `pass` and does not set `ok = False`. This means if a test aborts before any failures, the suite exits with code 0 instead of 1.

This is the same issue that was fixed in `test_runner.py` by PR #997 for issue #960, but `run_suite_class()` in `suite_runner.py` was missed.

## Fix

Change `pass` to `ok = False` in the `except signals.TestAbortAll` block so that aborting a test suite always results in a non-zero exit code.

### DIFF
--- a/mobly/suite_runner.py
+++ b/mobly/suite_runner.py
@@ -368,7 +368,7 @@ def run_suite_class(argv=None):
         ok = runner.results.is_all_pass
         print(ok)
       except signals.TestAbortAll:
-        pass
+        ok = False
     finally:
       suite.teardown_suite()
       suite_record.suite_end()


xpconanfan:
In our usage, abort is usually intentional, so we don't necessarily consider this "not ok".
Most common case is, a device doesn't support the feature a test is meant to test, hence abort.

So I'm not sure whether this should be considered "not ok".

devimano2011:
Thank you for the feedback! That's a valid point — I wasn't aware that abort is often intentional in your usage (e.g. device doesn't support the feature under test).

Looking at PR #997 more carefully, that fix was motivated by cases where `TestAbortAll` is raised due to an *unexpected* failure, where exiting with code 0 would silently hide the problem from CI.

Given your clarification, I see a few options:

1. **Revert this PR** — keep `pass` in `run_suite_class` since aborting is considered intentional there.
2. 2. **Make it configurable** — distinguish intentional aborts (feature not supported) from unexpected ones.
3. 3. **Keep the fix** — if consistency with `test_runner.py` is preferred regardless.
I'll defer to your judgment. Happy to close/revert if this doesn't align with the intended behavior.

xpconanfan:
Closing this PR as this is intended behavior.
One always has the option of interpreting this signal differently in their custom suite runner

xpconanfan:
Closing this PR as this is intended behavior.
One always has the option of interpreting this signal differently in their custom suite runner

PR #1010, opened 2026-07-14
Add key frame to asserts

skyyaoyao:
### Description
   This PR automatically captures key frame data (`action`, `timestamp`, and `details`) upon any Mobly assertion failure and injects it into the test records. 
   
   By injecting this into the assertion's `extras` dictionary, we ensure the metadata is seamlessly captured into the `test_summary.yaml` for downstream parsers, without requiring manual tracking or adding overhead to passing assertions. 
   
   A new `_add_key_frame` helper is implemented and tied into both standard unittest wrappers (`_call_unittest_assertion`) and direct Mobly assertions (`fail()`).
   
   ### Testing Done
   - Verified the `key_frame` block successfully serializes into `test_summary.yaml` logs upon assertion failures with real Mobly test.
   - Re-ran the full Mobly py-test suite (`pytest tests/mobly`). Updated the internal `base_test_test.py` tests that rigorously expected `actual_record.extras` to be `None` to instead gracefully accept the injected `key_frame` data on failures.

### DIFF
--- a/mobly/asserts.py
+++ b/mobly/asserts.py
@@ -15,12 +15,27 @@
 import re
 import unittest
 
+from mobly import logger
 from mobly import signals
 
 # Have an instance of unittest.TestCase so we could reuse some logic
 # from python's own unittest.
 _pyunit_proxy = unittest.TestCase()
 _pyunit_proxy.maxDiff = None
+
+
+def _add_key_frame(extras, action, details):
+  """Helper to inject key frame data into assertion extras."""
+  key_frame_data = {
+      'action': action,
+      'timestamp': logger.get_log_line_timestamp(),
+      'details': details,
+  }
+  if extras is None:
+    extras = {}
+  if isinstance(extras, dict):
+    extras['key_frame'] = key_frame_data
+  return extras
 
 
 def _call_unittest_assertion(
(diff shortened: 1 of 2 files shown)


xpconanfan:
Rejecting this implementation for the following reasons:

1. **Pollutes User-Owned `extras`**:
   `extras` is reserved for test authors to attach custom, domain-specific data. Injecting framework telemetry (`key_frame`) directly into `extras` breaks user space and changes `actual_record.extras` from `None` to a dict on failure (as seen by the breaking unit tests in `base_test_test.py`).
2. **Violates the Generic Data Channel Philosophy**:
   Following the Sponge pattern (`sponge.md`), Mobly decouples telemetry via the generic message stream (`record_data()` / `Type: UserData` in `test_summary.yaml`). Framework-level data should not be hardcoded into core test signal objects.
3. **Incomplete Assertion Visibility**:
   Hooking only into exception raising inside `except AssertionError` captures only *failing* assertions, completely missing *passing* assertions.
4. **Brittle & Invasive**:
   Mutates user dicts in-place, silently ignores non-dict `extras`, and tightly couples assertion execution to logging timestamp formatting.

PR #1011, opened 2026-07-31
Include py 3.13 and 3.14 in CI

xpconanfan:
IncludeInclude py 3.13 and 3.14 in CI


xianyuanjia (dismissed):
Run go/github-zizmor-help?polyglot=github-com#auto-fixes-via-the-fix-flag if the errors prevent merge

xianyuanjia (approved):

PR #1013, opened 2026-07-31
Exit with status code 1 when a test aborts all tests

### LINKED ISSUE #960: Mobly exits with status code 0 after TestAbortAll

Sanjays2402:
Fixes #960

`test_runner.main()` caught `signals.TestAbortAll` and did nothing, leaving `ok` unchanged, so a run that aborted before any failure or error was recorded exited with status code 0. If a failure occurred before the abort, the exit code was 1 — inconsistent for the same abort. The handler now sets `ok = False`, so an abort always exits 1.

Added `test_main_with_abort_all` to `tests/mobly/test_runner_test.py`; it fails on the current code (`sys.exit` never called) and passes with the fix.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/mobly/test_runner.py
+++ b/mobly/test_runner.py
@@ -79,7 +79,7 @@ def main(argv=None):
         runner.run()
         ok = runner.results.is_all_pass and ok
       except signals.TestAbortAll:
-        pass
+        ok = False
       except Exception:
         logging.exception('Exception when executing %s.', config.testbed_name)
         ok = False
--- a/tests/mobly/test_runner_test.py
+++ b/tests/mobly/test_runner_test.py
@@ -337,6 +337,25 @@ class TestRunnerTest(unittest.TestCase):
 
   @mock.patch(
       'mobly.test_runner._find_test_class',
+      return_value=type('SampleTest', (), {}),
+  )
+  @mock.patch(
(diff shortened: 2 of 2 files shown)


xpconanfan:
It's fair to say if there were previous failures/errors, abort should result in non-zero exist code.
But this PR doesn't do that.

As discussed in many other CLs, we want to distinguish intentional abort from the case you proposed.

xpconanfan:
I've sent a fix in #1019

PR #1023, opened 2026-08-08
Support using expects.expect_no_raises as a decorator

### LINKED ISSUE #480: Make a decorator for `expects.expect_no_raises`
So it can be easily applied to a whole function.

xpconanfan:
Fixes #480

### DIFF
--- a/mobly/expects.py
+++ b/mobly/expects.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import contextlib
+import functools
 import logging
 import time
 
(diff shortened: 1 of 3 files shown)


mobly/expects.py
```diff
@@ -135,31 +136,67 @@ def expect_equal(first, second, msg=None, extras=None):
...
+    def helper_function(arg):
+      do_something(arg)
+
+    @expects.expect_no_raises
+    def bare_decorated_function():
+      do_something()
 
   Args:
-    message: string, custom message to add to exception's `details`.
+    message: string or callable, custom message to add to exception's `details`.
```
xianyuanjia: Don't mention `callable` as an accepted arg type here. Even though the decorator internally calls the wrapped function as the same arg, it doesn't serve the same purpose as `message`, and this behavior shouldn't be surfaced to the user.

xianyuanjia (dismissed):

mobly/expects.py
xpconanfan: Good catch

xianyuanjia (approved):

PR #1031, opened 2026-09-11
Write setup_class error record to summary even if on_fail aborts

### LINKED ISSUE #1029: `setup_class` error record is not written to `test_summary.yaml` when `on_fail` raises `TestAbortAll`

hjfreyer:
When `setup_class` failed and the user's `on_fail` callback raised an abort signal (e.g. via `asserts.abort_all` or `asserts.abort_class`), the signal propagated out of `_exec_procedure_func` before `update_record()` and `summary_writer.dump()` ran. The error was counted in `self.results` but no corresponding record was ever written to `test_summary.yaml`, producing a summary where the error count did not match the records present.

Move the record finalization and dump into a `finally` block around the `on_fail` call, mirroring the pattern already used in `exec_one_test`. This covers both the exception path and the `expects` path; in the latter `test_error()` also moves into the `finally` block so the record is finalized even when `on_fail` aborts.

The existing regression test only checked the in-memory results, so it did not catch this. It now also asserts on the summary file contents, and new tests cover the `expects` path and `TestAbortClass`.

Fixes https://github.com/google/mobly/issues/1029


xpconanfan:
#1032 will cover this

